PR #6, opened 2020-02-05
src/tcltk/tcltk8.5.4/tcl8.5.4/Makefile: fix leftover $(MAKE) commands

flokli:
Those were missed in https://github.com/B-Lang-org/bsc/pull/5.


flokli:
cc @bpfoley

PR #9, opened 2020-02-05
README.md: document perl is required during build

flokli:
This is used to invoke src/stp/src/AST/genkinds.pl, which produces
ASTKind.{h,cpp}.


quark17:
I'd prefer to put this requirement in the later paragraph that documents requirements from STP.  I've pushed commit 562af069, in place of this request.  (But thanks for pointing it out!)

flokli:
Ok sure :-)

PR #15, opened 2020-02-05
yices/Makefile: set LDCONFIG to ldconfig

flokli:
The Makefile inside the git submodule hardcodes LDCONFIG to
`/sbin/ldconfig`, which doesn't necessarily exist at that path.

Instead, override it to be just `ldconfig`, meaning it'll pick
it from `$PATH`.


quark17:
After committing this, I discovered that it actually breaks the build on Debian10, which has /sbin/ldconfig but doesn't have /sbin in the path.  Any thoughts on a solution?

flokli:
Hmm, we could prepend $PATH there, but it's also a bit ugly.

I tried patching the ldconfig stuff out entirely, but then ran into issues of the linker trying to find the library with a wrong name.

What about doing something like https://github.com/NixOS/nixpkgs/commit/6ff8685057c1c6b0dbeed1c62d616696ae9b4dcf#diff-c25c55d2b460874ebb8c24ad4837e75f for yices, so we can avoid having to call ldconfig alltogether?

quark17:
I'm not sure I understand your suggestion (I don't offhand understand what ldconfig is being used for in the yices repo), but what if we conditionally set LDCONFIG earlier in the Makefile and then pass that value in:

`$(MAKE) -C $(VERSION) install LDCONFIG=$(LDCONFIG)`

(I would have thought that `export` would work for this, but it doesn't have the same effect.)

And I don't if the condition should be whether `/sbin/ldconfig` exists:

`ifeq (,$(wildcard /sbin/ldconfig))`
`LDCONFIG=ldconfig`
`else`
`LDCONFIG=/sbin/ldconfig`
`endif`

Or if the condition should be whether `ldconfig` is in the path (either by call `which ldconfig` in the shell or calling `ldconfig --version` in the shell).  I'm less sure how to express that though, so I'd go with the above.

flokli:
Hm, at least in the case of NixOS, I was able to just add set `LDCONFIG=true` to makeFlags to get bsc to build and run the smoketest, so maybe we don't need all the `ldconfig` stuff at all :laughing:

quark17:
So I can revert this commit?

flokli:
if that unblocks you for now, yes. For NixOS, we could just set `LDCONFIG=ldconfig` when running, or set it `LDCONFIG=true` (as it currently is in the [nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/79468))

However, I'd propose checking if defaulting `LDCONFIG=true` works for Debian10/other distros as well - that way, we wouldn't need to apply distro-specific configuration.

quark17:
Ah, yes, that works!  I pushed that change.  Thanks!

flokli:
I have to take that back - at least here, ghc puts a `NEEDED` `libyices.so.2.6` in `inst/bin/core/bsc`, but without executing `ldconfig`, I only have the following libraries in `src/yices/v2.6/yices2-inst/lib/`:

```
-rwxr-xr-x 1 nixbld1 nixbld 4231288 Feb  9 00:14 libyices.a
lrwxrwxrwx 1 nixbld1 nixbld      17 Feb  9 00:14 libyices.so -> libyices.so.2.6.2
-rwxr-xr-x 1 nixbld1 nixbld 1791040 Feb  9 00:14 libyices.so.2.6.2
```

meaning executing `bsc` will fail:

```
/build/source/inst/bin/bsc -stdlib-names -bdir /build/source/build/bsvlib -p . -vsearch /build/source/build/bsvlib -no-use-prelude Prelude.bs
/build/source/inst/bin/core/bsc: error while loading shared libraries: libyices.so.2.6: cannot open shared object file: No such file or directory
make[3]: *** [Makefile:22: /build/source/build/bsvlib/Prelude.bo] Error 127
```

As this would require patching inside the submodule (which we can't really), I propose not setting `LDCONFIG=` from a parent `Makefile` at all - defaulting to `/sbin/ldconfig` for most distros, and setting it to elsewhere on other distros.

Sorry for the noise :-/

PR #17, opened 2020-02-06
Make use of ln more portable

arjenroodselaar:
The stp and yices build steps rely on a GNU extension of ln (-T), causing the
build to fail on MacOS. Since the intended behavior seems to be to create
recursive links up the directory tree this diff should make that more portable.

### DIFF
--- a/src/stp/Makefile
+++ b/src/stp/Makefile
@@ -9,7 +9,7 @@ all: install
 
 install:
 	$(MAKE) -C $(SRC) install
-	ln -fsT HaskellIfc include_hs
+	ln -fsn HaskellIfc include_hs
 	install -m 755 -d $(PREFIX)/lib/SAT
 	install -m 644 lib/* $(PREFIX)/lib/SAT
 
--- a/src/yices/Makefile
+++ b/src/yices/Makefile
@@ -10,9 +10,9 @@ all: install
 
 install:
 	$(MAKE) -C $(VERSION) install
-	ln -fsT $(VERSION)/include    include
-	ln -fsT $(VERSION)/lib        lib
-	ln -fsT $(VERSION)/include_hs include_hs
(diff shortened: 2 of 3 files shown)


quark17:
Your version is how I originally had it, but the problem is that this command:

`ln -fs HaskellIfc include_hs`

created a link at `include_hs/include_hs` when `include_hs` already existed.  That's why I added the -T flag, not realizing that it was a GNU extension.

I now see that the BSD version has a flag `h`:
> If the target_file or target_dir is a symbolic link, do not follow it.  This is most useful with the -f option, to replace a symlink which may point to a directory.

And the BSD version also defines `n` as an alias for `h` for "compatibility with other ln implementations."  I see that the GNU version has `n`:
> treat LINK_NAME as a normal file if it is a symbolic link to a directory

So maybe we just replace `T` with `n`?

I also notice that both have `F` which maybe also gets the behavior that we want, but is unclear.  And, in the worst case, we could always define the LN command at the top of the Makefile and use if-def to define it with different arguments depending on OSTYPE.  But fortunately it looks like we can use `n` here, yes?

FYI, I'm also open to changing how all of this indirection works.  Before releasing BSC on GitHub, I attempted to cleanup how some of the "vendor" directories worked (while not changing too much, which included some of this indirection) and am not strongly committed to the choices that I made.  For example, it's unclear if separate version subdirectories are needed, so that's one level of indirection we can consider eliminating.

arjenroodselaar:
Thank you for explaining the history/intend here. I updated the PR with `n` since that indeed seems to have the intended effect and is portable.

Re: whether or not to eliminate one level of indirection. I've definitely done this as well and it turns out I rarely need to simultaneously support multiple versions of the same vendored library. But at this point I wouldn't go out of my way to remove it.

PR #21, opened 2020-02-06
Clean libraries

bpfoley:
A sequence of patches to clean up almost all of the warnings when building Libraries

They consist of
1. Renaming unused bindings to _ or _foo if the name seemed important
2. Fixing a lot of cases where bindings shadowed other bindings
3. Occasional pieces of code simplification where it was easier than renaming bindings

### DIFF
--- a/src/Libraries/Base1/Assert.bs
+++ b/src/Libraries/Base1/Assert.bs
@@ -32,7 +32,7 @@ staticAssert b s = if not testAssert || b then
 --@ function Action dynamicAssert(Bool b, String s);
 --@ \end{libverbatim}
 dynamicAssert :: Bool -> String -> Action
-dynamicAssert = if not testAssert then (\ b s -> noAction)
+dynamicAssert = if not testAssert then (\ _ _ -> noAction)
 		else (\ b s -> if b then noAction else 
                         action
                           $display (assertMessage "Dynamic" s)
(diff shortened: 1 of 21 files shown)


quark17:
You've fixed some shadowing by adding a prime to the name.  In some cases, the values are unrelated, they just use the same generic name, like `x` -- for example, when a function takes a value called `x` and then later does a pattern match on a list to get `x` and `xs`.  To me, a name like `x'` implies that it's a new version of `x`, and would be misleading in this case (so I'd prefer a different naming to resolve the shadowing).  Is that just me?

bpfoley:
That's true. `x'` usually means 'modified `x` to me too. Have you any preferred naming convention?

quark17:
No particular conventions.  On line 1026 of Prelude.bs, the `Cons x _` is taking the head of a list of bits, so either `b` or `h` works.  Similarly, lines 1033, 1604, and 1694 can use `b` and `bs`.  For method arguments like line 1891, I guess you could use `v` or `val`; or even just change the first use of `x` to be `a_val`, since its purpose seems to be merely to get a handle on the type `a` -- and I wonder if the instantiation of `x` is even needed and could just be replaced with `(_ :: a)` as the argument to `typeof`?  At line 3580, maybe use `x1` and `x2` to not conflict with `v1` and `v2`?  Etc.

In ListN.bs, I see that many places had local definitions of `unL` that shadowed a global definition.  Those all seem to be defining the same function, right?  Rather than add a prime to the name, can we just remove those local definitions and leave the expression as is, using `unL` as globally defined?

bpfoley:
Thanks for the good suggestions.

Issue #16, opened 2020-02-05
'Minisat::OutOfMemoryException' during build


flokli:
During build, `/build/source/inst/bin/bsc -stdlib-names -bdir /build/source/build/bsvlib -p . -vsearch /build/source/build/bsvlib -no-use-prelude Prelude.bs` is invoked.

It fails

```
terminate called after throwing an instance of 'Minisat::OutOfMemoryException'
make[3]: *** [Makefile:22: /build/source/build/bsvlib/Prelude.bo] Aborted (core dumped)
make[3]: Leaving directory '/build/source/src/Libraries/Base1'
make[2]: *** [Makefile:31: build] Error 2
make[2]: Leaving directory '/build/source/src/Libraries'
make[1]: *** [Makefile:16: install] Error 2
make[1]: Leaving directory '/build/source/src'
make: *** [Makefile:24: install] Error 2
```

Full Log: https://termbin.com/0nb3

Note the machine has ~60GiB of RAM, so I hope it's not really not enough memory ;-)

quark17:
Can you say more about the environment that you're running in (OS, GHC version, GCC version)?  I have run into this when building on Debian 10, which uses GHC 8.4.  Some issues prevented me from installing other GHC versions (without spending more time).  I saw that BSC was indeed taking as many gigs of memory as it could, before eventually failing in the STP solver.  I don't believe that STP is at fault, because I changed the solver to be Yices instead (using the -sat-yices flag or by changing the default value in FlagsDecode.hs) and BSC built with GHC 8.4 on Debian 10 would still increase in memory until it ran out.

@bpfoley has also reported seeing this error, even when just running "bsc -help", when building with GHC 8.4 and GCC 9.  It's unclear if his issue is the same, but similar message.

This error does not come up when using GHC 8.0 or earlier, on other OSes.

flokli:
Full repro, including the environment is here: https://github.com/flokli/nixpkgs/commit/e9db0940326b3d67f2c1a694c34abe69fdfdb5f7

This is currently using GHC 8.6.5 and GCC 9.2.0 - just invoke `nix-build -A bluespec-bsc` from the repo root (requires nix, can be installed from https://nixos.org/nix/).

I'll check with other compiler combinations and report back.

Is there a Slack / IRC channel to discuss these findings?

flokli:
I can confirm I also run into the same memory consumption issues with GHC 8.4.4.

[...]

PR #25, opened 2020-02-06
Remove CUDD 2.5.0 SAT solver (and BSC -sat-cudd option)

bpfoley:
It was an extremely old version of CUDD. CUDD is no longer supported
and both STP and yices are much newer and have better performance
characteristics.


flokli:
Removal of CUDD was suggested in https://github.com/B-Lang-org/bsc/issues/16#issuecomment-583094666.

PR #27, opened 2020-02-06
Remove unused top level bindings

bpfoley:


quark17:
While those definitions are not currently being used by our C Syntax construction, could they be used in the future?  It seems like a useful record to leave them commented out, rather than delete them, so I've done that in commit e4c3a833.

PR #30, opened 2020-02-07
General fixes

thoughtpolice:
I've gotten Bluespec building reliably with Nix and running the smoke test properly. While doing this, I wrote several other changes that I think can be merged independently and un-controversially.

This includes several touch-ups to the README.md, which I think improve the "up-front advertisement" in the first 30 seconds of the README. You can see what that looks like in my fork: https://github.com/thoughtpolice/bsc/tree/aseipp/wip (note I removed the "Line count" badge since it's so wildly inaccurate; we'd have to unvendor tcl/tk to get accurate counts, since the external badge tool doesn't support `.gitattributes`, apparently.)

The functional changes to the code (tclsh reuse, `-Werror`), are all necessary to get Bluespec building with Nix.

Every commit has detailed explanations in the message, and are independent of each other, so I strongly suggest reading them one-by-one.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
+charset=utf-8
+insert_final_newline=true
+trim_trailing_whitespace=true
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,16 @@ rem_build:
 install:
 	$(MAKE)  -C src  PREFIX=$(PREFIX)  install
 
+# In the future, this will be much more expansive, and run the actual test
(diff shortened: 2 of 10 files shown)


quark17:
Thank you!  These all look good.

I had also noticed the GCC 9 warning.  The code is fine, gcc is just naively warning.  We could consider changing the code -- rather than one call to strncpy with a conditional length, we could conditionally call strcpy or strncpy based on whether the string fits -- but that seemed too much.  I'm happy to leave out -Werror until someone has a better suggestion.

PR #32, opened 2020-02-07
tree-wide: fix shebangs

flokli:
There shouldn't be a space between #! and the executable.


flokli:
Oh wow, seems they're [indeed fine](https://unix.stackexchange.com/questions/276751/is-space-allowed-between-and-bin-bash-in-shebang). Sorry for the noise.

PR #33, opened 2020-02-07
feat(bsc): add -print-bs-libdir flag

thoughtpolice:
Previous versions of the Bluespec compiler required you to specify the
`BLUESPECDIR` environment variable to locate the compiler toolchain.
That's no longer necessary, which is great, but there is another use
case provided by that variable: it makes it easy to find the necessary
Verilog files to add to a Bluespec project. Tools such as `yosys-bsv`
rely on functionality like this, so they can automatically synthesize
working designs out of the box.

While there's various ways to coax this out of the compiler, this new
`-print-bs-libdir` flag just does one thing only: prints out the path to
the `lib/` subdirectory, and nothing else. The intention is essentially
to use it for scripting purposes, to find the proper library directory
of `bsc` without a lot of shennanigans. Putting this functionality in
`bsc` also makes it more robust: if the semantics of install dirs ever
change (e.g. a concrete `--prefix` flag), scripts using this feature
will keep working regardless.

### DIFF
--- a/src/comp/Flags.hs
+++ b/src/comp/Flags.hs
@@ -93,6 +93,7 @@ data Flags = Flags {
         printFlags :: Bool,
         printFlagsHidden :: Bool,
         printFlagsRaw :: Bool,
+        printBluespecDir :: Bool,
         preprocessOnly :: Bool,
         promoteWarnings :: MsgListFlag,
         readableMux :: Bool,
--- a/src/comp/FlagsDecode.hs
+++ b/src/comp/FlagsDecode.hs
@@ -171,7 +171,8 @@ decodeArgs prog args cdir =
                  then if (null anames)
                       then if ((printFlags flags) ||
                                (printFlagsHidden flags) ||
-                               (printFlagsRaw flags))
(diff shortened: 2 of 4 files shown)


quark17:
As discussed on IRC, this can be done with bluetcl as seen in the shell script that gets generated for Bluesim simulations (as constructed in src/comp/bsc.hs):

``BLUESPECDIR=`echo 'puts $env(BLUESPECDIR)' | bluetcl` ``

And it can be seen in the output of `bsc -v` and `bsc -print-flags` and `bsc -print-flags-raw` but requires extraction.  I am in favor of adding a flag for this.  But maybe we should think about what are all the values we want to print, and whether they each get a flag or we add some general flags for printing values, etc.

thoughtpolice:
I actually somewhat like the proposal I made on IRC to have a more general mechanism to dump search paths. I'll file an issue about it later and we can rework this PR to match what we come up with!

thoughtpolice:
Closing this for now in lieu of my proposal (which I'll dig up...)

PR #34, opened 2020-02-07
chore(repo): src/Parsec -> src/comp/Parsec

thoughtpolice:
This was actually already referred to as such in the COPYING file, so we
might as well clean up the `src` directory a bit and make that note
accurate.

This also removes the HTML copy of the Parsec license, since I believe
it's simply redundant.


quark17:
Parsec is now available as a library, so we should eliminate our snapshot (which is old) and update to use the current library.  I'd prefer to do that instead of these changes.

bpfoley:
...and in fact I'm working on that at the moment. I just need to figure out the parsec 2 -> 3 API changes :)

thoughtpolice:
That's totally fine with me! Unfortunately I can't help with Parsec 3 changes, it happened so long ago my foggy mind has forgotten all the changes. :)

I'd be more than happy to also help @bpfoley with any "Haskell modernization" attempts like this, e.g. re-using libraries like `parsec` where appropriate, and cleaning things up to use modern more modern features and libraries. Of course, for now, I'm working on packaging concerns, but this is also a very fruitful thing to do.

However, it'd be useful to think about what the toolchain/dependency requirements are; some dependencies will invariably imply limits on the toolchain and that's probably fine, it's just nice having everyone on the same page.

That should probably happen in a separate issue though. I'll close this one instead!

bpfoley:
Thank you, I appreciate it. See also #37

PR #43, opened 2020-02-09
yices: add missing symlink

flokli:
When not using ldconfig anymore, `libyices.so.2.6` (symlink to
`libyices.so.2.6.2`) might be missing, which could cause problems
invoking bsc, as `libyices.so.2.6` might be in the ELF's NEEDED, as seen
on NixOS.

This adds the missing symlink, like `ldconfig` would do, working around
these problems. Creating that symlink shouldn't cause any harm on
non-Linux platforms. It might be a bit ugly, bit it's probably less ugly
than patching around in the yices git submodule itself.

### DIFF
--- a/src/yices/Makefile
+++ b/src/yices/Makefile
@@ -13,6 +13,8 @@ install:
 	ln -fsn $(VERSION)/include
 	ln -fsn $(VERSION)/lib
 	ln -fsn $(VERSION)/include_hs
+	# add missing symlink libyices.so.2.6{->2}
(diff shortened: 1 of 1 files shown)


quark17:
I was wondering if maybe we should do this, but I already removed `LDCONFIG=true`.  I assume that it should be added back, with this change?

flokli:
Ah, I missed that commit. All good, in that case, I'll just set `LDCONFIG=ldconfig` downstream. No worries :-)

PR #45, opened 2020-02-10
Quell unused top level bindings warning by... using them.

bpfoley:
Not technically necessary, but I hate warnings, and as of GHC 8.8.1
there's no way of selectively disabling warnings for an individual
section of code.

### DIFF
--- a/src/comp/SimExpand.hs
+++ b/src/comp/SimExpand.hs
@@ -2009,7 +2009,12 @@ data UseInfo = UseInfo { user_name :: AId
                        }
 
 cvtARule :: ARule -> UseInfo
-cvtARule (ARule rId _ _ _ rPred rActs _ rOrig) = UseInfo rId [rPred] [] rActs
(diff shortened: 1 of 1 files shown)


quark17:
`PragmaCheck.hs` also has a non-exported data type with named fields where two of the field names are never used (`api_prefix` and `api_given_name`).  In GHC 8.8.2 (and earlier versions), this would also trigger a warning:

`PragmaCheck.hs:378:11: warning: [-Wunused-top-binds]`
`    Defined but not used: ‘api_prefix’`
`PragmaCheck.hs:379:11: warning: [-Wunused-top-binds]`
`    Defined but not used: ‘api_given_name’`

But in 8.4.4, 8.6.5, and 8.8.2, I don't see it.  And presumably you don't or you would have submitted a PR to fix that too? :)

Do you know why GHC has stopped warning about those, but continues to warn about the ones in SimExpand?

PR #46, opened 2020-02-11
Fix BSC compile when using Clang.

bpfoley:
Update gcc-script.sh to support clang.

Use script as GHC assembler, linker, and pre-processor as well as just
the compiler. This is necessary for shared libraries to link properly
with ghc when invoed with make CC=clang-6.0 CXX=clang++-6.0


arjenroodselaar:
Is anyone still building with a 3.x.x of GCC? Unless I underestimate how long the commercial release is supported on old platforms it is probably a good opportunity to clean this up. @quark17 thoughts?

quark17:
I had noticed this recently, but decided to leave the script there, in case anyone wanted to use it as a place to add special support for other compilers or versions.  We don't need to support GCC 3 any more.  But I also didn't feel like it was hurting anything, so I didn't bother to remove that part of the script yet.

PR #53, opened 2020-02-12
Vagrant build option

cbiffle:
This provides a mostly-reproduceable build environment that sets up an Ubuntu LTS image, installs the required packages, and can be used to build the toolchain by running:

```
$ cd util/vagrant
$ vagrant up
$ vagrant ssh
  # in ssh session
$ cd /vagrant && make all
```

Tested on various Linux distros and Mac. (Obviously the produced binaries run only on Linux.)

We're planning on using this internally to do consistent bsc builds, and if it's not general enough to upstream, that's fine -- but I wanted to offer it.

### DIFF
--- a/util/vagrant/Vagrantfile
+++ b/util/vagrant/Vagrantfile
@@ -8,16 +8,16 @@ Vagrant.configure("2") do |config|
   #
   # Note that we are *deliberately* exposing the .git subdirectory to the VM.
   # The build process needs access to it to generate version strings.
-  config.vm.synced_folder ".", "/vagrant", type: "rsync",
+  config.vm.synced_folder "../..", "/vagrant", type: "rsync",
     rsync__args: ["--verbose", "--archive", "-zz", "--copy-links"],
-    rsync__exclude: ["vagrant-out"]
+    rsync__exclude: ["../../vagrant-out"]
 
-  config.vm.synced_folder "vagrant-out", "/vagrant/inst", create: true
+  config.vm.synced_folder "../../vagrant-out", "/vagrant/inst", create: true
 
   config.vm.provider "virtualbox" do |vb|
     vb.memory = 4096
     vb.cpus = 4
   end
 
-  config.vm.provision :shell, path: "util/vagrant/provision.sh"
+  config.vm.provision :shell, path: "./provision.sh"
 end
--- a/util/vagrant/Vagrantfile
+++ b/util/vagrant/Vagrantfile
@@ -1,5 +1,5 @@
 Vagrant.configure("2") do |config|
-  config.vm.box = "hashicorp/bionic64"
+  config.vm.box = "bento/ubuntu-18.04"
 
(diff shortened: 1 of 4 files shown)


util/vagrant/Vagrantfile
```diff
@@ -0,0 +1,23 @@
+Vagrant.configure("2") do |config|
+  config.vm.box = "hashicorp/bionic64"
```
arjenroodselaar: Lets use "bento/ubuntu-18.04" for those of us who prefer a different Hypervisor :)

util/vagrant/provision.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
flokli:
```suggestion
#!/usr/bin/env bash
```

util/vagrant/Vagrantfile
flokli:
```suggestion
  config.vm.box = "bento/ubuntu-18.04"
```

:+1: on @arjenroodselaar 's suggestion.

util/vagrant/Vagrantfile
```diff
@@ -0,0 +1,23 @@
...
+  #
+  # Note that we are *deliberately* exposing the .git subdirectory to the VM.
+  # The build process needs access to it to generate version strings.
+  config.vm.synced_folder "../..", "/vagrant", type: "rsync",
+    rsync__args: ["--verbose", "--archive", "-zz", "--copy-links"],
+    rsync__exclude: ["../../vagrant-out"]
+
+  config.vm.synced_folder "../../vagrant-out", "/vagrant/inst", create: true
+
+  config.vm.provider "virtualbox" do |vb|
```
flokli: This obviously would need an additional section for other hypervisors.

util/vagrant/provision.sh
```diff
@@ -0,0 +1,20 @@
...
+    ghc \
+    git \
+    gperf \
+    iverilog \
+    libfontconfig1-dev \
+    libghc-old-time-dev \
+    libghc-regex-compat-dev \
+    libghc-syb-dev \
+    libx11-dev \
+    libxft-dev
```
flokli: This whole script looks more like a generic "install build dependencies for a ubuntu 18.04 box" - we probably shouldn't make it vagrant specific, but include it in a shared folder, that'll also be used for CI.

quark17:
I don't feel qualified to judge the general-ness of this, but we're in the early stages, so I am happy to accept contributions like this and we can always change later.  But I'd like to get consensus on the content, so I'll wait for @cbiffle 's reactions to the suggested edits.

quark17:
PR #56 added a similar provision script under `.github/`.  We might consider moving it to a shared location.

cbiffle:
I've accepted @flokli's changes. I agree that the two provisioning scripts should probably be merged.

PR #55, opened 2020-02-14
Remove Libs/HO, and move the one used combinator to a module.

bpfoley:


src/comp/IConv.hs
```diff
@@ -401,6 +400,9 @@ iAddWhen errh flags r env pvs qs ite =
         (i, t, e') -> (i, t, underLAM t e' [] $ \ t e -> iAps icPrimWhen [t] [p, bindFn e])
 
 
+uncurry3 :: (a -> b -> c -> d) -> (a, b, c) -> d
```
quark17: It feels almost too minor to mention, but the other utility functions are defined at the end of the file, so I'd tend to put it there.  Many modules in `src/comp/` define a single stage of the BSC compilation (as used in `bsc.hs`) and usually they export one entry point which is the first thing defined in the file.  I'm not particularly attached to that convention, but it's another reason to not put the utility function first.

PR #64, opened 2020-02-15
Remove unneeded type constraints (found by ghc -Wredundant-constraints)

bpfoley:
(A warning added in GHC 8.0 which I'd like to enable soon)

### DIFF
--- a/src/comp/BDD.hs
+++ b/src/comp/BDD.hs
@@ -29,11 +29,11 @@ bddFalse, bddTrue :: BDD a
 bddFalse = L False
 bddTrue  = L True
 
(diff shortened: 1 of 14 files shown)


quark17:
Whether some of these are redundant depends on the version of the libraries / GHC.  The CI build for Ubuntu 16.04 failed because it uses GHC 7.10.3, where presumably `denominator` still has `Integral` as a constraint.

bpfoley:
Absolutely. So I'm discovering, and I'm trying to get multiple versions of GHC set up on my work machine so I don't trip over this again...

PR #70, opened 2020-02-20
Add type signatures to a bunch of top level definitions

bpfoley:
This is incomplete, but will allow us to use the GHC option
-Wmissing-exported-signatures (introduced in GHC 8.0.1) to make sure
this doesn't regress.

I'd like this because the types serve as good documentation. Often
the type signature and the implementation complement each other well
when trying to understand the code.

Eg:

makeResultCStmts :: Position -> [Id] -> CExpr -> [IdProp]
                 -> SEM [EMsg] ISConvState [CStmt]
makeResultCStmts pos updatedVars inside props =
    do let resultId = addIdProps (idTheResult pos) props
           patVars = pMkTuple pos $ map CPVar updatedVars
           processDeclInfo (Just (Just t, [])) = Just t
           processDeclInfo _ = Nothing
       maybeRetType <- mapM getDeclInfo updatedVars >>=
                       (return . fmap (tMkTuple pos) .
                        sequence . map processDeclInfo)

### DIFF
--- a/src/comp/ABin.hs
+++ b/src/comp/ABin.hs
@@ -211,4 +211,3 @@ abemi_rule_relation_db m =
       Right r -> Just (asi_rule_relation_db (abmi_aschedinfo r))
 
 -- ===============
-
--- a/src/comp/GenABin.hs
+++ b/src/comp/GenABin.hs
@@ -857,4 +857,3 @@ instance Bin VFunction where
                    return (VFunction i r d s)
 
 -- ----------
-
--- a/src/comp/GenForeign.hs
+++ b/src/comp/GenForeign.hs
@@ -100,4 +100,3 @@ extractForeignFuncInfo defn =
                   Left _ -> internalError "GenForeign: getName"
             res -> internalError ("GenForeign: foreigns = " ++
                                   ppReadable res)
-
--- a/src/comp/GraphMap.hs
+++ b/src/comp/GraphMap.hs
@@ -98,6 +98,7 @@ instance (PPrint v, PPrint w) => PPrint (GraphMap v w) where
                                     | (v1, m') <- M.toList m]
 
 -- non-connected components
+ncc :: Ord v => GraphMap v w -> [[v]]
 ncc g | null g = [[]]
       | otherwise = ncvs:(ncc $ deleteVertices g ncvs)
       where (v:vs) = vertices g
--- a/src/comp/TCMisc.hs
+++ b/src/comp/TCMisc.hs
@@ -869,6 +869,8 @@ eqToVPred poss num_eq = do
   p <- eqToPred num_eq
   mkVPredFromPred poss p
 
+unifyNoEq :: (PPrint a, PVPrint a, HasPosition a)
+          => [Char] -> a -> Type -> Type -> TI ()
 unifyNoEq loc x t1 t2 = do
   ps <- unify x t1 t2
   when (not $ null ps) $
@@ -980,6 +982,8 @@ defaultUnifyError x t1 t2 =
 -----
 
 -- unify e td (a `fn` rt)
+unifyFnFrom :: (HasPosition a, HasPosition b, PPrint b, PVPrint b)
+            => a -> b -> Type -> Type -> TI (Type, [VPred])
 unifyFnFrom x e (TAp (TAp arr a) r) rt | arr == tArrow = do
     ps <- unify e r rt
     return (a, ps)
@@ -997,6 +1001,8 @@ unifyFnFrom x e t rt = do
     return (v, ps)
 
 -- unify e td (at `fn` r)
+unifyFnTo :: (HasPosition a, HasPosition b, PPrint b, PVPrint b)
+          => a -> b -> Type -> Type -> TI (Type, [VPred])
 unifyFnTo x e (TAp (TAp arr a) r) at | arr == tArrow = do
     ps <- unify e a at
     return (r, ps)
@@ -1017,6 +1023,8 @@ unifyFnTo x e t at = do
 -- and create new type variables only if it doesn't
 -- we never do anything that might generate numeric equalities
 -- unify e td (a `fn` r)
+unifyFnFromTo :: (HasPosition a, HasPosition b, PPrint b, PVPrint b)
+              => a -> b -> Type -> TI (Type, Type)
 unifyFnFromTo x e (TAp (TAp arr a) r) | arr == tArrow = do
     return (a, r)
 unifyFnFromTo x e t = do
--- a/src/comp/Type.hs
+++ b/src/comp/Type.hs
@@ -1,6 +1,6 @@
 module Type where
 import ErrorUtil(internalError)
-import Position(noPosition)
+import Position(Position, noPosition)
 import PreIds
 import PPrint(ppReadable)
 import CType(Type(..), TyVar(..), TyCon(..), TISort(..), Kind(..), StructSubType(..), cTNum)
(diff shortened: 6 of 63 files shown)


src/comp/ABin.hs
```diff
@@ -33,7 +33,7 @@ import Verilog(VProgram)
 
 
 -- ===============
-
+id_ABin :: String
```
quark17: I thought I had removed all the RCS $Id$ declarations, but it looks like four are still in.  I was going to remove them, which would conflict for you, or you can remove them here?

src/comp/GenABin.hs
```diff
@@ -31,6 +31,7 @@ import qualified Data.Map as M
 -- import Util(traceM)
 -- import Trace
 
+id_GenABin :: String
```
quark17: Remove this RCS $Id$ declaration

src/comp/GenForeign.hs
```diff
@@ -16,6 +16,7 @@ import TopUtils(putStrLnF)
 
 import Control.Monad(unless)
 
+id_GenForeign :: String
```
quark17: Remove this RCS $Id$ declaration

src/comp/GraphMap.hs
```diff
@@ -98,6 +98,7 @@ instance (PPrint v, PPrint w) => PPrint (GraphMap v w) where
                                     | (v1, m') <- M.toList m]
 
 -- non-connected components
+ncc :: Ord a1 => GraphMap a1 a2 -> [[a1]]
```
quark17: Change this to use better names, consistent with the other type declarations in this file:
`Ord v => GraphMap v w -> [[v]]`

src/comp/Type.hs
```diff
@@ -105,21 +146,27 @@ arrow a r = TAp (TAp tArrow a) r
...
 isActionValue (TAp av _) = av == tActionValue
 isActionValue _ = False
 
+getAVType :: Type -> Type
 getAVType (TAp av t) | av == tActionValue = t
 getAVType t = internalError("getAVType not ActionValue: " ++ ppReadable t)
 
-isActionWithoutValue (TAp av (TCon (TyNum 0 _))) = (av == tActionValue_)
+isActionWithoutValue :: Type -> Bool
+isActionWithoutValue (TAp av (TCon (TyNum 0 _))) = av == tActionValue_
```
quark17: FYI, situations like this (when the RHS is an expression with `==`) is where I think that parentheses help to make it clear what part is to the right of the `=`.  But I don't feel strongly enough to reject this change.

src/comp/TypeAnalysis.hs
```diff
@@ -373,18 +373,20 @@ getBitWidth flags symtab t =
...
 
+kVector, kList :: Kind
 kVector = Kfun KNum (Kfun KStar KStar)
 kList   = Kfun KStar KStar
 
+vsVector, vsList :: [Id]
 vsVector = map mk_homeless_id ["vsize", "element_type"]
 vsList   = map mk_homeless_id ["element_type"]
-vsTuple sz = if (sz > 8)
+vsTuple sz = if sz > 8
```
quark17: I also am a fan of using parens on the conditions of if-expressions, to make it pop visually.  Perhaps this is less an issue when you have good code highlighting.  (I'll also say that I'm partially colorblind.)  Again, I don't feel strongly enough to reject the change.

src/comp/TypeAnalysis.hs
```diff
@@ -373,18 +373,20 @@ getBitWidth flags symtab t =
...
              then internalError ("vsTuple: size > 7: " ++ ppReadable sz)
              else map mk_homeless_id (take sz ["a","b","c","d","e","f","g","h"])
 
 -- apply a function to the base of a qualified type
 apType :: (Type -> Type) -> Qual Type -> Qual Type
-apType f (ps :=> t) = (ps :=> f t)
+apType f (ps :=> t) = ps :=> f t
```
quark17: Another example of a RHS with an operator that contains an equal sign, which can be visually confusing without parentheses, in my opinion.

src/comp/Util.hs
```diff
@@ -407,14 +423,17 @@ makePairs (x:y:xs) = (x, y) : makePairs xs
 makePairs _ = internalError "Util.makePairs: failed"
 
 -- all pairs except pairs of an item with itself
+allPairs :: [b] -> [(b, b)]
```
quark17: Huh, why is `b` used here (and in the next few type declarations) instead if `a`?

src/comp/Util.hs
```diff
@@ -491,31 +510,52 @@ fromJustOrErr _   (Just v) = v
...
+fst2of4 (x,y,_,_) = (x,y)
+fst3of4 :: (a, b, c, d) -> (a, b, c)
 fst3of4 (x,y,z,_) = (x,y,z)
 
+fst4of5 :: (a, b, c, d, e) -> (a, b, c, d)
 fst4of5 (a,b,c,d,_) = (a,b,c,d)
+fifth5 :: (a, b, c, d, e) -> e
 fifth5 (_,_,_,_,x) = x
 
+ordPair :: Ord b => (b, b) -> (b, b)
```
quark17: Again, why `b` and not `a`?

src/comp/Util.hs
```diff
@@ -535,11 +575,12 @@ eqSnd (_,a) (_,b) = a == b
 
 -- =====
 -- Maybe utilities
-
+unJust :: Maybe p -> p
```
quark17: And why `p` here (and in the following prototype) and not `a`?

quark17:
Actually, I went ahead and pushed a change to remove the RCS $Id$ declarations.

src/comp/ABin.hs
bpfoley: OK. I'll remove them.

src/comp/ABin.hs
```diff
@@ -31,8 +31,6 @@ import Flags(Flags(..))
 import Verilog(VProgram)
 
 
--- ===============
```
quark17: Oops, can you leave this first delimiter in?

src/comp/Util.hs
```diff
@@ -535,11 +575,12 @@ eqSnd (_,a) (_,b) = a == b
 
 -- =====
 -- Maybe utilities
-
```
quark17: I would prefer to leave this space between the group header and the first function in the group.

src/comp/Util.hs
quark17: This one wasn't resolved?

bpfoley: Woops. Fixed.

bpfoley:
> Can we also turn on this flag in the Makefile, so that missing type declarations are checked by CI?

We can't yet because there are still some files (such as the parser and strings files) where most/all the top level definitions have no signatures, and it feels like a lot of overhead to add them in these cases.

I haven't yet decided what to do for these. Possibly add `{-# OPTIONS_GHC -fno-warn-missing-signature}` for these files.

src/comp/Util.hs
```diff
@@ -475,6 +494,7 @@ separate abs =
 
 
 -- =====
+
```
quark17: This isn't the line where I asked for a space, sorry!  The convention in this file is: group delimiter, title, then a blank line before the functions in that group.  Here, there should be no blank between the delimiter and the title.  It's further down in the file, where there's a title an no blank line after it..

src/comp/Util.hs
quark17: This is where the blank link needs to be put back in, sorry.

src/comp/Util.hs
bpfoley: No worries. It's good to get these things right, and thanks for your patience.

src/comp/TCMisc.hs
```diff
@@ -980,6 +982,8 @@ defaultUnifyError x t1 t2 =
 -----
 
 -- unify e td (a `fn` rt)
+unifyFnFrom :: (PVPrint a, HasPosition a, HasPosition b, PPrint a)
```
quark17: If I were to nitpick more: In this function and the one below it, I'd make the base type be `a -> b -> ...` (and not `b -> a` merely because of the order they appear in the context).

bpfoley: Done.

PR #71, opened 2020-02-20
Remove a bunch of dead code

bpfoley:
diffstat for this says
``41 files changed, 237 insertions(+), 1260 deletions(-)``

I hacked together a stack build for bluespec, and used Neil Mitchell's [weeder](https://github.com/ndmitchell/weeder) to help find dead code.

I left anything that seemed to be for debugging purposes, or was otherwise
'there for a reason' alone, but it's probable I've missed some cases.

Split into individual patches to make it easier to review/cherry pick.


quark17:
Many of the files in `src/comp/` are libraries; they define a set of useful functions, often on a data type.  Just because a function from that set is not currently being used doesn't mean that we should remove it, because maybe someday it will be needed.

A number of your changes are removing this kind of code, and I would prefer we didn't.  I'll try to comment on them individually.

(FYI: Some of these libraries might be better removed and replaced with something else.  For example, the BDD module and BExpr module -- and also BoolOpt -- are old home-grown or ad hoc attempts at boolean optimization, that are probably not particularly powerful nor efficient, and might be better replaced with code that uses the available SMT solver, now that one's available.)

bpfoley:
Thanks. I had a feeling that would be the case, so please push back on anything you'd like to stay.
The idea was to try to trim down some of the utilities (ListMap & co come to mind) to see if what remained could be replaced by a base/popular Haskell package.

src/comp/SCC.hs
```diff
@@ -1,4 +1,4 @@
-module SCC(scc, getCycles, tsort, Graph) where
+module SCC(scc, tsort, Graph) where
```
quark17: This is a library.

I'm a little surprised that `getCycles` isn't used, but maybe more places are using other modules now.  (For example, `GraphWrapper`, based on `Data.Graph`.)

src/comp/SEMonad.hs
```diff
@@ -1,5 +1,5 @@
 {-# LANGUAGE CPP #-}
-module SEMonad where
+module SEMonad(SEM(..), err, run) where
```
quark17: This is a library.  But also, it's now only used in one place (the BSV parser) and I'd guess that the other uses were replaced with (composed) standard monads, so maybe this could go away if the BSV parser is changed to use standard monads?

src/comp/SimCCBlock.hs
```diff
@@ -39,7 +39,6 @@ module SimCCBlock( SBId
                  , pfxPort
                  , pfxMeth
                  , renameIds
-                 , extractDefs
```
quark17: While this module is kind of a library, removing `extractDefs` is OK.  I see that, even in the commit where it was added, the code was never used.  (It may have been added while attempting to fix a bug, but the final fix didn't end up needing it, but it was checked in anyway.)

src/comp/Subst.hs
```diff
@@ -343,28 +289,3 @@ trimSubstByVars :: [TyVar] -> Subst -> Subst
 trimSubstByVars vs = if (null vs) then id else trimSubstBy filterFunc
   where filterFunc v t = not ((elem v vs) ||
                               (any (\u -> elem u vs) (tv t)))
-
-{-
--- This removes variables just from the LHS of the substitution,
```
quark17: It looks like `removeFromSubst` was used when debugging `apSub` in `CSyntaxTypes`.  There's commented out code that has been used for figuring out why something went wrong.  I dunno if that code will ever be needed again, but it might not hurt to leave this function?

src/comp/TIMonad.hs
```diff
@@ -656,19 +650,6 @@ updAssumpPos :: Id -> Assump -> Assump
 updAssumpPos i (i' :>: s) = setIdPosition (getIdPosition i) i' :>: s
 
 
-incrementFundepStack :: TI ()
-incrementFundepStack = do
-  k <- gets tsFundepStack
-  flags <- getFlags
-  if (k > (maxTIStackDepth flags))
-     then (err(noPosition, ETypeFundepStackOverflow))
```
quark17: If this code is removed, then `ETypeFundepStackOverflow` is not used.  We should edit `Error.hs` to comment out this tag in the data type and then block-comment the case-arm for it with a comment that it's retired.  (See how other error codes have been retired in that file, by searching for `retired` and `obsolete`, although there's not a lot of consistency in how they've been done.)

src/comp/Util.hs
```diff
@@ -417,10 +342,6 @@ uniquePairs (x:ys) = [(x,y) | y <- ys] ++ uniquePairs ys
 -- all pairs of an item with itself
 selfPairs xs = [(x, x) | x <- xs ]
 
-mapFstM :: (Monad m) => (a -> m c) -> [(a, b)] -> m [(c, b)]
```
quark17: Probably doesn't hurt to leave `mapFstM` in this library?

src/comp/Util.hs
```diff
@@ -553,10 +430,6 @@ fromMaybeM def m = do
 -- =====
 -- Either utilities
 
-apLeft :: (a -> c) -> Either a b -> Either c b
```
quark17: Might not hurt to leave `apLeft` in the library?

quark17:
Is there a way to add dead-code checks to the CI, so that new dead code isn't introduced?

src/comp/IExpandUtils.hs
```diff
@@ -22,7 +22,7 @@ module IExpandUtils(
         addStateVar, step, updHeap, getHeap, {- filterHeapPtrs, -}
         getSymTab, getDefEnv, getFlags, getCross, getErrHandle, getModuleName,
         getNewRuleSuffix, updNewRuleSuffix,
-        updatePExprPosition, mapPExprPosition,
```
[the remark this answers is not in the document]
bpfoley: That's interesting. Thanks for the context.

src/comp/Libs/Hash.hs
```diff
@@ -1,7 +1,7 @@
 -- Copyright (c) 1982-1999 Lennart Augustsson, Thomas Johnsson
```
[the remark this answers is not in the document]
bpfoley: Nice! This was exactly the sort of thing I was hoping might happen. :)

And yeah, it looks like Weeder doesn't check for typeclasses that have instances that are declared but never used.

src/comp/Libs/ListMap.hs
```diff
@@ -3,19 +3,10 @@
 --
```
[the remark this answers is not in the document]
bpfoley: Right. I'll try to do this as a follow up PR.

src/comp/Libs/ListSet.hs
```diff
@@ -17,36 +17,9 @@ singleton x = [x]
 union :: (Eq a) => [a] -> [a] -> [a]
```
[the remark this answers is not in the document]
bpfoley: Agreed. I'll investigate in a follow-up PR

src/comp/Libs/ListUtil.hs
```diff
@@ -12,42 +12,9 @@ unfoldr f p x | p x       = []
 chopList :: ([a] -> (b, [a])) -> [a] -> [b]
```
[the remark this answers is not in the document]
bpfoley: Yup. This all looks very 'standard-ish' to me too. I'll see if I can find a replacement in a followup PR.

src/comp/Pred.hs
```diff
@@ -4,7 +4,7 @@ module Pred(
 	    Qual(..), PredWithPositions(..), Pred(..), Class(..), Inst(..),
             getInsts,
 	    removePredPositions, getPredPositions, addPredPositions, mkPredWithPositions,
-	    isIn, expandSyn, predToType, qualToType, mkInst,
+	    expandSyn, predToType, qualToType, mkInst,
```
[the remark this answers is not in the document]
bpfoley: Done.

bpfoley:
> Is there a way to add dead-code checks to the CI, so that new dead code isn't introduced?

Currently, no. Weeder relies on being able to build everything using stack, although it would be possible to do the integration work required to invoke `ghc --make` with `-ddump-to-file -ddump-hi` and to pipe this information to weeder.

I'll keep this in mind, because I'd like dead-code CI checks too.

src/comp/SEMonad.hs
bpfoley: OK. I'll try to do this as a followup PR.

PR #73, opened 2020-02-23
MacOS build improvements for "make STP_STUB=1".

ret:


src/comp/wrapper.sh
```diff
@@ -15,8 +15,8 @@ if [ -n "$BLUESPEC_LD_LIBRARY_PATH" ] ; then
     LD_LIBRARY_PATH=${BLUESPEC_LD_LIBRARY_PATH}:${LD_LIBRARY_PATH}
     DYLD_LIBRARY_PATH=${BLUESPEC_LD_LIBRARY_PATH}:${DYLD_LIBRARY_PATH}
 fi
-LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${BLUESPECDIR}/SAT
-DYLD_LIBRARY_PATH=${DYLD_LIBRARY_PATH}:${BLUESPECDIR}/SAT
+LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${BLUESPECDIR}/SAT:${BLUESPECDIR}/VPI
```
quark17: Why do you need to add VPI here?  If there's a place that uses VPI where it's not in the path, we should fix it at that use.  I'd prefer to eventually remove this wrapper, not add to it.

Makefile
```diff
@@ -4,6 +4,8 @@ TOP := $(PWD)
 PREFIX   ?= $(TOP)/inst
 BUILDDIR ?= $(TOP)/build
 
+include $(TOP)/platform.mk
```
quark17: I have no objection to including `platform.mk` everywhere, just to be safe; but since nothing in this file uses it, there's no need to include it.  Unless this masks the issue that a later Makefile isn't including it?

Makefile
```diff
@@ -31,7 +33,7 @@ install:
 # in $PATH. it's not enough to just set bsc...
 .PHONY: check
 check:
-	@(export PATH=$(PREFIX)/bin:$(PATH); $(MAKE) -C examples/smoke_test smoke_test)
+	@(export PATH=$(PREFIX)/bin:"$(PATH)"; $(MAKE) -C examples/smoke_test smoke_test)
```
quark17: Is it better to put the quotes around the entire new path?

src/comp/wrapper.sh
ret: Hi! First of all, great experience on Ubuntu. Was tempted to use OSX, but I was wavering on the way ;-). Works on Mojave and Catalina now.
I needed to add the VPI .so path on Mojave 10.14.6. On Catalina 10.15.3 it wasn't needed (I don't understand why). On Mojave, if I don't have the VPI lib in the lib path I get:
```
$ bsc
dyld: Library not loaded: libbdpi.so
  Referenced from: /Users/reto/bin/inst/bin/core/bsc
  Reason: image not found
Abort trap: 6
```
On OSX, I'm using brew to get things like linuxbrew/xorg/libx11, autoconf, coreutils etc. GHC 8.8.2 installed from the usual compiler download (i.e. not using 'stack') and then  'cabal install' the 3 older packages per readme.

Makefile
ret: You're right, that would be even better!

Makefile
ret: I added it only because the VPI build needed the OS set to Darwin to use the right linker option.  I should probably figure out _why_ bsc pulls in the VPI library in the first place, right?

ret:
Retracting this PR to see if I can first get to the bottom of these OSX version specific library path differences.

quark17:
PR #98 has the MacOS build working (with real STP, not the stub).  I'd like to accept those changes, if there's consensus, so comments on that would be appreciated.

FYI, the `libbdpi` issue was due to a mistake in the STP stub Makefile; I had copied it from the `VPI` Makefile but had not replaced all occurrences of `libbdpi` with `libstp`.  That has been fixed in commit 23ca02a (from PR #88 ).

PR #74, opened 2020-02-23
Update GHC/cabal instructions

hasufell:
Reason for this PR is a user on `#haskell` IRC asking for help, because more recent cabal version don't support the installation method in the README as-is.

---

Wrt GHC/cabal installation: ubuntu/debian packages are out of date.
So are most other distros GHC/cabal packages. Some (e.g. Arch Linux)
even ship half-broken GHC. hvr's ppa is the recommended one for
debianoid distros. For others (including Mac) ghcup is a safe choice
and installs in $HOME/.ghcup without any sudo rights.

Since cabal-3, the semantics for 'cabal install' have changed. Cabal
now uses the new-install (nix-style) by default. This requires some
more steps to work with this build system (such as creating a package-env
file and telling GHC where to look for it).

The simplest solution for now is to use 'cabal v1-install', which
serves the legacy way of installing packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,24 +77,21 @@ The core of BSC is written in Haskell, with some libraries in C/C++.
 You will need the standard Haskell compiler `ghc` which is available for Linux,
 MacOS and Windows, along with some additional Haskell libraries. These are
 available as standard packages in most Linux distributions. For example, on
-Debian and Ubuntu systems, you can say:
+Debian and Ubuntu systems, you should use hvr's PPA for an up-to-date toolchain:
 
-    $ apt-get install ghc
-    $ apt-get install \
-        libghc-regex-compat-dev \
-        libghc-syb-dev \
-        libghc-old-time-dev
+    $ add-apt-repository ppa:hvr/ghc
+    $ apt-get update
 
-The second command will install the Haskell libraries `regex-compat`, `syb`,
-and `old-time`, as well as some libraries that they depend on.
+Then install the appropriate ghc and cabal:
 
-You can do the analogous package-install on other Linux distributions using
-their native package mechanisms, and use Macports on Apple OS X. Full details
-can be found at <https://www.haskell.org/>. On some systems, you may need to
-use the `cabal` command to install Haskell libraries:
+    $ apt-get install ghc-8.8.2
+    $ apt-get install cabal-install-3.0
 
-    $ apt-get install cabal-install
-    $ cabal install regex-compat syb old-time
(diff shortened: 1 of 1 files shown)


hasufell:
No input, no action. Abandoning.

quark17:
I appreciate that you submitted this, and I apologize for not commenting on it.  In early March, I had an offline discussion about improving the README, based on feedback that included the changes proposed here (thanks!).  Part of the difficulty was wanting to give full details while also keeping things simple, but also uncertainty about whether any changes to the README would become moot by impending improvements (that still haven't happened) or decisions around what versions and platforms we support etc.  And also, there was a hope that we would have by now tagged an initial BSC 4.0 version, that could be available as pre-build download and also put in packages for various OSes, so that the "simple" case would be to just download it and the compile instructions could be written for just a developer audience.

At the time, I thought that the following two additions to the README could be a reasonable compromise:
1. After the Debian/Ubuntu cabal example is shown, mention that newer versions of cabal have a different interface and that`v1-install` can be used for backwards compatibility.  That should be enough clue for people to know where to look, if that's the issue they encounter.  [The compiler makefile could alternatively be made to support the v2 interface, if GHC is told where to find the package databases, such as with GHC's `user-package-db` flag.]
2. After the discussion of GHC versions, mention how the reader could get a newer/latest version of GHC or cabal (or if they're on a different OS).  This could be done by mentioning hvr's PPA (as you suggest) but it could also be done by linking to the `ghcup` webpage (which gives a single `curl` command for all OSes) or by some other means.  (Both options might be a bit scary looking, to users not familiar with those sources.)

A counter proposal was to have the README describe an "easy" way of doing a build by using ghcup, cabal 3, etc -- with fewer decisions left to the user and installing the most modern tools (which will also keep developers happier) -- but that there could be a link at the top to READMEs for how to more natively build on other OSes.  (We'd keep our github CI using tools and packages from Ubuntu/Homebrew, so these other READMEs would just be documenting what we're already doing.)

It was suggested that: one of the sources of complexity in a README is the number of decisions that the user has to make, so the more we can avoid that, the better.  However, having the first decision be "what OS are you running?" is a sufficiently simple question to ask, so it could be reasonable to branch from there to a set of simple instructions per OS.

In any case, thank you.  This PR won't be accepted as-is, so closing it is fine.  I kept it open as a reminder of the issue, but it'd be worth having an actual issue number for that.  Perhaps issue #109 could be used for that, since I think every other part of that has been resolved; but it might be better to close that issue and open a new one, referencing the discussion here.

PR #77, opened 2020-02-26
Add another large batch of top level signatures

bpfoley:
Getting close to be able to switch on the compiler warnings.

### DIFF
--- a/src/Parsec/ParsecPrim.hs
+++ b/src/Parsec/ParsecPrim.hs
@@ -137,6 +137,8 @@ setParserState st   = updateParserState (const st)
 type Parser a           = GenParser Char () a
 
 newtype GenParser tok st a = Parser (State tok st -> IO (Consumed (Reply tok st a)))
+
+runP :: GenParser tok st a -> State tok st -> IO (Consumed (Reply tok st a))
 runP (Parser p)            = p
 
 data Consumed a         = Consumed a                --input is consumed
@@ -192,6 +194,7 @@ ioToParser computation =
     Parser (\state -> do result <- computation
                          return (Empty (Ok result state [])))
 
+parserReply :: Consumed a -> a
 parserReply result = case result of
         Consumed reply -> reply
         Empty reply    -> reply
@@ -277,6 +280,7 @@ parsecBind (Parser p) f
                               other        -> return other
                      Error evil err1  -> return (Empty (Error evil err1)))
 
+mergeErrorReply :: [EMsg] -> Reply tok st a -> Reply tok st a
 mergeErrorReply err1 reply
   = case reply of
       Ok x state err2 -> Ok x state (mergeErrors err1 err2)
@@ -299,6 +303,7 @@ mergeErrors' (True, rrec) errs errs' =
     foldl (mergeErrors' (False, rrec)) errs [[e] | e <- sortBy cmpEMsg errs]
 mergeErrors' _ errs errs' = sortBy cmpEMsg (errs ++ errs')
 
+cmpEMsg :: EMsg -> EMsg -> Ordering
 cmpEMsg (pos1, _) (pos2, _) = compare pos1 pos2
 
 -- combine error messages
@@ -566,5 +571,7 @@ tokens shows nextposs s
 
        in walk1 s input)
 
+errSyntax :: Position -> String -> [String] -> EMsg
 errSyntax pos tok toks = (pos, ESyntax tok (nub $ filter (not . null) toks))
+errSyntaxEof :: Position -> [String] -> EMsg
 errSyntaxEof pos toks = (pos, ESyntax "end of file" (nub $ filter (not . null) toks))
--- a/src/comp/ACheck.hs
+++ b/src/comp/ACheck.hs
@@ -314,26 +314,36 @@ chkAReset (AReset { areset_wire = wire }) = isBit1 (aType wire)
 chkAInout :: AInout -> Bool
 chkAInout (AInout { ainout_wire = wire }) = isInout_ (aType wire)
 
+isBit :: AType -> Bool
 isBit (ATBit _) = True
 isBit _ = False
 
+isBit1 :: AType -> Bool
 isBit1 t = isBit t && (aSize t == 1)
 
+isString :: AType -> Bool
 isString (ATString _) = True
 isString _ = False
 
+isReal :: AType -> Bool
 isReal (ATReal) = True
 isReal _ = False
 
+isClock :: AType -> Bool
 isClock t = t == aTClock
 
+isReset :: AType -> Bool
 isReset t = t == aTReset
 
+isInout_ :: AType -> Bool
 isInout_ (ATAbstract i _) = i == idInout_
 isInout_ _ = False
 
+isForeignArg :: AType -> Bool
 isForeignArg  e = isBit e || isString e || isReal e
+--isBitOrString :: AType -> Bool
 --isBitOrString e = isBit e || isString e
+isBitOrInout_ :: AType -> Bool
 isBitOrInout_ e = isBit e || isInout_ e
 
 -- compare 2 types considering string sizes
@@ -343,10 +353,13 @@ compatTypesWthStr (ATArray sz1 t1) (ATArray sz2 t2) = (sz1 == sz2) && (compatTyp
 compatTypesWthStr t1 t2                         = t1 == t2
 
 
+isRelOp :: PrimOp -> Bool
 isRelOp p = p `elem` [ PrimEQ, PrimULE, PrimULT, PrimSLE, PrimSLT, PrimEQ3 ]
 
+isShift :: PrimOp -> Bool
 isShift p = p `elem` [ PrimSL, PrimSRL, PrimSRA ]
 
+tracePP :: Show a => String -> a -> Bool -> Bool
 tracePP s x True = True
 tracePP s x False = trace ("acheck:" ++ s ++ "\n" ++ show x) False
 
@@ -460,9 +473,11 @@ chkMethAExpr (ATaskValue { }) = True
 chkMethAExpr _ = True
 
 -- Check the expressions in defs
+chkMethADef :: ADef -> Bool
 chkMethADef (ADef _ _ e _) = chkMethAExpr e
 
 -- Check the expressions in instantiations
+chkMethAVInst :: AVInst -> Bool
 chkMethAVInst aa =
     all chkMethAExpr (getPorts aa) &&
     all chkMethAExpr (getInouts aa) &&
@@ -470,8 +485,9 @@ chkMethAVInst aa =
     all chkMethAExpr (getResets aa)
 
 -- Check the expressions in foreign blocks
+chkMethAForeignBlock :: AForeignBlock -> Bool
 chkMethAForeignBlock (clks, fcalls) = all chkMethAForeignCall fcalls
 
+chkMethAForeignCall :: AForeignCall -> Bool
 chkMethAForeignCall (AForeignCall _ _ es _ rsts) =
     all chkMethAExpr (es ++ rsts)
-
--- a/src/comp/ASchedule.hs
+++ b/src/comp/ASchedule.hs
@@ -81,25 +81,42 @@ import PreIds(idErrorTask)
 -- Traces
 --
 
+trace_ncsets :: Bool
 trace_ncsets = "-trace-ncsets" `elem` progArgs
+trace_usemap :: Bool
 trace_usemap = "-trace-usemap" `elem` progArgs
+trace_mutatormap :: Bool
 trace_mutatormap = "-trace-mutatormap" `elem` progArgs
+trace_cfmap :: Bool
 trace_cfmap = "-trace-cfmap" `elem` progArgs
+trace_cfmaps :: Bool
 trace_cfmaps = "-trace-cfmaps" `elem` progArgs
+trace_pcmap :: Bool
 trace_pcmap = "-trace-pcmap" `elem` progArgs
+trace_pcmaps :: Bool
 trace_pcmaps = "-trace-pcmaps" `elem` progArgs
+trace_scmap :: Bool
 trace_scmap = "-trace-scmap" `elem` progArgs
+trace_scmaps :: Bool
 trace_scmaps = "-trace-scmaps" `elem` progArgs
+trace_scgraph :: Bool
 trace_scgraph = "-trace-scgraph" `elem` progArgs
+trace_seqgraph :: Bool
 trace_seqgraph = "-trace-seqgraph" `elem` progArgs
+trace_schedinfo :: Bool
 trace_schedinfo = "-trace-schedinfo" `elem` progArgs
+trace_sched_steps :: Bool
 trace_sched_steps = "-trace-sched-steps" `elem` progArgs
+trace_disjoint_tests :: Bool
 trace_disjoint_tests = "-trace-disjoint-tests" `elem` progArgs
+trace_defs :: Bool
 trace_defs           = "-trace-a-definitions" `elem` progArgs
+trace_no_urgency_edge :: Bool
 trace_no_urgency_edge = "-trace-no-urgency-edge" `elem` progArgs
+disable_urgency_warnings :: Bool
 disable_urgency_warnings = "-hack-disable-urgency-warnings" `elem` progArgs
 
-
+pp :: Doc -> String
 pp = pretty 78 78
 
 
@@ -325,9 +342,11 @@ getCSNId :: CSNode -> AId
 getCSNId (CSN_Sched i _) = i
 getCSNId (CSN_Exec  i _) = i
 
+isCSNSched :: CSNode -> Bool
 isCSNSched (CSN_Sched _ _) = True
 isCSNSched (CSN_Exec  _ _) = False
 
+isCSNExec :: CSNode -> Bool
 isCSNExec = not . isCSNSched
 
 -- convert to the external data type
@@ -1394,7 +1413,6 @@ aSchedule_step2 errh flags prefix pps urgency_pairs amod ( scConflictMap0
 --       in GHC. The split was done by duplicating some definitions and passing
 --       many values as a huge tuple between them. This was done simply because
 --       it was the easiest refactoring technique to apply.
-
 aSchedule_step3 errh flags prefix pps amod ( scConflictMap0
                                            , cfConflictMap0
                                            , scConflictMapFinal
@@ -1779,8 +1797,10 @@ scheduleEsposito cmap_sc method_names rs =
 --
 
 -- prettyprint a rule name, dropping RL_ prefix when appropriate
+pp_r :: Id -> Doc
 pp_r rule = text (pString_r rule)
 
+pString_r :: Id -> String
 pString_r rule =
     -- XXX use "dropRulePrefixId"?
     let drop_rule_pfx ('R':'L':'_':rest) = rest
@@ -3507,14 +3527,19 @@ verifyAssertionSr (_, rpragma) _ =
 
 -- --------------------
 
+assertErrorMessage :: (ARuleId, RulePragma) -> Maybe Doc -> EMsg
 assertErrorMessage (r,a) mreason =
     (getIdPosition r,
      ERuleAssertion (pfpString r) (getRulePragmaName a) mreason)
 
+errFireWhenEnabledBecauseBlocked :: (ARuleId, RulePragma) -> [ARuleId]
+                                 -> AScheduler -> EMsg
 errFireWhenEnabledBecauseBlocked a rs sch =
     errFireWhenEnabledBecause a rs sch
         "because it is blocked by rule"
 
+errFireWhenEnabledBecause :: (ARuleId, RulePragma) -> [ARuleId]
+                          -> AScheduler -> String -> EMsg
 errFireWhenEnabledBecause a rs sch because_string =
     let s = case rs of { [_] -> ""; _ -> "s" }
         reason = s2par (because_string ++ s) $$
@@ -3524,6 +3549,8 @@ errFireWhenEnabledBecause a rs sch because_string =
                  nest 2 (pPrint PDReadable 0 sch)
     in  assertErrorMessage a (Just reason)
 
+errCannotScheduleFirst :: (ARuleId, RulePragma) -> [CSNode] -> [CSNode]
+                       -> EMsg
 errCannotScheduleFirst a before_sched before_exec =
     let s_msg = describe "scheduled" before_sched
         e_msg = describe "executed" before_exec
@@ -3544,6 +3571,9 @@ errCannotScheduleFirst a before_sched before_exec =
                     ]
           in hdr $+$ (nest 2 (vsep lst))
 
+errCannotScheduleFirstViaMeth :: (ARuleId, RulePragma)
+                              -> [(MethodId, [Either ARuleId AMethodId])]
+                              -> EMsg
 errCannotScheduleFirstViaMeth a befores =
     let hdr = text ("because it uses methods which require rules to execute before it:")
         lst = map describe befores
@@ -3561,6 +3591,7 @@ errCannotScheduleFirstViaMeth a befores =
                ([],_)  -> mname <+> mdoc
                (_,_)   -> mname <+> rdoc <+> (text "and") <+> mdoc
 
+errCascadedDomainCrossing :: (ARuleId, RulePragma) -> [MethodId] -> EMsg
 errCascadedDomainCrossing a unsynced =
     let hdr = text ("because it uses methods which are themselves unsychronized domain crossings:")
         lst = map describe unsynced
@@ -3914,6 +3945,7 @@ drops' graph seen seenList skip root =
                        [(child, root, [CCycle (child:seenList)])
                             | child <- nasty]) good'
 
+errMissingRoot :: a
 errMissingRoot = internalError "ASchedule: root missing from rule graph (!)"
 
 warnCycleDrops :: AId -> [(ARuleId, ARuleId, [Conflicts])] -> ErrorMonad ()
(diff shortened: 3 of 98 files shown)


src/comp/TCMisc.hs
```diff
@@ -113,6 +122,7 @@ satisfyX dvs es ps = do
 
 -- break up preds into those affected by a substition and those not
 -- in order to loop more efficiently in satisfy'
+split_rs :: Types t => Subst -> [t] -> ([t], [t])
```
quark17: Consider replacing `t` with `VPred`

src/comp/TCMisc.hs
```diff
@@ -556,10 +567,12 @@ reducePred eps dvs (VPred w pp@(PredWithPositions pr@(IsIn c ts) pos)) = do
                     return $ Just ([], b, nullSubst, Nothing)
       else return r
 
+predUnify :: [TyVar] -> Pred -> Pred -> Bool
 predUnify bound_tyvars (IsIn c1 ts1) (IsIn c2 ts2)
     | c1 == c2 = isJust (mgu bound_tyvars ts1 ts2)
     | otherwise = False
 
+dvsSub :: a -> b -> b
```
quark17: Consider replacing `a` and `b` with `Subst` and `DVS`

src/comp/TCMisc.hs
```diff
@@ -630,6 +643,7 @@ lookfor bound_tyvars v@(VPred i pp) eps@(EPred e pr':ps) =
                            ppReadable (bound_tyvars, v, eps, inst_subst, fd_subst))
         Nothing -> lookfor bound_tyvars v ps
 
+commute :: [a] -> [a]
```
quark17: Consider replacing `a` with `Type`

src/comp/TCMisc.hs
```diff
@@ -974,7 +990,9 @@ reportUnifyError bound_vars x orig_t1 orig_t2 =
     in
         findMismatch 0 orig_t1 orig_t2
 
-
+defaultUnifyError :: (HasPosition a, PPrint a, PPrint b,
+                      PPrint c, PVPrint a, PVPrint b, PVPrint c) =>
+                      a -> b -> c -> TI d
 defaultUnifyError x t1 t2 =
```
quark17: Consider replacing `b` and `c` with `Type`

src/comp/VModInfo.hs
```diff
@@ -239,11 +240,14 @@ instance PPrint VArgInfo where
         ppMClk d mclk <+> ppMRst d mrst <>
         text ";"
 
+ppMClk :: PPrint a => PDetail -> Maybe a -> Doc
```
quark17: Consider changing `a` to `Id`

src/comp/VModInfo.hs
```diff
@@ -239,11 +240,14 @@ instance PPrint VArgInfo where
...
 ppMClk d mclk =
     let clk = case mclk of
                   Nothing -> text "no_clock"
                   Just c  -> pPrint d 0 c
     in  text "clocked_by (" <> clk <> text ")"
+
+ppMRst :: PPrint a => PDetail -> Maybe a -> Doc
```
quark17: Consider changing `a` to `Id`

src/comp/Verilog.hs
```diff
@@ -326,7 +330,7 @@ instance PPrint VMItem where
             pPrint d 0 inst_id $+$
             ppComment cs $+$
             pPrint d p stmt
-
```
quark17: Don't delete this line.  It separates the instance declaration from the utility function.

src/comp/Verilog.hs
```diff
@@ -326,7 +330,7 @@ instance PPrint VMItem where
...
+pv95params :: PPrint a => PDetail -> (Maybe String, a) -> Doc
```
quark17: Consider changing `a` to `VExpr`

src/comp/Verilog.hs
```diff
@@ -903,11 +907,13 @@ vIsValidIdent (c:cs) = (isLetterOrUnderscore c) &&
...
 
 -------
-
+ppLines :: PPrint a => PDetail -> [a] -> Doc
 ppLines d xs = foldr ($+$) empty (map (pPrint d 0) xs)
 
+ppLinesBy :: (a -> b -> Doc) -> a -> [b] -> Doc
 ppLinesBy f d xs = foldr ($+$) empty (map (f d) xs)
 
+labeledPPLines :: PPrint b => a -> PDetail -> [b] -> Doc
```
quark17: It's interesting that the `label` is unused.  In fact, that suggests that maybe all of this label-related code can be removed:

This function is used in the instances for `LabeledPPrint`, which seem to be doing nothing different from `PPrint`.  I suspect, then, that we can remove the function and the typeclasses and replaced calls to `labeledPPrint` with just calls too `pPrint`.  (It looks like this label code had been introduced long ago for XRef and did actually use the label back then, but it has been stripped away long since.)

src/comp/Verilog.hs
```diff
@@ -934,8 +940,10 @@ commaSepEmptyLine xs = foldr1 (nextLine) xs
               | (y == empty)                 = x
               | otherwise                    =  x <> text "," $+$ text "" $+$ y
 
+ppRange :: (PPrint a, PPrint b) => PDetail -> a -> b -> Doc
```
quark17: Consider replacing `a` and `b` with `VExpr`

src/comp/Verilog.hs
```diff
@@ -934,8 +940,10 @@ commaSepEmptyLine xs = foldr1 (nextLine) xs
...
 ppRange d a b = text "[" <> pPrint d 0 a <+> text ":" <+> pPrint d 0 b <> text "]"
 
+ppMRange :: (PPrint a, PPrint b) => PDetail -> Maybe (a, b) -> Doc
```
quark17: Consider replacing `(a, b)` with `VRange`

src/Parsec/ParsecPrim.hs
```diff
@@ -299,6 +303,7 @@ mergeErrors' (True, rrec) errs errs' =
     foldl (mergeErrors' (False, rrec)) errs [[e] | e <- sortBy cmpEMsg errs]
 mergeErrors' _ errs errs' = sortBy cmpEMsg (errs ++ errs')
 
+cmpEMsg :: Ord a => (a, b) -> (a, c) -> Ordering
```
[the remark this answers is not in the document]
bpfoley: Read the binding name, Brian! :) Done.

src/comp/ACheck.hs
```diff
@@ -460,18 +473,21 @@ chkMethAExpr (ATaskValue { }) = True
...
 -- Check the expressions in instantiations
+chkMethAVInst :: AVInst -> Bool
 chkMethAVInst aa =
     all chkMethAExpr (getPorts aa) &&
     all chkMethAExpr (getInouts aa) &&
     all chkMethAExpr (getClocks aa) &&
     all chkMethAExpr (getResets aa)
 
 -- Check the expressions in foreign blocks
+chkMethAForeignBlock :: (a, [AForeignCall]) -> Bool
```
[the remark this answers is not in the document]
bpfoley: Done. And elsewhere too.

src/comp/BinData.hs
```diff
@@ -538,10 +563,11 @@ instance Bin Integer where
                        else if b == endPos
                              then return n
                              else loop (n * baseI + toInteger (ord b))
-
```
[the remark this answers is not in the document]
bpfoley: It's mostly fatfingering on my part, not intentional.

src/comp/CSyntax.hs
```diff
@@ -203,7 +203,7 @@ data CExpr
         | CHasType CExpr CQType
         | Cif Position CExpr CExpr CExpr
         -- x[a]
-               | CSub Position CExpr CExpr
```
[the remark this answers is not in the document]
bpfoley: It was actually just an incorrectly expanded tab, so I de-dented it to line up.

src/comp/CSyntax.hs
```diff
@@ -1145,25 +1156,28 @@ instance PPrint CMStmt where
     pPrint d p (CMinterface e) = pPrint d p (cVApply (idReturn (getPosition e)) [e])
     pPrint d p (CMTupleInterface _ es) = text"(" <> sepList (map (pPrint d p) es) (text ",") <> text ")"
 
+prDump :: PPrint a => PDetail -> [a] -> Doc
```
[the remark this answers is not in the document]
bpfoley: ACK. Will put on my todo list.

src/comp/CVPrint.hs
```diff
@@ -337,14 +346,18 @@ newIdsn n = ("x" ++ itos n):(newIdsn (n+1))
 newIds :: [Doc]
 newIds = map t (newIdsn 1)
 
+isFn :: Type -> Bool
 isFn (TAp (TAp (TCon arr) _) _) = isTConArrow arr
 isFn _         = False
 
+ppUntypedId :: p -> Doc -> [Doc] -> Doc
```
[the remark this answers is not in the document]
bpfoley: That's a good point. Used bindings must not be on yet. I see another round of cleanups in our future!

src/comp/Verilog.hs
bpfoley: Will do as a followup PR.

src/comp/Deriving.hs
```diff
@@ -895,12 +917,15 @@ mkConv con coCon (x:xs) v (TAp (TAp (TCon (TyCon arr _ _)) a) r) | arr == idArro
 mkConv _ _ _ v t = \ e -> e
 
 -- generate errors for duplicate tag encodings
+duplicate_tag_encoding_errors :: Id -> [CInternalSummand] -> [(Position, ErrMsg)]
```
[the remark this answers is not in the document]
bpfoley: Done here and elsewhere.

src/comp/Libs/Parse.hs
```diff
@@ -72,6 +76,7 @@ processAlts f rs (None False f':rws) = processAlts (bestFailAt f f') rs rws
 processAlts f rs (One b k as f':rws) = processAlts (bestFailAt f f') (rs++[(b,k,as)]) rws
 processAlts f rs (Many rs' f'  :rws) = processAlts (bestFailAt f f') (rs++rs') rws
 
+doMany :: (t -> b) -> [(t, Int, a)] -> FailAt a -> ParseResult a b
```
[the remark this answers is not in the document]
bpfoley: Let's go with c.

src/comp/PreStrings.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-missing-signatures #-}
```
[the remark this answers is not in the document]
bpfoley: Yup. This is one of those cases of files that are mostly just a list of convenience bindings of one/a few simple types. I'm not sure how much we benefit adding many many many signatures here. What do you think?

src/comp/TCMisc.hs
```diff
@@ -53,17 +53,26 @@ import IOUtil(progArgs)
...
 rtrace s x = if doRTrace then traces s x else x
 
-doTraceSatisfy= elem "-trace-tc-satisfy" progArgs
+doTraceSatisfy :: Bool
+doTraceSatisfy = elem "-trace-tc-satisfy" progArgs
+
+satTrace :: String -> a -> a
 satTrace s x = if doTraceSatisfy then traces s x else x
+
+satTraceM :: Monad f => String -> f ()
```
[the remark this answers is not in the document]
bpfoley: You're completely right. m is the convention, isn't it...

src/comp/TCheck.hs
```diff
@@ -1739,6 +1753,9 @@ tiVar _ _ _ _ = internalError "TCheck.tiVar"
 
 -- tiVar_WithImplRead sel = tiVar (ImplRead sel)
 
+noDesugar :: (PPrint a, PVPrint a, HasPosition a, Types b)
+          => Type -> TI (Type, [VPred], b)
+          -> a -> TI ([VPred], b)
```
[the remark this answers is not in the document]
bpfoley: Yeah, it doesn't quite fit, but a also works as CExpr.

src/comp/TCheck.hs
```diff
@@ -1769,6 +1786,8 @@ tryReadDesugar base_expr (ImplRead msel) as td orig_expr = do
          _ -> read_res
      _ -> def_res
 
+tiSub :: ImplReadTag -> [Assump] -> Type -> Position -> CExpr -> CExpr
```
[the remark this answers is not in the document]
bpfoley: Actually, it's not quite. You can't unify
```ImplReadTag -> [Assump] -> Type -> Position -> CExpr -> CExpr -> TI ([VPred], CExpr)```
to
```ImplReadTag -> [Assump] -> Type -> e -> TI ([VPred], e)```

File: src/comp/TCheck.hs [hunk 24]
```diff
@@ -2696,6 +2715,8 @@ type Impl   = (Id, ([CClause], [CQual]))
 --   clauses =   the clauses of the definition
 --   quals =     the implicit condition patterns
 --               (an empty list if the def is not for a method)
+tiImpl :: [Assump] -> Type -> (a, ([CClause], [CQual]))
+       -> TI ([VPred], ([CClause], [CQual]))
```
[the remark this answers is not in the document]
bpfoley: Actually not quite: (a, ([CClause], [CQual])) /= ([CClause], [CQual])

bpfoley:
Thanks for the painstaking review. The more specific types really help document the intent of a lot of the code.

src/comp/PreStrings.hs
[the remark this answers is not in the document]
quark17: I agree, turning off the warning seems like the right thing for this file.

Unless there was some benefit to writing each line like this:
`fsFoo :: FString = mkFString "foo"`

src/comp/TCheck.hs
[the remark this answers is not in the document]
quark17: Ah, you're right.  Ok, well, let's at least replace `a` with `Id`

src/comp/BinData.hs
```diff
@@ -190,14 +197,22 @@ data BinTable = BinTable { string_table   :: Table FString
...
+set_type_table :: BinTable -> Table Type -> BinTable
 set_type_table     bt t = bt { type_table = t }
--- set_loc_table      bt t = bt { loc_table = t }
+
+set_itype_table :: BinTable -> Table IType -> BinTable
 set_itype_table    bt t = bt { itype_table = t }
--- set_aexpr_table    bt t = bt { aexpr_table = t }
 
+unknownTable :: BinTable
 unknownTable = BinTable unknown unknown unknown unknown unknown -- unknown unknown
```
quark17: Can also delete the commented-out fields here, too

src/comp/CSyntax.hs
```diff
@@ -994,13 +998,18 @@ ppIfcPragma detail ps =
...
         <+> text "#-}"
 
+ppFDs :: PDetail -> CFunDeps -> Doc
 ppFDs d [] = empty
 ppFDs d fd = text " |" <+> sepList (map (ppFD d) fd) (t",")
+
+ppFD :: PDetail -> ([Id], [Id]) -> Doc
 ppFD d (as,rs) =
     sep (map (ppVarId d) as) <+> t "->" <+> sep (map (ppVarId d) rs)
+ppPreds :: PDetail -> [CPred] -> Doc -> Doc
```
quark17: Insert a line between these functions (particularly if you're going to add one between `ppFDs` and `ppFD`)

src/comp/CSyntax.hs
```diff
@@ -1010,12 +1019,14 @@ instance PPrint IdK where
     pPrint d p (IdKind i k) = pparen True $ pp d i <+> t "::" <+> pp d k
     pPrint d p (IdPKind i pk) = pparen True $ pp d i <+> t "::" <+> pp d pk
 
+pBlock :: p -> Int -> Bool -> [Doc] -> Doc
```
quark17: `p` should be PDetail

src/comp/CSyntax.hs
```diff
@@ -1164,6 +1178,7 @@ ppCase detail scrutinee arms =
                  ppQuals detail (cca_filters arm) <+> t "-> ",
                  nest 2 (pp detail (cca_consequent arm))]
 
+ppOp :: (PPrint a, PPrint b) => PDetail -> Int -> Id -> a -> b -> Doc
```
quark17: `a` and `b` should be `CExpr`

src/comp/CVPrint.hs
```diff
@@ -715,13 +747,18 @@ ppCase detail scrutinee arms =
               nest 2 (ppCaseBody detail (cca_consequent arm))] <> t";"
         ppPat pt = pp detail pt
 
+findPs :: CExpr -> [CExpr]
 findPs (CBinOp e1 i e2) | getBSVIdString i == "," = e1:(findPs e2)
 findPs e = [e]
 
+-- The 'tuple' here refers to the fact that CSyntax constructors take one
```
quark17: After I told you to add this comment, I think maybe it should be removed.  I do see that `ppTuple` is also used for printing actual tuples (in `ppOp`)

src/comp/GenFuncWrap.hs
```diff
@@ -286,15 +287,16 @@ funcDef errh symt i oqt@(CQType octxs ot) i_ n (CQType _ t) =
 -- ---------------
 
 -- make the foreign function declaration (with escaped id) to be wrapped.
---   n   = the number of arguments
+--   _   = the number of arguments
```
quark17: The comment on `funcDef_` doesn't mention the final argument, `args`.  It appears that when `args` was added, it made the `n` argument unnecessary.  I would suggest just removing the now unused first argument, removing the comment about it (rather than changing `n` to `_` in the comment), and added a new bullet to the comment for `args`: names of the arguments to the function.

src/comp/ISimplify.hs
```diff
@@ -69,10 +71,10 @@ getTuple (IAps (ICon iii (ICDef { iConDef = body })) ts []) =
         case iSimpAp False body ts [] of
         IAps (ICon _ (ICTuple { })) _ ms -> Just ms
         _ -> Nothing
-
```
quark17: If you're removing this blank, you might as well remove the blank between the signature and the first arm.

src/comp/ISyntax.hs
```diff
@@ -1221,6 +1221,7 @@ instance PPrint (IRules a) where
         foldr ($+$) (text "") (map (pPrint d 0) sps) $+$
         foldr ($+$) (text "") (map (pPrint d 0) rs)
 
+ppQuant :: PPrint a => String -> PDetail -> Int -> Id -> a -> IExpr b -> Doc
```
quark17: Change `a` to `IType`

src/comp/ITransform.hs
```diff
@@ -1319,18 +1324,22 @@ iLog2 i =
         else
              Nothing
 
+inc :: IExpr a1 -> IExpr a2
```
quark17: I don't think this is ever used to create a new type, so just replace `a1` and `a2` with the same `a`

src/comp/ParseOp.hs
```diff
@@ -324,13 +328,15 @@ doOp ft a op rs os es =
                             doOne ft rs os (e' : es')
             _ -> internalError ("ParseOp.doOp: Bad operator arity (2) for "++pfpString op)
 
+precOfA :: FixTable -> a -> Id -> (Int, Fixity)
```
quark17: Replace `a` with `Int`

src/comp/Synthesize.hs
```diff
@@ -636,7 +647,9 @@ type ShiftFun = Int -> [AExpr] -> [AExpr]
 
 permSL :: ShiftFun
 permSL  n xs = take (length xs) (replicate n aFalse ++ xs)
+permSRL :: Int -> [AExpr] -> [AExpr]
```
quark17: The type of these two functions is just `ShiftFun` (as used in the function above)

src/comp/TCheck.hs
```diff
@@ -55,16 +56,24 @@ import Util(traceM)
...
 doETrace = elem "-trace-type-expl" progArgs
+rtrace :: String -> a -> a
 rtrace s x = if doRTrace then traces s x else x
+etrace :: String -> a -> a
 etrace s x = if doETrace then traces s x else x
 
+doTraceSatisfy :: Bool
 doTraceSatisfy= elem "-trace-tc-satisfy" progArgs
 -- satTrace s x = if doTraceSatisfy then traces s x else x
+satTraceM :: Monad f => String -> f ()
```
quark17: It now occurs to me that this might be used only in one monad, so we could just put that (`TI`) instead of `f` (if you care)

src/comp/Verilog.hs
```diff
@@ -903,11 +908,13 @@ vIsValidIdent (c:cs) = (isLetterOrUnderscore c) &&
             (isAlpha x) || (isDigit x) || (x == '_') || (x == '$')
 
 -------
-
```
quark17: I wouldn't delete this blank line

src/comp/CVPrint.hs
bpfoley: :)

src/comp/ISyntax.hs
bpfoley: Can't do this. It's used both for IKind and IType in different places

src/comp/ASchedule.hs
```diff
@@ -3507,14 +3527,21 @@ verifyAssertionSr (_, rpragma) _ =
...
 
+errFireWhenEnabledBecauseBlocked :: (PVPrint a, PPrint a, PPrint b)
+                                 => (Id, RulePragma) -> [a] -> b
+                                 -> EMsg
 errFireWhenEnabledBecauseBlocked a rs sch =
     errFireWhenEnabledBecause a rs sch
         "because it is blocked by rule"
 
+errFireWhenEnabledBecause :: (PVPrint a, PPrint a, PPrint b)
+                          => (Id, RulePragma) -> [a] -> b -> String
```
quark17: Change `a` to `ARuleId` and `b` to `AScheduler`

src/comp/ASchedule.hs
```diff
@@ -3507,14 +3527,21 @@ verifyAssertionSr (_, rpragma) _ =
 
 -- --------------------
 
+assertErrorMessage :: (Id, RulePragma) -> Maybe Doc -> EMsg
 assertErrorMessage (r,a) mreason =
     (getIdPosition r,
      ERuleAssertion (pfpString r) (getRulePragmaName a) mreason)
 
+errFireWhenEnabledBecauseBlocked :: (PVPrint a, PPrint a, PPrint b)
```
quark17: Change `a` to `ARuleId` and `b` to `AScheduler`

src/comp/ASchedule.hs
```diff
@@ -3544,6 +3572,9 @@ errCannotScheduleFirst a before_sched before_exec =
                     ]
           in hdr $+$ (nest 2 (vsep lst))
 
+errCannotScheduleFirstViaMeth :: (PPrint a, PPrint b, PPrint c)
+                              => (Id, RulePragma) -> [(a, [Either b c])]
```
quark17: Change `a`, `b`, and `c` to `AMethodId`, `ARuleId`, and `AMethodId` (it's a filtered lookup result from `RuleBeforeMap`)

bpfoley: a is sometimes Id, sometimes AMethodId

quark17: `AMethodId` is an alias for `Id`.  And this function is only called in one place (and that one place is calling it on Ids that are methods, so I suggest using `AMethodId` as the name, for documenting.)  Further the `Id` in the type `(Id, RulePragma)` might be better expressed as `ARuleId`, which is also an alias for `Id` and that's the alias that's used in the one place that calls this function.

src/comp/ASchedule.hs
```diff
@@ -3507,14 +3527,19 @@ verifyAssertionSr (_, rpragma) _ =
 
 -- --------------------
 
+assertErrorMessage :: (Id, RulePragma) -> Maybe Doc -> EMsg
 assertErrorMessage (r,a) mreason =
     (getIdPosition r,
      ERuleAssertion (pfpString r) (getRulePragmaName a) mreason)
 
+errFireWhenEnabledBecauseBlocked :: (Id, RulePragma) -> [ARuleId] -> AScheduler
```
quark17: Use the alias `ARuleId` in place of `Id`, here and in the other functions that take `(Id, RulePragma)`

src/comp/ASchedule.hs
bpfoley: Sorry, I mistyped. I meant 'sometimes Id, sometimes MethodId'

If I change a to Id, I get the compile error
```ASchedule.hs:3505:62: error:
    • Couldn't match type ‘MethodId’ with ‘Id’
      Expected type: [(Id, [Either ARuleId AMethodId])]
        Actual type: [(MethodId, [Either ARuleId AMethodId])]
    • In the second argument of ‘errCannotScheduleFirstViaMeth’, namely
        ‘befores’
      In the expression: errCannotScheduleFirstViaMeth a befores
      In the expression: [errCannotScheduleFirstViaMeth a befores]```

If I change it to MethodId I get the error
```ASchedule.hs:3580:27: error:
    • Couldn't match type ‘MethodId’ with ‘Id’
      Expected type: (Id, RulePragma)
        Actual type: (MethodId, RulePragma)
    • In the first argument of ‘assertErrorMessage’, namely ‘a’
      In the expression: assertErrorMessage a (Just reason)
      In the expression:
        let
          hdr
            = text
                ("because it uses methods which require rules to execute before it:")
          lst = map describe befores
          reason = hdr $+$ (nest 2 (vsep lst))
        in assertErrorMessage a (Just reason)
     |
3580 |     in assertErrorMessage a (Just reason)```

bpfoley: Nevermind. Realized my mistake eventually :)

quark17: I was mistaken, too.  But it looks like we've figured it out.  Thanks!

PR #78, opened 2020-02-28
System tcl

bpfoley:


src/tcltk/HaskellTcl/Makefile.bluespec
```diff
@@ -1,64 +1,28 @@
...
-ITCLVER ?= 3.4
-ITCLDIR ?= ../itcl$(ITCLVER)
-
-ITKVER ?= 3.4
-ITKDIR ?= ../itk$(ITCLVER)
-
-
-INCPATH = $(TCLDIR)generic/
-CCFLAGS += -I$(INCPATH)
+INCPATH = /usr/include/tcl
```
quark17: I don't think this path is portable?

bpfoley: Yeah, I wasn't entirely happy hardcoding a path like this, but I think it's probably OK across Linuxes.

Unfortunately I don't have any other environment set up to test in yet, but if it comes to it we can probably figure out how to get at the system's tclConfig.sh to get include paths etc.

PR #80, opened 2020-02-28
Add the last set of top-level sigs, and enable GHC warnings

bpfoley:
While here, turn on GHC tab warnings, and remove a duplicate 'head -1'
in the Makefile.

### DIFF
--- a/src/comp/BackendNamingConventions.hs
+++ b/src/comp/BackendNamingConventions.hs
@@ -37,7 +37,7 @@ import Error
 import PPrint
 import Flags(Flags, removeReg, removeCross)
 
-import Position(getPosition, noPosition)
+import Position(Position, getPosition, noPosition)
 
 import ASyntax
 import VModInfo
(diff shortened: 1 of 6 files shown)


src/comp/BackendNamingConventions.hs
```diff
@@ -116,12 +122,13 @@ isCRegInst avi = (isCRegN avi) || (isCRegUN avi) || (isCRegA avi)
 -- ---------------
 -- Names of CReg methods
 -- (inlining happens before methods are renamed to ports)
-
```
quark17: Don't delete this blank line; the comment is for the larger group of functions not just these first three

src/comp/BackendNamingConventions.hs
```diff
@@ -132,25 +139,29 @@ cregWriteArgId i n = mkMethId i (cregWriteId n) Nothing (MethodArg 1)
 
 -- ---------------
 -- Names of ports and parameters on primtive CReg
-
```
quark17: I would also not delete this blank

src/comp/Makefile
```diff
@@ -84,7 +84,7 @@ WISHFLAGS += $(EXTRAWISHLIBS)
 GHC ?= ghc
 GHCJOBS ?= 1
 
-GHCVERSION=$(shell $(GHC) --version | head -1 | egrep -o "[0-9]+\.[0-9]+\.[0-9]" | head -1)
+GHCVERSION=$(shell $(GHC) --version | head -1 | egrep -o "[0-9]+\.[0-9]+\.[0-9]" )
```
quark17: I don't know enough about this command (and the range of possible GHC version strings) to know if this is safe to do?

bpfoley: It doesn't really matter: `ghc --version | head -1` is only ever going to print one line, and the `egrep -o` prints out a subset of that line, so the final `| head -1` is definitely redundant.

PR #88, opened 2020-02-29
Various bug fixes and minor changes

thoughtpolice:
These changes start getting things in order for some upcoming improvements to the build system (and other assorted work). Notably it includes cleanups for the `src` directory to make it easier to navigate (some more of these will come). But also importantly two bugfixes are included: one for installing utilities like `bsv2bsc`, and one for `libstp_stub`.

The cleanup of hctl using `CApiFFI` means that we now at least need GHC 7.10 (~2015), which I think is probably just fine. For example, Ubuntu 16.04 still works, and will actually be EOL'd within ~1 year, so I think that's a reasonable minimum bound for support. It also refactors the code to be just one C file for starting the RTS and a finalizer callback, since that can't be worked around with *any* language extension, just to keep things simple.


.github/workflows/build.yml
```diff
@@ -1,14 +1,23 @@
...
 jobs:
+  build-check:
+    name: "Check: code cleanliness"
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check tabs and whitespace
+        shell: bash
+        run: ".github/workflows/check_whitespace.sh"
+
```
bpfoley: Good idea! Do the whitespace check before getting any other deps/installing other packages, and only on one OS variant...

thoughtpolice:
Thinking more about it, I think requiring a newer Haskell compiler is probably pretty reasonable, actually, even if the platform itself is older (e.g. Ubuntu 16.04, RHEL 7, etc.) Herbert's PPA for Ubuntu and Jens package for CentOS etc are both very reliable and very well supported ways of using a newer GHC on older machines and are in use by many people -- I used to help maintain GHC itself, and recommended both of these solutions back then, as well! It's not at all strange to rely on them. So we could just say GHC 8.6.5 (or newer) is the supported compiler. All of the user-linkable products of the build system are written in C/C++, so it's not like there's anything holding us back there...

Speaking from experience, I think it's really a matter of how you expect users to get their hands on the compiler and play with it. What problem do you want to address by supporting the "sanctioned" RHEL version of GHC, 7.6.3? GHC 7.10.3 is already 5 years old. Supporting that has a lot of different requirements than just *supporting* RHEL users in practice. Likewise, if a user wants to just compile Bluespec themselves, then can pretty easily add Herbert/Jens packages to their system, install the requisite Haskell compiler, and be on their way. I think this is a reasonable request (things are otherwise pretty simple!)

My point here is that I suspect the *intended* problem you want to head off here is, "what if a RHEL user wants Bluespec without 3rd party things"? That's not outrageous, because lots of FPGA tools support RHEL. So here's the solution: just provide a package for Bluespec! After all, if we provide a package, that's the most 1st-party thing you can ask for! Most people in fact do not want to compile Bluespec, I suspect, they want to use it. We should just provide `.rpm` files they can install. You're trying to solve the problem package maintainers solve, i.e. someone who wants to package Bluespec for their distro, but I assure you -- like me, they'll come complaining when they need some strange Haskell compiler to be supported. Arch Linux maintainers will be fine figuring things out I'm sure. Likewise, when you just tell users "Install this sanctioned repository" because they actually want to compile it themselves, they'll do it 9/10 times without issue.

So I think if you want RHEL users to be comfortable, the better goal here is to actually have the CI system produce artifacts that users can use, rather than fixate on a particular version range of the Haskell compiler. In fact if we produce binary packages, we should mandate on exactly 1 compiler version for all of those results, if possible. (This will also segue into having artifacts for releases, but that's another thing.) In practice you only *actually* need to (and should) support ~3ish GHC versions, because that's reasonably all you're going to get support for, realistically.

I can definitely help fix up the README to add notes on installing the proper versions of the compiler tools from Jens/Herbert's PPA.

.github/workflows/build.yml
thoughtpolice: Yes, I think we should add most new checks this way. We could also add `clang-format` (for bluesim), `weeder` for the compiler, etc, all as independent checks.

quark17:
@thoughtpolice : That makes sense, I agree.  In fact, a primary goal of this open source release was to also have prebuilt releases, recognizing that most users will just want to download it and not compile it themselves.  I'd lost track of that and it's good to be reminded.

thoughtpolice:
@quark17 Good to know! My inevitable goal is in fact to segue from this work into fixing up the build system, to fixing up the release process, to providing packages... I have plenty of cleanups that I think will help smooth things out a lot, and this is just the start.

quark17:
I added the GHC 7.10.1 requirement to the README, just briefly, but @thoughtpolice if you want to submit more significant improvements to the README, please do!

PR #96, opened 2020-03-02
Use slices and eta reduction to simplify lambdas

bpfoley:
(As recommended by hlint)


bpfoley:
Personally, I'm used to this style, but if you feel that any of these make things unnecessarily confusing, please push back.

PR #98, opened 2020-03-02
Complete macOS build

ncihnegn:
Add macOS build action as well.
The environment set-up is in _install_dependencies_macos.sh_.
It also include a workaround for _bluetcl_ exit crash.

### DIFF
--- a/.github/workflows/install_dependencies_macos.sh
+++ b/.github/workflows/install_dependencies_macos.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+brew install cabal-install fontconfig gperf icarus-verilog pkg-config && \
+    cabal update && \
+    cabal v1-install old-time regex-compat split syb
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -168,7 +168,6 @@ GHCCOMPILEFLAGS = -o $@ -c $<
 # On Mac OS, we'll need to update the dylib location,
 # so linking needs to pad the field so that there's room to overwrite
 ifeq ($(OSTYPE), Darwin)
-#GHC += -optc -Wno-deprecated-declarations
 GHC += -optl -headerpad_max_install_names
 endif
 
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -168,7 +168,7 @@ GHCCOMPILEFLAGS = -o $@ -c $<
 # On Mac OS, we'll need to update the dylib location,
 # so linking needs to pad the field so that there's room to overwrite
 ifeq ($(OSTYPE), Darwin)
-GHC += -optc -Wno-deprecated-declarations
+#GHC += -optc -Wno-deprecated-declarations
 GHC += -optl -headerpad_max_install_names
 endif
 
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -55,7 +55,10 @@ TCL_HS    = ../vendor/htcl
 TCL_ARGS  = -L$(TCL_HS) $(shell pkg-config --silence-errors --cflags-only-I tcl tk || echo -I/usr/include/tcl)
 TCL_LIBS  = -ltcl$(TCL_VER) -ltclstub$(TCL_VER) \
             -ltk$(TCL_VER) -ltkstub$(TCL_VER) \
-            -litcl$(ITCL_VER) -litclstub$(ITCL_VER) -lhtcl
+            -litcl$(ITCL_VER) -lhtcl
+ifeq ($(OSTYPE), Darwin)
+TCL_LIBS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitcl$(ITCL_VER).dylib))"
+endif
 
 # STP
 STP_HS      = ../vendor/stp/include_hs
@@ -76,8 +79,12 @@ BSCBUILDLIBS = \
 
 EXTRAWISHLIBS = $(shell pkg-config --libs fontconfig xft)
 
-WISHFLAGS = -litkstub$(ITCL_VER) -litk$(ITCL_VER) $(shell pkg-config --libs x11)
+WISHFLAGS = -litk$(ITCL_VER) $(shell pkg-config --libs x11)
 WISHFLAGS += $(EXTRAWISHLIBS)
+ifeq ($(OSTYPE), Darwin)
+WISHFLAGS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitk$(ITCL_VER).dylib))"
+WISHFLAGS += -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Tk.framework/Headers/
+endif
 
 # -----
 # GHC
@@ -161,6 +168,7 @@ GHCCOMPILEFLAGS = -o $@ -c $<
 # On Mac OS, we'll need to update the dylib location,
 # so linking needs to pad the field so that there's room to overwrite
 ifeq ($(OSTYPE), Darwin)
+GHC += -optc -Wno-deprecated-declarations
 GHC += -optl -headerpad_max_install_names
 endif
 
--- a/src/vendor/htcl/haskell.c
+++ b/src/vendor/htcl/haskell.c
@@ -26,6 +26,7 @@ void
 htcl_finalizeTclObj(Tcl_Obj* o)
 {
 #if defined(__APPLE__) && defined(__MACH__)
+/* Workaround for https://sourceforge.net/p/tcl/bugs/4043/ */
   if (Tcl_IsShared(o) == 1)
 #endif
   Tcl_DecrRefCount(o);
--- a/src/vendor/htcl/haskell.c
+++ b/src/vendor/htcl/haskell.c
@@ -25,5 +25,8 @@ htcl_initHaskellRTS(int *argc, char **argv[])
 void
 htcl_finalizeTclObj(Tcl_Obj* o)
 {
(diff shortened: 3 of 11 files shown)


bpfoley:
Thanks for the legwork on this. I'm replicating what you've done on my local MacBook, and I'll try to review in detail shortly.

.github/workflows/install_dependencies_macos.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+brew cask install xquartz
```
bpfoley: I'm a little surprised we need xquartz. The native /usr/bin/wish has a Cocoa UI and doesn't use X at all.

.github/workflows/install_dependencies_macos.sh
```diff
@@ -0,0 +1,7 @@
...
+brew install cabal-install ghc && cabal v1-install old-time regex-compat split syb
```
bpfoley: To be pedantic you could drop the `ghc` here. It's a dependency of `cabal-install`.

ncihnegn: Agree.

.github/workflows/install_dependencies_macos.sh
ncihnegn: It's a different issue to configure for Cocoa.
Currently the Makefile looks for X11 and will link X11 if found.
If not found it will only issue warnings.
So XQuartz is actually optional.

src/comp/Makefile
```diff
@@ -76,7 +76,7 @@ BSCBUILDLIBS = \
 
 EXTRAWISHLIBS = $(shell pkg-config --libs fontconfig xft)
 
-WISHFLAGS = -litkstub$(ITCL_VER) -litk$(ITCL_VER) $(shell pkg-config --libs x11)
+WISHFLAGS = -litk$(ITCL_VER) $(shell pkg-config --libs x11)
```
bpfoley: This errorred for me saying it couldn't find X11/xlib.h. I needed to add --cflags to the pkg-config invocation.

src/comp/Makefile
```diff
@@ -212,6 +212,7 @@ GHCCOMPILEFLAGS = -o $@ -c $<
 # On Mac OS, we'll need to update the dylib location,
 # so linking needs to pad the field so that there's room to overwrite
 ifeq ($(OSTYPE), Darwin)
+GHC += -optc -Wno-deprecated-declarations
```
bpfoley: I didn't seem to need this for my build

src/VPI/Makefile
```diff
@@ -33,6 +33,7 @@ COPT ?= -O3
 CDEBUG ?=
 CFLAGS += $(PROF) $(COPT) $(CDEBUG)
 
+OSTYPE = $(shell $(TOP)/platform.sh ostype)
```
bpfoley: I didn't seem to need this for my build either. It seemed to work fine with Apple's clang-1100.0.33.16 too.

.github/workflows/install_dependencies_macos.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+ln -sf /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Tk.framework/Headers/X11 /usr/local/include
```
bpfoley: I think we can replace this with the `(pkg-config --cflag x11)` in src/comp/Makefile.

.github/workflows/install_dependencies_macos.sh
```diff
@@ -0,0 +1,10 @@
...
+ln -sf /System/Library/Tcl/8.5/itcl3.4/libitcl3.4.dylib /usr/local/lib
+ln -sf /System/Library/Tcl/8.5/itk3.4/libitk3.4.dylib /usr/local/lib
```
bpfoley: I understand this is a hack, but it's one we'll definitely have to fix properly, because it means that bluesim and bluewish will try to dynamically load these libraries from /usr/local/lib, which we  won't want to depend on if we try to make a binary redistributable.

.github/workflows/install_dependencies_macos.sh
```diff
@@ -0,0 +1,10 @@
...
+
+brew cask install xquartz && \
+    brew install cabal-install coreutils fontconfig gperf icarus-verilog pkg-config && \
```
bpfoley: I'd prefer not to drag in the dependency of coreutils if we don't have to. Instead we can replace the single invocation of realpath in src/comp/wrapper.sh with

```function normpath() {
  # Remove all /./ sequences.
  local path=${1//\/.\//\/}

  # Remove dir/.. sequences.
  while [[ $path =~ ([^/][^/]*/\.\./) ]]; do
    path=${path/${BASH_REMATCH[0]}/}
  done
  echo $path
}

ABSPATH=$(normpath -e "$PWD/$0")
```

Sure, it uses bash-isms, but this wrapper already uses bash.

src/VPI/Makefile
ncihnegn: Yes, ``OSTYPE``` is defined in _platform.mk_.

src/comp/Makefile
ncihnegn: Perhaps it is because I am using clang-1100.0.33.17.

.github/workflows/install_dependencies_macos.sh
ncihnegn: I took a different approach.

src/vendor/htcl/haskell.c
```diff
@@ -25,5 +25,8 @@ htcl_initHaskellRTS(int *argc, char **argv[])
 void
 htcl_finalizeTclObj(Tcl_Obj* o)
 {
+#if defined(__APPLE__) && defined(__MACH__)
+  if (Tcl_IsShared(o) == 1)
```
ncihnegn: I am not sure about this. But at least it works.

bpfoley: Good enough for now! Could you add a comment to this something like
```
/* Workaround for https://sourceforge.net/p/tcl/bugs/4043/ */
```

bpfoley:
Looks good enough to get useful work done with for now. We can polish it further later. Julie, what do you think? Shall we get this puppy merged?

src/comp/Makefile
ncihnegn: Somehow I don't need it anymore.

quark17:
There's a conflict in the Makefile, now that the tcl stubs are not explicitly linked.  Would you mind resolving that?

ncihnegn:
> There's a conflict in the Makefile, now that the tcl stubs are not explicitly linked. Would you mind resolving that?
If you want to merge, I can remove them.
But I don't think removing tcl/tk stubs is portable. On archlinux the build will fail.
You may want to revisit #102.

bpfoley:
> > There's a conflict in the Makefile, now that the tcl stubs are not explicitly linked. Would you mind resolving that?
> 
> If you want to merge, I can remove them.
> But I don't think removing tcl/tk stubs is portable. On archlinux the build will fail.
> You may want to revisit #102.

Rats. I'll take a look. You don't happen to know if there are any custom GitHub Action runners for distros/OSes other than Ubuntu/macOS, do you?

ncihnegn:
> > > There's a conflict in the Makefile, now that the tcl stubs are not explicitly linked. Would you mind resolving that?
> > 
> > 
> > If you want to merge, I can remove them.
> > But I don't think removing tcl/tk stubs is portable. On archlinux the build will fail.
> > You may want to revisit #102.
> 
> Rats. I'll take a look. You don't happen to know if there are any custom GitHub Action runners for distros/OSes other than Ubuntu/macOS, do you?

Sorry, I need to take it back. I just test it on archlinux and it works fine. My bad.

jameyhicks:
You can use openSUSE Build Service to build packages for many distros, including archlinux. I imagine you could trigger an OBS build from a GitHub Action runner.

jameyhicks:
Nice work on this pull request!

quark17:
Why does the CI build for macOS take 53 minutes?

bpfoley:
At a guess, homebrew is building something from source that hasn't yet got a bottle (binary package) for macOS 10.15.

bpfoley:
> At a guess, homebrew is building something from source that hasn't yet got a bottle (binary package) for macOS 10.15.

Hmm. Nope. I'm wrong. Installing the dependencies only takes a couple of minutes. I'll eyeball a build to see where the time is being spent.

ncihnegn:
Probably the assigned Mac machine was simply too slow. Sometimes it finished much faster, like https://github.com/B-Lang-org/bsc/runs/481637505?check_suite_focus=true. Maybe that was a fast Mac.

mtikekar:
To build binaries that also work on older MacOS versions, use `MACOSX_DEPLOYMENT_TARGET=10.13 make -j2`. I used the `actions/upload-artifact@v1` action to save the inst folder.

BracketMaster:
So how do you build on MacOS? when running make, I currently get most way through the build,
and then GHC gets stuck on:

``ghc: on the commandline: missing argument for flag: -L``

quark17:
I assume that the makefile definitions at the top of `src/comp/Makefile` aren't checking the results, and one of them is set to an empty string.  If so, it'd be nice if the Makefile detected that error early and gave a better error about the problem, rather than it being detected by GHC in this way.

To answer your question, though: Have a look at [.github/workflows/](https://github.com/B-Lang-org/bsc/blob/master/.github/workflows/). The `build.yml` file has an entry for macOS that shows how it's run.  The `Install dependencies` step calls the file `install_dependencies_macos.sh` which has the commands for setting up the environment.  And the `Build` step has the `make` command line to run.

PR #100, opened 2020-03-03
Splitfmt

bpfoley:


src/comp/IInline.hs
```diff
@@ -257,37 +250,28 @@ promoteSome :: IExpr a -> (IExpr a, Bool)
 promoteSome e |  t /= itFmt = (e, False)
               where t = iGetType e
 
-promoteSome (IAps ci@(ICon _ (ICPrim _ PrimIf)) ti [cond, (IAps cc@(ICon _ (ICPrim _ PrimFmtConcat)) tc [e00, e01]),
-                                                            (IAps    (ICon _ (ICPrim _ PrimFmtConcat)) _  [e10, e11])])
+promoteSome (IAps ci@(ICon _ (ICPrim _ PrimIf)) ti [cond, (IAps cc@(ICon _ (ICPrim _ PrimFmtConcat)) tc [e00, e01]), (IAps    (ICon _ (ICPrim _ PrimFmtConcat)) _  [e10, e11])])
```
quark17: Was there a reason that you decided to make this one long line?  GitHub's web page can at least show me the whole of the original lines, but I have to scroll sideways to see the new one-line version.  I'd prefer if people didn't have to side-scroll to see code; we can be longer than 80 columns, since people can generally widen their editor or terminal beyond that, but this seems too wide.  (Also, maybe it's me, but I find lists of complex elements to be easier to read when they are split into lines.)

src/comp/ISplitFmt.hs
```diff
@@ -0,0 +1,594 @@
...
+           in (IAps c ts es')
+        combineFmts e = e
+
+        -- next move primFmtConcats up so that after this step, all the primFmtConcats in any
+        -- expression come first
+        -- bottom up processing
+        promoteConcat r (IAps ci@(ICon _ (ICPrim _ PrimIf)) ti es@[cond, (IAps cc@(ICon _ (ICPrim _ PrimFmtConcat)) tc [e0, e1]), e2]) =
+          promoteConcat False (IAps cc tc [(IAps ci ti [cond, e0, e2]), (IAps ci ti [cond, e1, emptyFmt])])
+        promoteConcat r (IAps ci@(ICon _ (ICPrim _ PrimIf)) ti [cond, e2, (IAps cc@(ICon _ (ICPrim _ PrimFmtConcat)) tc [e0, e1])]) =
+          promoteConcat False (IAps cc tc [(IAps ci ti [cond, e2, e0]), (IAps ci ti [cond, emptyFmt, e1])])
```
quark17: Good catch on noticing that `emptFmtM` wasn't doing anything in the monad and eliminating the do-block here.  Does anything in the rest of this function use the monad, though? Can we eliminate the do-blocks in the other arms, too?

src/comp/IInline.hs
bpfoley: No good reason. I moved them all onto one line so I could understand the differences between the different patterns, and forgot to undo it. I agree they're too long now. I'll undo that.

bpfoley: By the way, I don't know how you feel about -XPatternSynonyms, but since we're going to support only GHC 7.8 and later, we could selectively introduce some pattern synonyms to avoid the more repetitive patterns.

src/comp/ISplitFmt.hs
bpfoley: > Moving `splitFmts` into its own package is a good idea.
> 
> I notice in `bsc.hs` that the `inlineFmts` stage calls both `splitFmts` and (first) `iInlineFmts`. Does `iInlineFmts` look like it can be pulled out, too? perhaps putting both functions into the same file, and making one entry point from `bsc.hs` that does them together? Since the stage is called `inlineFmts` perhaps name the file based on that instead of `ISplitFmt`? Perhaps `IInlineFmt` and `iInlineFmt`, unless that creates too much confusion with the existing name `iInlineFmts`?

I'll try doing that in a separate PR. It's a little error prone trying to pull all the functions and their transitive deps into another file, and I'd prefer not to break anything I've done so far.

bpfoley: > Good catch on noticing that `emptFmtM` wasn't doing anything in the monad and eliminating the do-block here. Does anything in the rest of this function use the monad, though? Can we eliminate the do-blocks in the other arms, too?

Even better catch! Nice! Done.

quark17:
Can you resolve the latest conflict?  It looks like `addFileArg` in `IInline.hs` had a small cleanup, and that code is now in `ISplitFmt.hs` which will need the cleanup applied.

src/comp/ISplitFmt.hs
```diff
@@ -0,0 +1,591 @@
...
+fsplitFmt (IAps x ts es) =
+    do  es' <- mapM fsplitFmt es
+        return (IAps x ts es')
+fsplitFmt x = return x
+
+addFileArg :: IExpr a -> IExpr a -> IExpr a
+addFileArg e (IAps (ICon fid f@(ICForeign {iConType = t})) [] es)
+             | isFileId fid = (IAps (ICon fid f {iConType = t'}) [] es')
+    where (_ , rt) = itGetArrows (getInnerType t)
+          es'      = [e] ++ es
```
quark17: Part of resolving the conflict should have been to change this expression to `e : es`

bpfoley: Argh, my bad! Thanks.

PR #104, opened 2020-03-05
hlint cleanups, part 2

bpfoley:
The first few patches may help improve performance. The later ones are just simplifications/readability improvements.

### DIFF
--- a/src/comp/ADumpSchedule.hs
+++ b/src/comp/ADumpSchedule.hs
@@ -265,8 +265,7 @@ doShowSchedule errh flags prefix pkg
     schedFileH <- openFileCatch errh emptyContext schedName WriteMode
     unless (quiet flags) $ putStrLn $ "Schedule dump file created: " ++
                                       (getRelativeFilePath schedName )
-    let   outputHandles = [schedFileH] ++
-                          if (hasDump flags DFschedule) then [stdout] else []
+    let   outputHandles = schedFileH : [stdout | hasDump flags DFschedule]
     --
     --
     let schedPutStr :: String -> IO ()
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -2888,13 +2888,11 @@ evalAp' f@(ICon i (ICDef t e)) as = do
         step i
         e' <- cacheDef i t e
         when doFunExpand $ do
-            u <- return (mkAp f as) --unheapAll (mkAp f as)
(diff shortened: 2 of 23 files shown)


src/comp/ADumpSchedule.hs
```diff
@@ -265,8 +265,7 @@ doShowSchedule errh flags prefix pkg
     schedFileH <- openFileCatch errh emptyContext schedName WriteMode
     unless (quiet flags) $ putStrLn $ "Schedule dump file created: " ++
                                       (getRelativeFilePath schedName )
-    let   outputHandles = [schedFileH] ++
-                          if (hasDump flags DFschedule) then [stdout] else []
+    let   outputHandles = schedFileH : [stdout | hasDump flags DFschedule]
```
quark17: Huh, does that list comprehension work?

src/comp/IExpand.hs
```diff
@@ -2891,12 +2891,12 @@ evalAp' f@(ICon i (ICDef t e)) as = do
...
         e' <- cacheDef i t e
         when doFunExpand $ do
-            u <- return (mkAp f as) --unheapAll (mkAp f as)
+            let u = mkAp f as --unheapAll (mkAp f as)
             traceM ("expand " ++ ppReadable u)
         r <- evalAp "ICDef" e' as
         when doFunExpand2 $ do
             let P _ re = r
-            u <- return re --unheapAll re
+                u = re --unheapAll re
```
quark17: Here and above: the code that's commented out (`unheapAll`) is monadic, which is why it's been replaced with a monadic no-op.  This code has been commented out since 2002, so I assume that we can remove it -- and probably remove the declaration of `u` and just inline it.

src/comp/ADumpSchedule.hs
bpfoley: Yeah. It's one of these surprising Haskell list comprehension tricks: if you don't provide a generator, you get a single item back, and if you filter on that, you can get 1 or 0 items.

```
λ> [1|True]
[1]
λ> [1|False]
[]
```

bpfoley: ...and mathematically, this kinda makes sense, because 1 is the identity operator for multiplication, and the size of a list comprehension is the product of the size of each of the generators uses. So
```
length [x|i<-[1..10], j<-[1..5],True] == 50
length [x|i<-[1..10],True] == 10
length [x|True] == 1
```

PR #105, opened 2020-03-05
Remove non-ASCII characters from source (it confuses hlint)

bpfoley:


src/comp/Prim.hs
```diff
@@ -76,7 +76,7 @@ data PrimOp =
         | PrimCase
 
         -- Only used in intermediate code
-        -- primSelect �k �m �n e  selects k bits at position m from n bits
+        -- primSelect k m n e  selects k bits at position m from n bits
```
quark17: The dot here does indicate type parameters instead of value parameters, so I'm not sure that just removing it is OK.  Maybe put `#` in place?  This goes back to when BSC has a user-readable `.bi` file alongside the `.bo` file, and that `.bi` file was in ASCII and this non-typical character was used to avoid namespace issues, or something.  Now that the `.bi` is gone (and all the info is in `.bo`), I don't know how necessary any of this (if any of it is still around -- but I think it still does if you pretty-print internal syntax in debug messages etc).  Anyway, since this is a comment it doesn't need to be parsable so `#` should be OK.

bpfoley: Ah, that's what it was for. How about I use @ like Haskell does for type application?

quark17: That works

PR #107, opened 2020-03-09
Move iInlineFmts out of IInline and merge with iSplitFmts

bpfoley:


src/comp/ISyntaxUtil.hs
```diff
@@ -1116,3 +1116,23 @@ emptyFmt = (IAps (ICon idFormat (ICForeign {fName    = getIdString(unQualId(idFo
...
          t = iGetType e
          tt = (t `itFun` itFmt)
+
+iSubst :: M.Map Id (IExpr a) -> M.Map Id (IExpr a) -> IExpr a -> IExpr a
+iSubst subMap defMap e = sub e
+  where sub (IAps f ts es) = IAps (sub f) ts (map sub es)
+        sub d@(ICon i val@(ICValue {})) =
+            case M.lookup i subMap of
+            Nothing ->
+              let ev = fromJustOrErr ("IInline.iSubst ICValue def not found: " ++ ppReadable i)
```
quark17: Change the qualifier in this error message from `IInline` to `ISyntaxUtil` (or remove it?)

src/comp/ISyntaxUtil.hs
```diff
@@ -1116,3 +1116,23 @@ emptyFmt = (IAps (ICon idFormat (ICForeign {fName    = getIdString(unQualId(idFo
...
+                                     (M.lookup i defMap)
+              in ICon i (val { iValDef = ev })
+            Just e -> e
+        sub u@(ICon i (ICUndet t k (Just v))) = ICon i (ICUndet t k (Just (sub v)))
+        sub c@(ICon {}) = c
+        sub ee = internalError ("iSubst: " ++ ppReadable ee)
+
+
+inlineIfc :: M.Map Id (IExpr a) -> M.Map Id (IExpr a) -> IEFace a -> IEFace a
```
quark17: Consider a follow-on cleanup to improve the names of `inlineIfc` and `iSubst`, to indicate that they're related.  I think that `inlineIfc` is just a version of `iSubst` on the type `IEFace` instead of `IExpr`.  I could imagine one of the following choices: (1) leave `iSubst` as is and rename `inlineIfc` as `iSubst_IEFace` (or `iSubstIEFace`, or even `iSubstIfc`); (2) also rename `iSubst` to be `iSubst_IExpr` (or `iSubstIExpr`); or (3) declare a typeclass `ISubst t` with the single method `iSubst` on type `t` and define instances for `IExpr` and `IEFace`.  I feel like all three forms are used at various places in the BSC source.

src/comp/ISyntaxUtil.hs
bpfoley: The problem is iSubst is used by inlineIfc which is used by both IInline and IInlineFmt.

src/comp/IInlineFmt.hs
```diff
@@ -589,3 +592,137 @@ ssplitFmt_ifc ifc_list
...
+    in if (change) then iInlineFmtsPhase1 imod'' else imod''
+
+iInlineFmtsT :: ((IExpr a) -> Bool) -> IModule a -> IModule a
+iInlineFmtsT tst imod@(IModule { imod_local_defs = ds,
+                                 imod_rules      = rs,
+                                 imod_interface  = ifc}) =
+    let smap = M.fromList [ (i, iSubst' tst smap dmap e) | IDef i t e _ <- ds, (t == itFmt) ] -- inline any def of type Fmt
+        ds' = iDefsMap (iSubst' tst smap dmap) ds
+        dmap = M.fromList [ (i, e) | IDef i t e _ <- ds' ]
+        ifc' = map (inlineIfc smap dmap) ifc
```
quark17: Ah, I see why you had to move `iSubst` and `inlineIfc` to `ISyntaxUtil`.  Because here we use the ordinary `inlineIfc`, and not a special version with a `tst` argument that calls `iSubst'`.

Perhaps create a new `IInlineUtil.hs` and put `iSubst` and `iSubst'` (or `iSubstWhen`) in that file, to be imported only by these two packages?

I do also wonder if the call to `iSubst'` could be eliminated by calling `iSubst'` with a filtered `smap`: For `iSubst`, when an ID is encountered that is not in the map, the expression is left alone but the `iValDef` field is updated; for `iSubst'`, the same is true, but there's a third case of when the ID is found in the `smap` but fails the `tst`, and in that case it leaves the expression entirely alone.  I suspect that leaving it alone and updating the `iValDef` are equivalent, so the need for `iSubst'` could be removed by simply pruning those IDs from the map.  (The fact that it's updating a field of all other defs as it goes seems bad, unless it really is benign; I'd want to explore that further.)

It's starting to seem like this cleanup might be getting more involved and require access to the testsuite, to make sure it's not breaking anything.

src/comp/ISyntaxUtil.hs
bpfoley: And excellent point, iSubst' is just iSubst with a test. That lets us rewrite iSubst as a one-liner...

src/comp/IInlineFmt.hs
quark17: Maybe start by just putting both `iSubst` and `iSubst'` into a new `IInlineUtil.hs` file, and leave a comment suggesting that `iSubst'` could possible be eliminated, as a future cleanup?

bpfoley: Sounds like something to follow up on later. Will the one line implementation of iSubst do for now?

src/comp/ISyntaxUtil.hs
```diff
@@ -4,7 +4,7 @@ import System.IO(Handle, BufferMode(..))
 import qualified Data.Map as M
 import qualified Data.Set as S
 import Data.Maybe(fromMaybe)
-import Util(flattenPairs)
+import Util(flattenPairs, fromJustOrErr)
```
quark17: This change should be removed

bpfoley: Woops. Thanks. Done.

PR #110, opened 2020-03-11
SEMonad: Avoid the use of refutable patterns (and so the need for fail)

bpfoley:


bpfoley:
I ran across this while refactoring SEMonad to use monad transformers. I was wondering what your thoughts about this are.

PR #113, opened 2020-03-11
Make the CI macOS GHC build parallel too.

bpfoley:


quark17:
I see you recently force-pushed a new version.  Is this version ready to be merged, or should I wait for your signal?  The macOS check is down to 40m, which is better.

bpfoley:
> I see you recently force-pushed a new version. Is this version ready to be merged, or should I wait for your signal? The macOS check is down to 40m, which is better.

It's ready.

I've found what the build slowness problem is BTW: the PREBUILDCOMMAND and BUILDCOMMAND for bluetcl & bluewish are using different flags on macOS, so everything ends up being compiled from scratch multiple times because GHC detects the flag changes. I'm working on a fix for that.

quark17:
> the PREBUILDCOMMAND and BUILDCOMMAND for bluetcl & bluewish are using different flags

Ah!

Note that a result of having the same flags is that bluetcl and bluewish have the same linking requirements; people have rightfully complained that bluetcl should not require `libX11` etc to run.  So if we could find a way to link them differently, but not recompile the individual modules, that'd be wonderful.  (And similarly, if bluetcl/bluewish didn't need to recompile modules that bsc already compiled, that'd be great.  I don't know if there's a run-time efficiency benefit to compiling them differently, though?)

bpfoley:
> > the PREBUILDCOMMAND and BUILDCOMMAND for bluetcl & bluewish are using different flags
> 
> Ah!
> 
> Note that a result of having the same flags is that bluetcl and bluewish have the same linking requirements; people have rightfully complained that bluetcl should not require `libX11` etc to run. So if we could find a way to link them differently, but not recompile the individual modules, that'd be wonderful. (And similarly, if bluetcl/bluewish didn't need to recompile modules that bsc already compiled, that'd be great. I don't know if there's a run-time efficiency benefit to compiling them differently, though?)

Yeah, this is all a bit of a mess at the moment. I'll try to disentangle it.

mtikekar:
I had noticed the multiple recompile problem too. From what I saw, unnecessary libraries specified at link time don't show up as run-time dependencies. So the original approach of specifying all libraries for bsc/bluetcl/bluewish is fine. 

Also, I did not look at it very carefully, but I think it is differences in include flags that trigger recompilation; different link flags are (mostly) fine.

PR #115, opened 2020-03-11
Reduce -fmax-pmcheck-iterations for CI builds so they complete faster.

bpfoley:


quark17:
Isn't the CI where we want to do checks?  I had increased the value, to try to reduce the number of warnings.  But I have to admit to not understanding what this flag is checking and whether increasing the number is the right way to deal with it.  Can the offending code be better written to not need this many iterations?  And in cases where the check gives up early, what is the potential error in the code that we're not detecting?

bpfoley:
> Isn't the CI where we want to do checks? I had increased the value, to try to reduce the number of warnings. But I have to admit to not understanding what this flag is checking and whether increasing the number is the right way to deal with it. Can the offending code be better written to not need this many iterations? And in cases where the check gives up early, what is the potential error in the code that we're not detecting?

It's doing exhaustivity checking on pattern matching. Unfortunately this seems to be exponential in constructor argument count:

https://stackoverflow.com/questions/40879983/what-sort-of-iterations-does-ghc-do-while-compiling-patterns

quark17:
Rather than giving the CI builds a smaller value, should we just change the default value to be smaller then?  (Or leave the flag off, using GHC's default -- potentially allowing it to be added if `GHCCHECKITERS` is defined?)

bpfoley:
> Rather than giving the CI builds a smaller value, should we just change the default value to be smaller then? (Or leave the flag off, using GHC's default -- potentially allowing it to be added if `GHCCHECKITERS` is defined?)

OK. We can do that, and just leave it up to the user to do `make GHC=ghc -fmax-pmcheck-iterations=...` if that's what they want.

PR #121, opened 2020-03-12
IIlineFmt: Re-add code accidentally removed during refactor in #107

bpfoley:
This caused an internal compiler error for some inputs.


quark17:
Oops, sorry for not coordinating better, I already pushed this fix

PR #122, opened 2020-03-13
Upload artifacts

mtikekar:
Upload build artifacts. I ran into https://github.com/actions/upload-artifact/issues/38 so I first tar the inst folder and then Github zips the tar.gz. Once that issue is fixed, we can upload the inst folder directly.

Also included another small change to allow the macos-10.15 build to work on 10.13 and above.


quark17:
Thank you!

PR #124, opened 2020-03-15
Fix build on Debian 10+ and Ubuntu 19.04+

cyrozap:
itcl 4 on Debian 10+ and Ubuntu 19.04+ is currently [version 4.1.2](https://packages.debian.org/buster/amd64/tcl-itcl4/filelist), but itk 4 is [version 4.1.0](https://packages.debian.org/buster/amd64/tk-itk4/filelist), so the build fails when it tries to link to libitk4.1.2. To fix this, if the build detects itcl 4, only link using the major version of the libraries (which symlink to the full version numbers).

I implemented this the best way I could think of, but if anyone has any better ideas, I'd be happy to try them.

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -49,8 +49,14 @@ BUILDDIR=$(TOP)/build
 PARSEC_HS = ../Parsec
 
 # Tcl
-TCL_VER   = $(shell echo 'puts [info tclversion]' | tclsh)
-ITCL_VER  = $(shell echo 'puts [package require Itcl]' | tclsh)
+TCL_VER        = $(shell echo 'puts [info tclversion]' | tclsh)
+ITCL_VER_FULL  = $(shell echo 'puts [package require Itcl]' | tclsh)
+ITCL_VER_MAJ   = $(shell echo $(ITCL_VER_FULL) | sed -e 's/^\([0-9]\).*/\1/')
+ifeq ($(ITCL_VER_MAJ), 4)
+ITCL_VER = $(ITCL_VER_MAJ)
+else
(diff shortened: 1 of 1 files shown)


quark17:
If I understand, the problem is that we're getting the itcl version and using that not only for the link name of itcl:

`TCL_LIBS = ... -litcl$(ITCL_VER) ...`

but we also use it for the link name of itk:

`WISHFLAGS = -litk$(ITCL_VER) ...`

Is there a reason that we can't separately get the itk version?:

`ITK_VER = $(shell echo 'puts [package require Itk]' | tclsh)`

@bpfoley ?  It looks like this command opens the DISPLAY, so maybe that was why it wasn't used?

I'll go ahead an accept this change, and we can later change to something cleaner if we want.

cyrozap:
@quark17 You're exactly right on all counts: The root of the problem is that the itcl version is being used for itk, but we can't get the itk version the same way as itcl because `require`-ing the package tries to open the DISPLAY (which of course fails on a headless build system).

I'm not familiar enough with Tcl/Tk to know how to get the version of the package without `require`-ing it, and any other solution I came up with would have to either use distro-specific methods to get the package version or use CMake (or some other fancy build system like Meson) to find the library, and at _that_ point you might as well re-write the build system for the whole project. So by comparison, this solution seemed much more concise, and more importantly, it was much less work!

bpfoley:
Sooo... the standard solution to this sort of thing is to use a headless X server like Xvfb if you happen to be running in an environment not running and X server already. I'll see if I can tweak the build to do this without too much damage.

bpfoley:
See #127 for the Xvfb implementation

PR #127, opened 2020-03-15
Explicitly check itk ver instead of assuming it's the same as itcl

bpfoley:
NB this will require xvfb to be installed for headless environments.
Update README and github build scripts to take this into account.

### DIFF
--- a/.github/workflows/install_dependencies_ubuntu.sh
+++ b/.github/workflows/install_dependencies_ubuntu.sh
@@ -21,4 +21,5 @@ sudo apt-get install -y \
   libghc-syb-dev \
   libghc-split-dev \
   libx11-dev \
-  libxft-dev
+  libxft-dev \
+  xvfb
--- a/README.md
+++ b/README.md
@@ -106,14 +106,15 @@ BSC builds with the latest version at the time of this writing, which is 8.8.2.
 ### Additional requirements
 
 For building the Bluespec Tcl/Tk shell, you will need the `tcl`, `tk`,
-`fontconfig` and `Xft` libraries:
+`fontconfig` and `Xft` libraries, and `Xvfb`:
 
     $ apt-get install \
         tcl-dev \
         tk-dev \
         libfontconfig1-dev \
         libx11-dev \
-        libxft-dev
+        libxft-dev \
+        xvfb
 
 The tcl shell also requires the `itcl` and `itk` libraries. For Debian 8,
 Debian 9, Ubuntu 16.04, and Ubuntu 18.04 version 3.4 of these libraries
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -45,21 +45,17 @@ BUILDDIR=$(TOP)/build
 # Parsec
 PARSEC_HS = ../Parsec
 
-#Have makefile avoid brew's install of tcl on Mac
+# Have Makefile avoid Homebrew's install of tcl on Mac
 ifeq ($(OSTYPE), Darwin)
-TCL_VER        = $(shell echo 'puts [info tclversion]' | /usr/bin/tclsh)
-ITCL_VER_FULL  = $(shell echo 'puts [package require Itcl]' | /usr/bin/tclsh)
+TCL_VER  = $(shell echo 'puts [info tclversion]' | /usr/bin/tclsh)
+ITCL_VER = $(shell echo 'puts [package require Itcl]' | /usr/bin/tclsh)
+ITK_VER  = $(shell echo 'catch {puts [package require Itk]};exit' | /usr/bin/wish)
 else
-TCL_VER        = $(shell echo 'puts [info tclversion]' | tclsh)
-ITCL_VER_FULL  = $(shell echo 'puts [package require Itcl]' | tclsh)
+TCL_VER  = $(shell echo 'puts [info tclversion]' | tclsh)
+ITCL_VER = $(shell echo 'puts [package require Itcl]' | tclsh)
+ITK_VER  = $(shell echo 'catch {puts [package require Itk]};exit' | xvfb-run -a /usr/bin/wish)
 endif
 
-ITCL_VER_MAJ   = $(shell echo $(ITCL_VER_FULL) | sed -e 's/^\([0-9]\).*/\1/')
-ifeq ($(ITCL_VER_MAJ), 4)
-ITCL_VER = $(ITCL_VER_MAJ)
-else
-ITCL_VER = $(ITCL_VER_FULL)
-endif
 TCL_HS    = ../vendor/htcl
 TCL_ARGS  = -L$(TCL_HS) $(shell pkg-config --silence-errors --cflags-only-I tcl tk || echo -I/usr/include/tcl)
 TCL_LIBS  = -ltcl$(TCL_VER) -ltk$(TCL_VER) -litcl$(ITCL_VER) \
@@ -92,10 +88,10 @@ endif
 
 EXTRAWISHLIBS = $(shell pkg-config --libs fontconfig xft)
 
-WISHFLAGS = -litk$(ITCL_VER) $(shell pkg-config --libs x11)
+WISHFLAGS = -litk$(ITK_VER) $(shell pkg-config --libs x11)
 WISHFLAGS += $(EXTRAWISHLIBS)
 ifeq ($(OSTYPE), Darwin)
-WISHFLAGS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitk$(ITCL_VER).dylib))"
+WISHFLAGS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitk$(ITK_VER).dylib))"
 endif
 
 # -----


cyrozap:
I have a few thoughts on this:

1. Looking at the CI failure, I guess maybe itk doesn't print its version when it's loaded?
2. If you're using this method to determine the full version of itk, you can revert my commit and remove the `ITK_VER_FULL` (etc.) lines and just use `ITK_VER` the way `ITCL_VER` was being used before. Otherwise, this doesn't provide any benefit over my fix for the original problem, other than being "more correct"--the itk and itcl major versions are most likely always going to be the same.
3. Is this really a good idea? I feel like determining if an X server needs to be spawned (it can't just check if an X server is already running--it needs to know that the current user has permission to access that server as well), spawning the new X server, and making sure the display number of the spawned X server doesn't conflict with any running X servers on the machine adds way too much complexity for the benefit it gives (being able to link to the exact version of itk that exists on the machine, instead of just using the major version, which probably needs to match itcl's anyways). The build process is already pretty complex as it is (having to build and link Haskell and C++ code), and making it more complex will only increase the likelihood that it will fail for new users, who will then open tickets because it doesn't build on their systems, which will then require the build process to be even more complex to handle those new cases. And in that case, like I said in #124, if the build system is being made much more complex anyways, it might as well be rewritten so the added complexity can at least be handled in a comprehensible manner.

bpfoley:
> I have a few thoughts on this:
> 
>     1. Looking at the CI failure, I guess maybe itk doesn't print its version when it's loaded?

It seems to be an issue in the version of xvfb used in 16.04 where it doesn't shut down in a timely manner.  The solution is simply to let `xvfb-run` find and allocate a free `$DISPLAY` number itself.

>     2. If you're using this method to determine the full version of itk, you can revert my commit and remove the `ITK_VER_FULL` (etc.) lines and just use `ITK_VER` the way `ITCL_VER` was being used before. Otherwise, this doesn't provide any benefit over my fix for the original problem, other than being "more correct"--the itk and itcl major versions are most likely always going to be the same.

Oh, OK. I think that escaped my attention. I'll undo the `VER_FULL` stuff so.

>     3. Is this really a good idea? I feel like determining if an X server needs to be spawned (it can't just check if an X server is already running--it needs to know that the current user has permission to access that server as well), spawning the new X server, and making sure the display number of the spawned X server doesn't conflict with any running X servers on the machine adds way too much complexity for the benefit it gives (being able to link to the exact version of itk that exists on the machine, instead of just using the major version, which probably needs to match itcl's anyways).

Thankfully that logic is mostly handled by `xvfb-run --auto-servernum` itself, so we don't have to do anything special.

>  The build process is already pretty complex as it is (having to build and link Haskell and C++ code), and making it more complex will only increase the likelihood that it will fail for new users, who will then open tickets because it doesn't build on their systems, which will then require the build process to be even more complex to handle those new cases. And in that case, like I said in #124, if the build system is being made much more complex anyways, it might as well be rewritten so the added complexity can at least be handled in a comprehensible manner.

I agree that the build is relatively complicated, but I don't feel up to a full scale rewrite at the moment (although I believe @thoughtpolice is looking into it), so I'm treating this a stopgap to fix the version detection issues.

quark17:
I feel similar to @cyrozap, that this seems to complicate things, and we already implemented a reasonable enough stopgap (to assume that the major version of ITk is the same, just not the minor version, and that linking with just the major version will work).

There may be a benefit to detecting the lack of ITk early (if a failure to require the package results in an early failure).

I might be willing to merge this anyway, but I'm concerned about the fact that calling "wish" uses X  ... when X exists.  This PR is about fixing the cases when X doesn't exist, but I wasn't even thinking about that when I made my initial comments to @cyrozap on the earlier PR (to use the major version).

On my mac (which has X but doesn't have itk, among other things that would be needed to build bsc), when I run this from the shell:

`echo 'puts [package require Itk];exit' | wish`

it both prints to stdout:

`can't find package Itk`

and opens a window (just a gray square) that doesn't go away until I kill it.  When I put that in a Makefile, it does the same thing: at the point where the value is needed, a window pops up and things hang until I manually close the window.

Is there way that we can avoid manual intervention by the user on machines that have X (and preferably not even open a window, even if it does close automatically)?

bpfoley:
> I feel similar to @cyrozap, that this seems to complicate things, and we already implemented a reasonable enough stopgap (to assume that the major version of ITk is the same, just not the minor version, and that linking with just the major version will work).
> 
> There may be a benefit to detecting the lack of ITk early (if a failure to require the package results in an early failure).
> 
> I might be willing to merge this anyway, but I'm concerned about the fact that calling "wish" uses X ... when X exists. This PR is about fixing the cases when X doesn't exist, but I wasn't even thinking about that when I made my initial comments to @cyrozap on the earlier PR (to use the major version).

I maybe be a bit confused, but I think there are two independent issues here. One, which I'm not trying to address at all at the moment, is building bluespec without any dependencies on Tk or X. I imagine we'll need some some sort of build flag to not build bluewish, and to ifdef out the use of Tk, Itk, X11. The second issue is detecting the version of Itk, preferably without a running X server.  That's the one I'm trying to fix here. Note that xvfb is not a particularly unusual requirement. In fact, as I discovered when looking at the build logs, it's already installed by default in the CI Ubuntu images.

> On my mac (which has X but doesn't have itk, among other things that would be needed to build bsc), when I run this from the shell:
> 
> `echo 'puts [package require Itk];exit' | wish`
> 
> it both prints to stdout:
> 
> `can't find package Itk`
> 
> and opens a window (just a gray square) that doesn't go away until I kill it. When I put that in a Makefile, it does the same thing: at the point where the value is needed, a window pops up and things hang until I manually close the window.
> 
> Is there way that we can avoid manual intervention by the user on machines that have X (and preferably not even open a window, even if it does close automatically)?

Oh. Nasty. On my system where I do have Itk, it exits successfully, but we need to add error handling for the failure case. 

```echo 'catch {puts [package require NonexistentPAackage]};exit' | wish```

does the job.

quark17:
Aha, adding "catch" fixes it. Great!

And, sorry, I wasn't being clear: My concern was that calling `wish` had the potential to display something on the X server (such as the gray window that popped up for me).  When I talked about X existing or not, I meant whether an X server or DISPLAY existed to connect to.

I do have one more suggestion on something to change:  Before checking whether a DISPLAY exists, there's an if-else that checks whether the OS is Darwin or not (and then doesn't do anything in the case of Darwin).  I think we should remove that and do the check in any case.  On my Mac, if I get this behavior without DISPLAY set in my environment:

`$ unsetenv DISPLAY`
`$ echo 'catch {puts [package require NonexistentPAackage]};exit' | wish`
`application-specific initialization failed: no display name and no $DISPLAY environment variable`

It is interesting that the CI build on Mac is passing, without the use of `xvfb`.  Could that build environment have a DISPLAY?  (Can you dump some info in the Makefile to see?)

bpfoley:
> Aha, adding "catch" fixes it. Great!
> 
> And, sorry, I wasn't being clear: My concern was that calling `wish` had the potential to display something on the X server (such as the gray window that popped up for me). When I talked about X existing or not, I meant whether an X server or DISPLAY existed to connect to.

Hmm. That's true. And if you're doing a build in the background on your workstation, that has the potential to steal focus from the window you're currently typing into. So maybe in that case we should just unconditionally use xvfb just in case the build of wish happens to use X. 

> I do have one more suggestion on something to change: Before checking whether a DISPLAY exists, there's an if-else that checks whether the OS is Darwin or not (and then doesn't do anything in the case of Darwin). I think we should remove that and do the check in any case. On my Mac, if I get this behavior without DISPLAY set in my environment:
> 
> `$ unsetenv DISPLAY`
> `$ echo 'catch {puts [package require NonexistentPAackage]};exit' | wish`
> `application-specific initialization failed: no display name and no $DISPLAY environment variable`

That's kinda surprising to me. The System version of wish on macOS isn't X based. I wonder have you installed a version of Tk that does use X? In any case, that argues further in favor of always using xvfb. At the worst it's just an extra build-time dependency and a very minor build overhead.

quark17:
Oh, indeed.  The `wish` being used was from MacPorts.  We support a particular mac setup, and I need to remember that my mac doesn't match that.  Incidentally, if I do the same experiment but pipe it through the system wish (`/usr/bin/wish`), I get this:

`DEPRECATION WARNING: The system version of Tk is deprecated and may be removed in a future release. Please don't rely on it. Set TK_SILENCE_DEPRECATION=1 to suppress this warning.`

bpfoley:
> `DEPRECATION WARNING: The system version of Tk is deprecated and may be removed in a future release. Please don't rely on it. Set TK_SILENCE_DEPRECATION=1 to suppress this warning.`

Oh man, don't get me started with Catalina (macOS 10.15). It's got so many deprecation warnings, it's really annoying. The other prominent thing it warns about is bash.

bpfoley:
Rather maddeningly, XQuartz on macOS does have Xvfb, but doesn't have the run-xvfb wrapper script that we use to find a free port, and it looks like it adds more annoying complexity to re-implement this for ourselves.

I think the solution is to merge #138 which forces the macOS build to use the system version of tcl, and to update this PR to do the same thing with wish. That way we don't need to worry about X servers on macOS.

PR #129, opened 2020-03-17
README: Add cabal update command required after new install of cabal

bpfoley:


quark17:
Did you run into a case where this was needed?  I have no objection to mentioning it, but I thought that installing the package does an update as part of the install.

bpfoley:
Yup. When I've been setting up fresh virtual machines to do builds in, each time the cabal install fails unless I do a cabal update first.

PR #135, opened 2020-03-20
fix link fault on Centos6 platform (-lpthread)

chenm001:
The build failed on Centos platform, I look into this issue, the reason is office GHC 8.2.2 necessary link with pthread library.

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -83,6 +83,10 @@ BSCBUILDLIBS = \
 	$(STP_LIBS) \
 	$(YICES_LIBS) \
 
+ifeq ($(OSTYPE), Linux)
+BSCBUILDLIBS += -lpthread
+endif
+
 EXTRAWISHLIBS = $(shell pkg-config --libs fontconfig xft)
 
 WISHFLAGS = -litk$(ITCL_VER) $(shell pkg-config --libs x11)


bpfoley:
Thanks for the pull request.

The macOS build failure here looks like a transient thing where a package failed to install correctly. I think it will go away on the next build.

What version of GHC are you using and where did you get it from? GHC itself should be providing the `-lpthread` option, and I'm curious why it's not.

Could you add a comment to say that the `-lpthread` is needed for CentOS 6?

quark17:
I was also curious what was requiring the flag.  My understanding, too, is that GHC needs it and that typically GHC is configured to automatically include it when needed, but in some cases it may not (whether because of the age of the GHC, a bug in certain versions, or misconfigured).  It's probably benign to explicitly give the flag, so I'm not opposed to adding it.  I'd want a comment, though instead of saying CentOS, maybe just say that some GHC installations need it explicitly.

On the other hand, there was some idea of just making sure that we build on certain supported systems.  Note that RHEL 6 reaches retirement this year and CentOS 6 is marking that as its end-of-life.

chenm001:
> Thanks for the pull request.
> 
> The macOS build failure here looks like a transient thing where a package failed to install correctly. I think it will go away on the next build.
> 
> What version of GHC are you using and where did you get it from? GHC itself should be providing the `-lpthread` option, and I'm curious why it's not.
> 
> Could you add a comment to say that the `-lpthread` is needed for CentOS 6?

I download from https://www.haskell.org/ghc/download_ghc_8_2_2.html , it is latest version for Centos 6, the newer version need Centos 7.
The bsc latest link stage didn't include pthread, so I got a link error from a GHC lib (undefined symbol pthread_setname_np@GLIBC_2_12).

chenm001:
> I was also curious what was requiring the flag. My understanding, too, is that GHC needs it and that typically GHC is configured to automatically include it when needed, but in some cases it may not (whether because of the age of the GHC, a bug in certain versions, or misconfigured). It's probably benign to explicitly give the flag, so I'm not opposed to adding it. I'd want a comment, though instead of saying CentOS, maybe just say that some GHC installations need it explicitly.
> 
> On the other hand, there was some idea of just making sure that we build on certain supported systems. Note that RHEL 6 reaches retirement this year and CentOS 6 is marking that as its end-of-life.

Thank you point out, I feeling GHC became fragmentation.
Both RHEL6 and RHEL7 are not proivde GHC 7.10.x, the latest EPEL6 version is 7.0.4 and EPEL7 is 7.6.4.

btw: I also like Debian/Ubuntu Linux, but it is not EDA friendly, for example, Synopsys tools will be crash in most time on them, some time crash on RHEL7, but work fine on RHEL6, I trace into to find out the issues, it sounds GLIBC's data struct mismatch as product's.

PR #143, opened 2020-04-02
Remove compiler-script.sh command line filter.

bpfoley:
Realistically, we don't care about the versions of GCC it supports
(GCC <4.4), because we require C++11 support (GCC >=4.8) elsewhere,
and it causes some versions of GHC on Linux to break on certain
systems.

In particular it interacts badly with systems where -fPIE is the
C toolchain default and the GHC setting "C compiler supports-no-pie"
is "YES", because it stops GHC passing -no-pie as CFLAGS and LDFLAGS.

See:
https://gitlab.haskell.org/ghc/ghc/issues/17919
https://github.com/B-Lang-org/bsc/issues/109

Thanks to @leonschoorl for tracking this down.


src/comp/Makefile
```diff
@@ -197,10 +194,7 @@ GHCFLAGS = \
 	$(GHCTMP) \
 	$(TCL_ARGS) \
 	$(STP_ARGS) \
-	$(YICES_ARGS) \
-	'-pgma ./compiler-script.sh' \
-	'-pgmc ./compiler-script.sh' \
-	'-pgml ./compiler-script.sh'
+	$(YICES_ARGS)
```
quark17, line 197: I tend to leave trailing slashes on the last line of a list like this (and making sure the next line remains blank), so that adding or removing items from the list doesn't change the history for unrelated lines.

bpfoley: ACK. Done.

PR #144, opened 2020-04-05
Tidy up Tk lib/args variables

bpfoley:
As a micro-optimization this lets us merge two pkg-conifg invocations

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -45,56 +45,51 @@ BUILDDIR=$(TOP)/build
 # Parsec
 PARSEC_HS = ../Parsec
 
-# Have Makefile avoid Homebrew's install of tcl on Mac
-ifeq ($(OSTYPE), Darwin)
-TCL_VER  = $(shell echo 'puts [info tclversion]' | /usr/bin/tclsh)
-ITCL_VER = $(shell echo 'puts [package require Itcl]' | /usr/bin/tclsh)
-ITK_VER  = $(shell echo 'catch {puts [package require Itk]};exit' | /usr/bin/wish)
-else
-TCL_VER  = $(shell echo 'puts [info tclversion]' | tclsh)
-ITCL_VER = $(shell echo 'puts [package require Itcl]' | tclsh)
-ITK_VER  = $(shell echo 'catch {puts [package require Itk]};exit' | xvfb-run -a /usr/bin/wish)
-endif
-
-TCL_HS    = ../vendor/htcl
-TCL_ARGS  = -L$(TCL_HS) $(shell pkg-config --silence-errors --cflags-only-I tcl tk || echo -I/usr/include/tcl)
-TCL_LIBS  = -ltcl$(TCL_VER) -ltk$(TCL_VER) -litcl$(ITCL_VER) \
-            -lhtcl
-ifeq ($(OSTYPE), Darwin)
-TCL_ARGS += -I$(shell xcrun --show-sdk-path)/System/Library/Frameworks/Tk.framework/Headers
-TCL_LIBS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitcl$(ITCL_VER).dylib))"
-endif
-
 # STP
-STP_HS      = ../vendor/stp/include_hs
-STP_ARGS = -I../vendor/stp/include -L../vendor/stp/lib
-STP_LIBS = -lstp
+STP_HS = ../vendor/stp/include_hs
+STP_INC_FLAGS = -I../vendor/stp/include
+STP_LIB_FLAGS = -L../vendor/stp/lib -lstp
 
 # Yices
-YICES_HS      = ../vendor/yices/include_hs
-YICES_ARGS = -I../vendor/yices/include -L../vendor/yices/lib
-YICES_LIBS = -lyices
+YICES_HS = ../vendor/yices/include_hs
+YICES_INC_FLAGS = -I../vendor/yices/include
+YICES_LIB_FLAGS = -L../vendor/yices/lib -lyices
 
-# Libs which are only needed to build bsc/bluetcl
-# and which shouldn't be used to build other Haskell binaries
-BSCBUILDLIBS = \
-	$(TCL_LIBS) \
-	$(STP_LIBS) \
-	$(YICES_LIBS) \
+# HTCL
+HTCL_HS = ../vendor/htcl
+HTCL_INC_FLAGS = -L$(HTCL_HS)
+HTCL_LIB_FLAGS = -lhtcl
 
-ifeq ($(OSTYPE), Linux)
-# Some GHC installations may fail to include this automatically,
-# and it's benign to include otherwise
-BSCBUILDLIBS += -lpthread
-endif
-
-EXTRAWISHLIBS = $(shell pkg-config --libs fontconfig xft)
-
-WISHFLAGS = -litk$(ITK_VER) $(shell pkg-config --libs x11)
-WISHFLAGS += $(EXTRAWISHLIBS)
+# TCL/TK versions
+# TCL and TK are always the same, but ITCL/ITK may have different minor versions
 ifeq ($(OSTYPE), Darwin)
-WISHFLAGS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitk$(ITK_VER).dylib))"
+# Have Makefile avoid Homebrew's install of tcl on Mac
+TCLTK_VER = $(shell echo 'puts [info tclversion]' | /usr/bin/tclsh)
+ITCL_VER  = $(shell echo 'puts [package require Itcl]' | /usr/bin/tclsh)
+ITK_VER   = $(shell echo 'catch {puts [package require Itk]};exit' | /usr/bin/wish)
+else
+TCLTK_VER = $(shell echo 'puts [info tclversion]' | tclsh)
+ITCL_VER  = $(shell echo 'puts [package require Itcl]' | tclsh)
+ITK_VER   = $(shell echo 'catch {puts [package require Itk]};exit' | xvfb-run -a /usr/bin/wish)
 endif
+
+# TCL/TK include flags
+TCLTK_INC_FLAGS = $(shell pkg-config --silence-errors --cflags-only-I tcl tk || echo -I/usr/include/tcl)
+ifeq ($(OSTYPE), Darwin)
+TCLTK_INC_FLAGS += -I$(shell xcrun --show-sdk-path)/System/Library/Frameworks/Tk.framework/Headers
+endif
+
+# TCL library flags
+ifeq ($(OSTYPE), Darwin)
+TCL_LIB_FLAGS = -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitcl$(ITCL_VER).dylib))"
+endif
+TCL_LIB_FLAGS += -ltcl$(TCLTK_VER) -litcl$(ITCL_VER)
+
+# TK library flags
+ifeq ($(OSTYPE), Darwin)
+TK_LIB_FLAGS = -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitk$(ITK_VER).dylib))"
+endif
+TK_LIB_FLAGS += -ltk$(TCLTK_VER) -litk$(ITK_VER) $(shell pkg-config --libs x11 fontconfig xft)
 
 # -----
 # GHC
@@ -187,7 +182,7 @@ GHCINCLUDES =  \
 	-i$(PARSEC_HS) \
 	-i$(STP_HS) \
 	-i$(YICES_HS) \
-	-i$(TCL_HS)
+	-i$(HTCL_HS)
 GHCTMP = '-tmpdir $(TMPDIR)'
 FVIA ?= -fasm
 GHCFLAGS = \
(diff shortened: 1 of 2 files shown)


src/comp/Makefile
```diff
@@ -88,13 +94,7 @@ ifeq ($(OSTYPE), Linux)
...
 endif
 
-EXTRAWISHLIBS = $(shell pkg-config --libs fontconfig xft)
-
-WISHFLAGS = -litk$(ITK_VER) $(shell pkg-config --libs x11)
-WISHFLAGS += $(EXTRAWISHLIBS)
-ifeq ($(OSTYPE), Darwin)
-WISHFLAGS += -L"$(shell dirname $(shell $(FIND) /System/Library/Tcl -type f -name libitk$(ITK_VER).dylib))"
-endif
+WISHFLAGS=$(TK_ARGS) $(TK_LIBS)
```
quark17, line 97: Are both TK_ARGS and TK_LIBS needed in all the places where WISHFLAGS is used?  Or would it be worth defining like this (to parallel definitions like BSCBUILDLIBS):

`WISHBUILDFLAGS=$(TK_ARGS)`
`WISHBUILDLIBS=$(TK_LIBS)`

Though, while there is BSCBUILDLIBS, there is no BSCBUILDFLAGS.  Instead, it looks like there's one set of flags that gets used to build all the tools (even though possibly some of those tools don't need all of flags that BSC does).

That is, if you look at the `bsc`, `dumpbo`, etc targets, you'll see `BSCBUILDLIBS` are libraries that only get added to the `bsc` command line.  And `WISHFLAGS` is only added to the `bluewish` (and `bluetcl`) targets -- so that's why I suggest renaming the library part to `WISHBUILDLIBS`.  But then, I notice that all of the targets use the same `BUILDFLAGS` (including `bluetcl` and `bluewish`), and that is defined to include `GHCFLAGS` which is where the `TCL_ARGS` (and `STP_ARGS` and `YICES_ARGS`) are added.  It seems like maybe those should not be in `GHCFLAGS` but should be assigned to a separate `BSCBUILDFLAGS` that's only given to the `bsc` target.  Or, if there's some reason they need to be there, then why shouldn't the `TK_ARGS` also be there.

Separately, I notice that the `bluetcl` and `bluewish` targets run `BUILDCOMMAND` twice.  I suspect that only the TK_ARGS are needed in the first call (not the libs).

bpfoley: This feels a little bit 'angels dancing on the head of a pin'. What are we trying to achieve here?

`TK_ARGS` (currently) only has `-L` args to tell ghc where to find shared libraries, and `TK_LIBS` (currently) only has `-l` args to tell ghc which libraries to link with. As far as I can see, every place you provide `-l...` you're going to have to provide `-L...` too; and conversely, if you don't provide `-l...` you don't need `-L...` either. So shouldn't they just be one variable?

quark17: If there's an easy cleanup here, I'd like to apply it, but certainly it's not worth spending much time (particularly if this will all eventually get replaced with some better build mechanism).

I think I'm also asking about many different things at once.  So let me try to pull them apart.

I think your point about `-L` and `-l` is good, that they are only used together.  I was thinking just about `-I`.  But I think you've clarified something for me: For TCL, STP, and YICES, we define separate `_ARGS` and `_LIBS`, where the `_ARGS` are the `-I` and `-L` flags that tend to come early on the command line and the `_LIBS` are the `-l` flags that come later (because their order can matter, at least on gcc command lines).

But your point comes into play: any place that uses `STP_LIBS` must have `STP_ARGS` and other places don't need it.  So why then are `TCL_ARGS`, `STP_ARGS`, and `YICES_ARGS` added into `GHCFLAGS` which gets used in all the targets (in all calls to GHC in this Makefile), while the corresponding `TCL_LIBS`, `STP_LIBS`, and `YICES_LIBS` are only used in the `bsc`, `bluetcl`, and `bluewish` targets (as `BSCBUILDLIBS`)?

Item 1: I would propose that we remove `TCL_ARGS`, `STP_ARGS`, and `YICES_ARGS` from `GHCFLAGS` and instead put them in a new variable, `BSCBUILDFLAGS`, which is only used in the GHC command in the `bsc`, `bluetcl`, and `bluewish` targets (since that's also the only places where `BSCBUILDLIBS` gets used).

Item 2: The other point you're making is that, if the `_ARGS` and the `_LIBS` are always used together, then why bother defining separate variables?  For example, here's part of the `bsc` target as it currently exists:

`$(BUILDCOMMAND) -main-is Main_$@ \`
`    $(BUILDFLAGS) $(BSCBUILDLIBS) $(EXTRAOBJ)`

My "Item 1" proposal would change that to this:

`$(BUILDCOMMAND) -main-is Main_$@ \`
`    $(BUILDFLAGS) $(BSCBUILDFLAGS) $(BSCBUILDLIBS) $(EXTRAOBJ)`

where `BSCBUILDFLAGS` is now defined like this:

`BSCBUILDFLAGS = \`
`    $(TCL_ARGS) \`
`    $(STP_ARGS) \`
`    $(YICES_ARGS) \`

But if there's no reason to have the `_ARGS` separated from the `_LIBS`, then we could eliminate the separate `BSCBUILDLIBS` variable and just put everything in `BSCBUILDFLAGS`:

`$(BUILDCOMMAND) -main-is Main_$@ \`
`    $(BUILDFLAGS) $(BSCBUILDFLAGS) $(EXTRAOBJ)`

(The only reason that I can imagine putting the `_ARGS` in all targets is to avoid recompilation due to flags changing.  Do `-I`, `-L`, `-l` flags get considered in that decision?)

Item 3: Whatever we decide to do for Items 1&2 (regarding the libraries for `bsc`) we should apply the same convention to the bluewish libraries.  These are analagous things, so we should treat them symmetrically, so that it's clear when reading the Makefile.  Specifically, we should rename `WISHFLAGS` to be `WISHBUILDFLAGS` (to match `BSCBUILDFLAGS` and `BUILDFLAGS`).  And if for Item 2 we decide to have a separate `BSCBUILDLIBS`, then we should also have a separate `WISHBUILDLIBS` (but if we don't have a separate variable for bsc, then we shouldn't have it for bluewish either).

src/comp/Makefile
```diff
@@ -55,7 +56,6 @@ TCL_VER  = $(shell echo 'puts [info tclversion]' | tclsh)
 ITCL_VER = $(shell echo 'puts [package require Itcl]' | tclsh)
 ITK_VER  = $(shell echo 'catch {puts [package require Itk]};exit' | xvfb-run -a /usr/bin/wish)
 endif
-
 TCL_HS    = ../vendor/htcl
 TCL_ARGS  = -L$(TCL_HS) $(shell pkg-config --silence-errors --cflags-only-I tcl tk || echo -I/usr/include/tcl)
 TCL_LIBS  = -ltcl$(TCL_VER) -ltk$(TCL_VER) -litcl$(ITCL_VER) \
```
quark17, line 61: Should `-ltk$(TCL_VER)` be moved to the `TK_LIBS` down below?

Or ... actually ... right now `bluetcl` and `bluewish` build with the same flags.  We really ought to remove the Tk flags from the `bluetcl` build, so that it doesn't require X libraries when run -- but, until then, we could put both the TCL and TK flags into one variable.

But the bigger realization is that `bsc` doesn't need any of the Tcl flags.  Unless there's a reason to keep the flags the same to avoid recompilation, we should remove the `TCL_ARGS` and `TCL_LIBS` from the variables used in the `bsc` target (like `BSCBUILDLIBS`).  And then maybe define `BLUETCLBUILDFLAGS` to hold it.  And the current `WISHFLAGS` could be put into that (until the time when we can separate the `BLUEWISHBUILDFLAGS` out from it).

src/comp/Makefile
```diff
@@ -45,8 +45,9 @@ BUILDDIR=$(TOP)/build
 # Parsec
 PARSEC_HS = ../Parsec
 
-# Have Makefile avoid Homebrew's install of tcl on Mac
+# TCL
 ifeq ($(OSTYPE), Darwin)
+# Have Makefile avoid Homebrew's install of tcl on Mac
 TCL_VER  = $(shell echo 'puts [info tclversion]' | /usr/bin/tclsh)
 ITCL_VER = $(shell echo 'puts [package require Itcl]' | /usr/bin/tclsh)
 ITK_VER  = $(shell echo 'catch {puts [package require Itk]};exit' | /usr/bin/wish)
```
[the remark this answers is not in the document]
bpfoley: Done.

src/comp/Makefile
bpfoley: So, as far as I can tell, ghc does use `-I` and `-L` as part of its decision about whether to recompile.

http://hackage.haskell.org/package/ghc-8.6.5/docs/DynFlags.html#t:DynFlags

quark17: I see that `-I` and `-L` are both in `DynFlags`, but how are you making the leap that `DynFlags` is entirely part of the recompilation decision?  My experimentation with GHC 8.4.4 shows that `-I` matters but that `-L` does not.  Looking at GHC source, I see that there's a package `FlagChecker` which generates fingerprints for a number of flag types (not just `DynFlags`) and that, for the `DynFlags`, only certain flags are considered (see `fingerprintDynFlags`).  It seems to only consider `-I` because it can affect the operation of CPP.  Anyway, I don't know what all can be relied on in the source; I'd prefer to see a passage in the documentation that describes when recompilation occurs.

quark17:
I'm guessing that you're not seeing my out-of-band emails.  I've been working on my own version of the Makefile.  You seem to have arrived at a similar point, although I have some further cleanups and made different decisions in places.  I'm attaching my Makefile here -- let me know what you think, and see if we can converge.

[Makefile](https://github.com/B-Lang-org/bsc/files/4435508/Makefile.txt)
[bluetcl_Main.hsc](https://github.com/B-Lang-org/bsc/files/4435509/bluetcl_Main.hsc.txt)

With this Makefile, I can run `ldd` on the generated `bsc` binary and see that it no longer requires any Tcl or Tk/X libraries from the system!  And, similarly, the `bluetcl` binary no longer requires any Tk/X!  For this to work, though, I had to remove the few Tk references from `bluetcl_Main.hsc`.  It appears that they were added at one time, to support some feature of the Workstation GUI, but I don't entirely see why.  (Is there some part of Tk that can be used without X?)

bpfoley:
> I'm guessing that you're not seeing my out-of-band emails. I've been working on my own version of the Makefile. You seem to have arrived at a similar point, although I have some further cleanups and made different decisions in places. I'm attaching my Makefile here -- let me know what you think, and see if we can converge.
> 
> [Makefile](https://github.com/B-Lang-org/bsc/files/4435508/Makefile.txt)
> [bluetcl_Main.hsc](https://github.com/B-Lang-org/bsc/files/4435509/bluetcl_Main.hsc.txt)

Relative to my version of the Makefile you've

* changed FOO_ARGS and FOO_LIBS to FOO_ARG_FLAGS and FOO_LIB_FLAGS, moving -L... from _ARGS to _LIB_FLAGS. Good!
* split all the htcl args out of TCL_ARGS, TCL_LIBS into their own HTCL_* variables, and used it in GHCINCLUDES. Good!
* Fused TCLTK_INC_FLAGS. Good!
* Distinguished between TCL_LIB_FLAGS and TK_LIB_FLAGS. Good!
* Moved BUILDFLAGS directly into GHCFLAGS; removed BSCBUILDLIBS, and added more focussed {BSC,BLUETCL,BLUEWISH}_BUILDLIBS. Good!
* Fixed a bug in tags where it would have looked in too many places. Good! :)

Your bluetcl_Main.hsc removes the Tcl_StaticPackage for Tk and Itk. Should we remove these from bluewish too?
 
> With this Makefile, I can run `ldd` on the generated `bsc` binary and see that it no longer requires any Tcl or Tk/X libraries from the system! And, similarly, the `bluetcl` binary no longer requires any Tk/X! For this to work, though, I had to remove the few Tk references from `bluetcl_Main.hsc`. It appears that they were added at one time, to support some feature of the Workstation GUI, but I don't entirely see why. (Is there some part of Tk that can be used without X?)

On both macOS and Ubuntu 20.04 I am seeing Tcl etc linked into bluetcl:

```
$ ldd bluetcl
	linux-vdso.so.1 (0x00007ffffafa7000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3322310000)
	libstp.so.1 => not found
	libyices.so.2.6 => not found
	libtcl8.6.so => /lib/x86_64-linux-gnu/libtcl8.6.so (0x00007f3322154000)
	libitcl4.so => /lib/x86_64-linux-gnu/libitcl4.so (0x00007f332210b000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f3322100000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f33220f9000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f33220f3000)
	libgmp.so.10 => /lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f332206f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3321f20000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3321d2e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f332233e000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f3321d12000)
```

```
$ otool -L bluetcl

bluetcl:
	libstp.dylib (compatibility version 0.0.0, current version 0.0.0)
	/Users/bfoley/opensource/bpfoley-bsc/src/vendor/yices/v2.6/yices2-inst/lib/libyices.2.dylib (compatibility version 2.6.0, current version 2.6.2)
	/System/Library/Frameworks/Tcl.framework/Versions/8.5/Tcl (compatibility version 8.5.0, current version 8.5.9)
	/System/Library/Tcl/8.5/itcl3.4/libitcl3.4.dylib (compatibility version 3.4.0, current version 3.4.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
	/usr/lib/libcharset.1.dylib (compatibility version 2.0.0, current version 2.0.0)
```

If you like I'll remove the extra Tcl_StaticPackages and upload a new patch with all the rest of your changes.

src/comp/Makefile
bpfoley: I couldn't find any authoritative documentation on this, but lacking that, your experiments makes sense. `-I` will, if nothing else, affect the `#include` in Haskell code if we use `{-# LANGUAGE CPP #-}`, whereas `-L` just affects linking, which only comes into play when building the final binaries.

bpfoley:
> If you like I'll remove the extra Tcl_StaticPackages and upload a new patch with all the rest of your changes.

Nevermind. I'm being an idiot. I realized what you did: you just removed the Tk/ITk (ie GUI) initialization from the non-GUI bluetcl. No extra work needed for bluewish.

bpfoley:
> On both macOS and Ubuntu 20.04 I am seeing Tcl etc linked into bluetcl:

Argh. Also nevermind this. Like you, I see `bsc` isn't linking against tcl/tk on macOS or Ubuntu. Great!

PR #146, opened 2020-04-11
Fix CFLAGS handling in htcl and build parallelism in stp

thotypous:
This patch fixes some build failures observed when packaging bsc.

* The htcl makefile tries to pass CFLAGS to ghc. This is going to fail when packaging for most distributions, since it is pretty common for package build tools to set CFLAGS.
```
make[2]: Entering directory '/home/matias/git/bluespec-git/src/bsc/src/vendor/htcl'
ghc -dynamic -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall  -c haskell.c
ghc: unrecognised flag: -march=x86-64
unrecognised flag: -mtune=generic
unrecognised flag: -pipe
did you mean one of:
  -pie
unrecognised flag: -fno-plt
did you mean one of:
  -fno-it
```

* **POSTPONED** Vanilla itcl installs to `/usr/lib/itcl$(ITCL_VER)` and vanilla itk installs to `/usr/lib/itk$(ITK_VER)`. Debian-based distributions move the libraries to `/usr/lib/x86_64-linux-gnu`. Other distributions, such as Arch Linux, keep the original paths. Here we pass these original paths to the `-L` and `-optl-Wl,-rpath` options. If the paths are non-existent (e.g. bsc is being built in Debian or Ubuntu) this appears not to cause any issues, so I think we don't need to check if the paths exist.

* There is a race condition in the stp sat makefile. Since `Solver.or` is part both of `core/libcore_release.a` and `simp/libsimp_release.a`, the dependencies of target `libminisat.a` cannot be built in parallel, otherwise `simp/libsimp_release.a` may be generated when `Solver.or` is not completely written to disk by gcc. As a result, `libminisat.a` gets a corrupt version of `Solver.or`. This happens quite often when building bsc with `-jN` make flags, and is such a pain because the build system will fail only very later on, when linking bsc:
```
/home/matias/git/bluespec-git/src/bsc/inst/bin/core/bsc: symbol lookup error: /home/main-builder/pkgwork/src/bsc/inst/lib/SAT/libstp.so.1: undefined symbol: _ZTIN7Minisat6SolverE
```

### DIFF
--- a/src/vendor/htcl/Makefile
+++ b/src/vendor/htcl/Makefile
@@ -1,9 +1,9 @@
-CFLAGS += -Wall $(shell pkg-config --silence-errors --cflags-only-I tcl || echo -I/usr/include/tcl)
+GHCFLAGS += -Wall $(shell pkg-config --silence-errors --cflags-only-I tcl || echo -I/usr/include/tcl)
 GHC ?= ghc
 
 # We use GHC to compile this, so it has the proper RTS includes
 %.o: %.c
-	$(GHC) $(CFLAGS) -c $<
+	$(GHC) $(GHCFLAGS) -c $<
 
 libhtcl.a: haskell.o
 	ar -r $@ $(filter %.o, $+)
--- a/src/vendor/stp/src/sat/Makefile
+++ b/src/vendor/stp/src/sat/Makefile
@@ -13,6 +13,7 @@ export COPTIMIZE=$(CFLAGS_M32) $(CFLAGS_FPIC) -O3
 core: $(LIB)
 
 # $(LIB) depends on */lib$(SUB)_release.a and will be rebuilt only if they have been updated
+.NOTPARALLEL:
 $(LIB): core/libcore_release.a core_prop/libcore_prop_release.a simp/libsimp_release.a utils/libutils_release.a cryptominisat2/libminisat.a $(OBJS)
 	$(RM) $@
 	$(call arcat,$@,$(filter %.a,$^))


bpfoley:
> This patch fixes some build failures observed when packaging bsc.

Awesome! Thanks for your contributions.

>     * The htcl makefile tries to pass CFLAGS to ghc. This is going to fail when packaging for most distributions, since it is pretty common for package build tools to set CFLAGS.
> 
> 
> ```
> make[2]: Entering directory '/home/matias/git/bluespec-git/src/bsc/src/vendor/htcl'
> ghc -dynamic -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall  -c haskell.c
> ghc: unrecognised flag: -march=x86-64
> unrecognised flag: -mtune=generic
> unrecognised flag: -pipe
> did you mean one of:
>   -pie
> unrecognised flag: -fno-plt
> did you mean one of:
>   -fno-it
> ```

Yes. This is clearly an error. We're using CFLAGS to mean the subset of flags that both the C compiler and GHC understand (namely `-L`, `-I`, `-I`, and `-i`). I think your fix is the correct thing to do.

> 
>     * Vanilla itcl installs to `/usr/lib/itcl$(ITCL_VER)` and vanilla itk installs to `/usr/lib/itk$(ITK_VER)`. Debian-based distributions move the libraries to `/usr/lib/x86_64-linux-gnu`. Other distributions, such as Arch Linux, keep the original paths. Here we pass these original paths to the `-L` and `-optl-Wl,-rpath` options. If the paths are non-existent (e.g. bsc is being built in Debian or Ubuntu) this appears not to cause any issues, so I think we don't need to check if the paths exist.

Perhaps I'm missing something, but I'm less convinced about this patch.

On Arch is `/usr/lib` not on ld.so's search path? If it is, why do you need the `-L/usr/lib...` and the `-optl-Wl,-rpath,/usr/lib/...`? What does `/etc/ld.so.conf` contain? And what does /sbin/ldconfig -p | grep itk` say?

>     * There is a race condition in the stp sat makefile. Since `Solver.or` is part both of `core/libcore_release.a` and `simp/libsimp_release.a`, the dependencies of target `libminisat.a` cannot be built in parallel, otherwise `simp/libsimp_release.a` may be generated when `Solver.or` is not completely written to disk by gcc. As a result, `libminisat.a` gets a corrupt version of `Solver.or`. This happens quite often when building bsc with `-jN` make flags, and is such a pain because the build system will fail only very later on, when linking bsc:

Ouch, nasty. We have a very old version of STP and upstream have competely rewritten their build system using CMake. I believe our slightly longer term plan is to remove STP entirely, but in the meantime, this is a good fix.

thotypous:
Thank you for the review!

> On Arch is `/usr/lib` not on ld.so's search path? If it is, why do you need the `-L/usr/lib...` and the `-optl-Wl,-rpath,/usr/lib/...`? What does `/etc/ld.so.conf` contain? And what does /sbin/ldconfig -p | grep itk` say?

`/etc/ld.so.conf` contains just `include /etc/ld.so.conf.d/*.conf` and `grep -Pri 'itcl|itk' /etc/ld.so.conf.d` returns nothing.

`/sbin/ldconfig -p | grep -Pi 'itcl|itk'` returns nothing.

I think the itcl and itk packages are originally intended to be loaded from a Tcl script using `[load]`, which then dynamically loads the `.so` library (probably using something like `dlopen`).

There is even a comment in their `pkgIndex.tcl` saying "There is no libitcl.so that will `[load]` into both Tcl 8 and Tcl 9." Probably they were looking for a way for incompatible versions to coexist in the same system.

Therefore I don't think `libitcl*.so` and `libitk*.so` were intended to be in ld.so search path at all.

As a side note, besides Arch Linux, I have observed the same layout in Alpine Linux - itcl is also at `/usr/lib/itcl$(ITCL_VER)` and is not included in ld musl search path. Unfortunately, I gave up (for now) packaging for Alpine since it looks like STP would need to be patched to work with musl libc.

bpfoley:
> Thank you for the review!
> 
> > On Arch is `/usr/lib` not on ld.so's search path? If it is, why do you need the `-L/usr/lib...` and the `-optl-Wl,-rpath,/usr/lib/...`? What does `/etc/ld.so.conf` contain? And what does /sbin/ldconfig -p | grep itk` say?
> 
> `/etc/ld.so.conf` contains just `include /etc/ld.so.conf.d/*.conf` and `grep -Pri 'itcl|itk' /etc/ld.so.conf.d` returns nothing, also ld does not search `/usr/lib` recursively.
> 
> `/sbin/ldconfig -p | grep -Pi 'itcl|itk'` returns nothing.
> 
> I think the itcl and itk packages are originally intended to be loaded from a Tcl script using `[load]`, which then dynamically loads the `.so` library (probably using something like `dlopen`).

Yes, that's right. You can see what actually happens behind the scenes with
```$ tclsh
% package require Itcl
3.4
% package ifneeded Itcl 3.4
load /usr/lib64/tcl8.5/itcl3.4/../../libitcl3.4.so Itcl
```

Unfortunately, `ifneeded` returns an arbitrary Tcl script that's meant to be `exec`d, so parsing this would be fragile.

> 
> There is even a comment in their `pkgIndex.tcl` saying "There is no libitcl.so that will `[load]` into both Tcl 8 and Tcl 9." Probably they were looking for a way for incompatible versions to coexist in the same system.
>
> Therefore I don't think `libitcl*.so` and `libitk*.so` were intended to be in ld.so search path at all.
> 
> As a side note, besides Arch Linux, I have observed the same layout in Alpine Linux - itcl is also at `/usr/lib/itcl$(ITCL_VER)` and is not included in ld musl search path. Unfortunately, I gave up (for now) packaging for Alpine since it looks like STP would need to be patched to work with musl libc.

Hmm. On my Ubuntu (20.04) VM I have
```
$ cat /etc/ld.so.conf
include /etc/ld.so.conf.d/*.conf

$ cat /etc/ld.so.conf.d/x86_64-linux-gnu.conf
# Multiarch support
/usr/local/lib/x86_64-linux-gnu
/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu

$ /sbin/ldconfig -p|grep itk
	libitk4.so (libc6,x86-64) => /lib/x86_64-linux-gnu/libitk4.so
```

Whereas on CentOS 6 I get
```
$ cat /etc/ld.so.conf
include ld.so.conf.d/*.conf

$ cat /etc/ld.so.conf.d/*.conf | grep -v '#^'
hwcap 1 nosegneg
/usr/lib64/mysql

$ /sbin/ldconfig -p|grep itk
	libitk3.4.so (libc6,x86-64) => /usr/lib64/libitk3.4.so
```

I have nagging reservations about unconditionally adding `/usr/lib` to the library path, as it feels like it's a distro specific concern, and as we can see, different distros make different choices. Would you be OK to keep this as an Arch-specific patch?

The GHC `CFLAGS` patch and the STP parallelism fix look good to me though. @quark17, any thoughts?

bpfoley:
And... just to make it explicit, on CentOS /usr/lib contains 32bit libraries (for backwards compatibility purposes), and /usr/lib64 has the x86-64 libraries. So I think on CentOS 6, explicitly adding /usr/lib to the search path will actually break 64-bit bsc binaries if you've also got 32-bit libraries installed.

quark17:
Both the CFLAGS and STP parallelism fixes look good.

On the itcl/itk location: I would prefer not to hardcode paths.  I'm guessing that `pkg-config` can't be used to compute the location here, though.  If it's the case that itcl/itk are supposed to be loaded from inside tcl, rather than loaded at execution, then maybe the real fix is to change bluetcl/bluewish to do that, and not mention it in the Makefile.  (Maybe it could even be done inside the init functions in the `.hsc` source?  Although it'd be worth identifying where itcl/itk are used -- if they're not used in the native bluetcl commands, then we could even eliminate this dependency, and just make it the responsibility of any scripts to load it when required. I admit to not knowing much about why itcl/itk was added or where it's used.)  This change would not only mean that the building of a release doesn't need itcl/itk, but it would change how itcl/itk is discovered in the environment of anyone using bluetcl (and would delay the point at which they would discover that itcl/itk can't be found) -- maybe that's a good thing.

In any case, PR #144 may be significantly reorganizing the tcl/tk/itcl/itk flags, so I'd wait until that is committed before applying any changes to those flags.

thotypous:
OK, so let's postpone the itcl/itk stuff. I will patch it on Arch Linux side in the mean time. I've force pushed a new patch fixing only the CFLAGS and build parallelism issues.

PR #147, opened 2020-04-11
Add Dockerfile

thotypous:
It would be superb if you could set up [automated builds](https://docs.docker.com/docker-hub/builds/) on Docker Hub using this.

An official image on Docker Hub would be very useful for setting up automated CI builds on Bluespec designs. It would also help to attract more newcomers (as an alternative or complement to ordinary binary releases).


bpfoley:
That sounds like a pretty neat idea. At the very least, we can commit this PR.
The instructions for setting up a Docker Hub repo are at https://docs.docker.com/get-started/part3/

 @quark17, what do you think?

PR #148, opened 2020-04-13
Fix mkFixedPointSquareRooter and add safe-guard to mkSquareRooter

thotypous:
This patch:

* Adds provisos as a safe-guard to `mkSquareRooter`, forcing the designer to choose an even number of bits for the integer type (`UInt#(m)`). The implementation works in `m/2` iterations and doesn't seem to work for odd values of `m`.

* Fixes the logic which computes the amount of shift in `mkFixedPointSquareRooter`, which currently gives results far off from what would be expected (*e.g.* sqrt(0.5)==0.0 and sqrt(0.25)==0.0).

* Fixes the test bench, which currently displays the runtime computed values as "expected", and compile-time computed values as the operation result. Also, adds two new test values for `mkFixedPointSquareRooter`.

I don't know if my solution is optimal. I have written it some years ago and [submitted](http://forum.bluespec.com/viewtopic.php?t=638) it to the forums when Bluespec was proprietary. Suggestions are welcome!

### DIFF
--- a/src/Libraries/Base3-Math/SquareRoot.bsv
+++ b/src/Libraries/Base3-Math/SquareRoot.bsv
@@ -17,9 +17,9 @@ export mkNonPipelinedSquareRooter;
 
 module mkSquareRooter#(Integer n)(Server#(UInt#(m),Tuple2#(UInt#(m),Bool)))
    provisos(
-      // per request of bsc
-      Add#(a__, 2, m),
-      Log#(TAdd#(1, m), TLog#(TAdd#(m, 1)))
+      Div#(m, 2, m2),
+      // m must be even for this implementation to work
+      Mul#(m2, 2, m)
       );
 
    FIFO#(UInt#(m)) fRequest <- mkLFIFO;
@@ -48,7 +48,7 @@ module mkSquareRooter#(Integer n)(Server#(UInt#(m),Tuple2#(UInt#(m),Bool)))
    FIFO#(Tuple4#(Maybe#(Bit#(m)),Bit#(m),Bit#(m),Bit#(m))) fThis = fFirst;
    FIFO#(Tuple4#(Maybe#(Bit#(m)),Bit#(m),Bit#(m),Bit#(m))) fNext;
 
-   for (Integer i = 0; i < (valueOf(m)/2)/n+1; i = i + 1) begin
+   for (Integer i = 0; i < valueOf(m2)/n+1; i = i + 1) begin
       fNext <- mkLFIFO;
       rule work;
 	 //match {.res, .s, .r, .b} <- toGet(fThis).get;
@@ -59,7 +59,7 @@ module mkSquareRooter#(Integer n)(Server#(UInt#(m),Tuple2#(UInt#(m),Bool)))
 	 Bit#(m) b = tpl_4(x);
 
 	 for (Integer j = 0; j < n; j = j + 1) begin
-	    if ((i + j) <= (valueOf(m)/2)) begin
+	    if ((i + j) <= valueOf(m2)) begin
 	       if (res matches tagged Invalid) begin
 		  if (b == 0) begin
 		     res = tagged Valid r;
@@ -100,11 +100,9 @@ endmodule
 module mkFixedPointSquareRooter#(Integer n)(Server#(FixedPoint#(isize,fsize),Tuple2#(FixedPoint#(isize,fsize),Bool)))
    provisos(
       Add#(isize,fsize,m),
-      // per request of bsc
-      Add#(a__, 2, m),
-      Add#(b__, TLog#(TAdd#(1, m)), TAdd#(TLog#(m), 1)),
-      Log#(TAdd#(1, m), TLog#(TAdd#(m, 1))),
-      Add#(c__, TLog#(TAdd#(m, 1)), TAdd#(TLog#(m), 1))
+      // m must be even for this implementation to work
+      Mul#(TDiv#(m, 2), 2, m),
+      Add#(a__, 1, TLog#(TAdd#(1, m)))
       );
 
    FIFO#(FixedPoint#(isize,fsize)) fRequest <- mkLFIFO;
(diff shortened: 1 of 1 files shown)


src/Libraries/Base3-Math/SquareRoot.bsv
```diff
@@ -17,6 +17,9 @@ export mkNonPipelinedSquareRooter;
 
 module mkSquareRooter#(Integer n)(Server#(UInt#(m),Tuple2#(UInt#(m),Bool)))
    provisos(
+      Div#(m, 2, m2),
+      // m must be even for this implementation to work
+      Mul#(m2, 2, m),
       // per request of bsc
       Add#(a__, 2, m),
       Log#(TAdd#(1, m), TLog#(TAdd#(m, 1)))
```
quark17, line 25: These additional provisos are no longer needed

src/Libraries/Base3-Math/SquareRoot.bsv
```diff
@@ -101,6 +104,7 @@ module mkFixedPointSquareRooter#(Integer n)(Server#(FixedPoint#(isize,fsize),Tup
    provisos(
       Add#(isize,fsize,m),
       // per request of bsc
+      Mul#(TDiv#(m, 2), 2, m),
```
quark17, line 107: I would put the comment `m must be even for this implementation to work` on this proviso.  The proviso is required because `mkSquareRooter` requires an even size, and it's being used here at size `m`.

src/Libraries/Base3-Math/SquareRoot.bsv
```diff
@@ -101,6 +104,7 @@ module mkFixedPointSquareRooter#(Integer n)(Server#(FixedPoint#(isize,fsize),Tup
...
       Add#(a__, 2, m),
       Add#(b__, TLog#(TAdd#(1, m)), TAdd#(TLog#(m), 1)),
       Log#(TAdd#(1, m), TLog#(TAdd#(m, 1))),
```
quark17, line 110: The other provisos can also be deleted, except for the following:
`Add#(a__, TLog#(TAdd#(1, m)), TAdd#(TLog#(m), 1))`

FYI, this is needed because `countZerosMSB` returns a value of type `UInt#(TLog#(TAdd#(m,1))` but we're extending that value to compute the shift amount, which is stored in a FIFO of type `Int#(TAdd#(TLog#(m),1))`.  When the code uses `extend(unpack(pack(...)))` to convert `zeros` into the appropriate type for `shift`, that's where the proviso is introduced.

src/Libraries/Base3-Math/SquareRoot.bsv
```diff
@@ -120,17 +124,24 @@ module mkFixedPointSquareRooter#(Integer n)(Server#(FixedPoint#(isize,fsize),Tup
...
-      shift = (fromInteger(valueOf(isize)) - cExtend(zeros)) >> 1;
-      shift = shift + 1;
-      if ((shift & 1) == 0) begin
-	 value = value >> 1;
-      end
+      // A conversion to UInt behaves like the value was multiplied by 2**fsize.
+      // Then, the value is shifted by zeros, which multiplies it by 2**zeros.
+      // Therefore, after computing the square root of the resulting value,
+      // it must be divided by sqrt(2**(fsize + zeros)) = 2**((fsize + zeros) / 2)
+      shift = extend(unpack(pack( (zeros + fromInteger(valueOf(fsize))) >> 1 )));
```
quark17, line 140: Your expressions does the arithmetic in the bit width of `zeros` and then extends to the bit width of `shift` afterward.  I suspect you might need to do the extend first, so that the arithmetic is done in the possibly larger bit width:
`shift = (extend(unpack(pack(zeros))) + fromInteger(valueOf(fsize))) >> 1;`

thotypous: Thank you for the review!

Your suggestion unfortunately doesn't work because `zeros` is unsigned and `shift` is signed. Please note '(zeros+fsize)/2` will always be positive, but it may become negative afterwards when we subtract by `fsize`.

However, I have rewritten the code so it hopefully becomes more readable and generates less circuit:

```
      Int#(TAdd#(TLog#(m),2)) shift_dividend =
         extend(unpack(pack(zeros))) - fromInteger(valueOf(fsize));
      shift = truncate(shift_dividend >> 1);
```

I will force-push the new commit, please let me know if you have more suggestions.

quark17: There are two ways that `zero` is being transformed: one is the conversion from `UInt` to `Int` and two is the extend.  That is, `zero` has type `UInt#(x)` and `shift` has type `Int#(y)`, where `y` is possible 1 bit wider than `x`.  I did not change the type of `shift`, I was just changing when (in the course of the arithmetic) the adding of the bit and the casting to a signed value occur.  Your version does the adding and shifting on a value of type `UInt#(x)` and then added an additional bit and a sign.  My version did both conversions first.  I admit that adding the sign first may not be good; fortunately that's separable from the extending an extra bit, and can be left to the end:
```
shift = unpack(pack( (extend(zeros) + fromInteger(valueOf(size))) >> 1 );
```
My concern was the addition might overflow the `x` bits, in your version.  By doing the extend first, we make sure that we don't overflow and lose data.

quark17:
I showed your changes to someone who is more familiar with the module.  He welcomes this fix but thinks it is written in an unnecessarily complicated way.  He proposes the following patch (to the original code):
```
@@ -118,19 +121,23 @@ module mkFixedPointSquareRooter#(Integer n)(Server#(FixedPoint#(isize,fsize),Tup
       UInt#(m) value = unpack({op.i,op.f});
 
       let zeros = countZerosMSB(pack(value));
-      Int#(TAdd#(TLog#(m),1)) shift;
 
-      // align input
-      value = value << (zeros - 1);
-
-      // compute shift for output
-      shift = (fromInteger(valueOf(isize)) - cExtend(zeros)) >> 1;
-      shift = shift + 1;
-      if ((shift & 1) == 0) begin
-        value = value >> 1;
-      end
+      // The decimal place in 'value' is at 'fsize' bit place.
+      // After alignment, it is 'zeros + fsize' bit place.
+      // After square root, it is '(zeros + fsize) / 2' bit place.
+      //
+      // The shift restores the decimal to 'fsize' bit place, thus a
+      // right (left, if negative) shift by '(zeros + fsize) / 2 -
+      // fsize' or '(zeros - fsize) / 2'.
+      //
+      // Since it is not possible to shift by half a bit, adjust
+      // zeros, if necessary, so that shift is an integer.
+      Int#(TAdd#(TLog#(m),2)) shift2 = cExtend(zeros) - fromInteger(valueOf(fsize));
+      Int#(TAdd#(TLog#(m),1)) shift = cExtendLSB(shift2);
+      UInt#(1) lsb = cExtend(shift2);
 
-      shift = fromInteger(valueOf(isize)) - shift;
+      // align input
+      value = value << (zeros - extend(lsb));
 
       sqrt.request.put(value);
       fShift.enq(shift);
@@ -141,7 +148,10 @@ module mkFixedPointSquareRooter#(Integer n)(Server#(FixedPoint#(isize,fsize),Tup
       let shift <- toGet(fShift).get;
 
       // shift result as necessary
-      result = result >> shift;
+      if (shift >= 0)
+         result = result >> shift;
+      else  // invert direction if shift is negative
+         result = result << (-shift);
 
       FixedPoint#(isize,fsize) fx;
       fx.i = cExtendLSB(result);
```

In place of this proviso that your version needs:
```
Log#(TAdd#(1, m), TLog#(TAdd#(m, 1)))
```
this version only needs this proviso:
```
Add#(a__, 1, TLog#(TAdd#(1, m)))
```
which I think is just a roundabout way to require that `m` is non-zero.

What do you think of this version?  I have not had a chance to digest these new changes, yet, to understand the difference.

thotypous:
> What do you think of this version? I have not had a chance to digest these new changes, yet, to understand the difference.

Looks like it implements the same logic as mine but in a much less convoluted way. Please say thanks to this person.

These `cExtend` / `cExtendLSB` functions are very handful. I should have paid more attention to the BUtils package.

I have force-pushed a new commit integrating the suggested changes to my other fixes.

quark17:
Thanks again for this.  I hacked together some tests for these modules, which I'll add to the testsuite (in `bsc.lib/SquareRoot/`).

The tests include random inputs on a few instances of each module and a check that each result is within expected error.  For `mkSquareRooter`, that check seems simple.  For `mkFixedPointSquareRooter`, I'm less clear on what the expected error range is -- presumably it's a function of the size parameters of the `FixedPoint`.  For `FixedPoint#(48,16)`, for example, the error isn't just in the fractional part but extends into the whole part -- so my naive attempt at a self-check thinks that the result is too far away.  I may still check in my tests, which the current output as the expected output, since it seems reasonable (and a lot better than what the library was computing before).

Issue #149, opened 2020-04-13
itcl/itk linking


quark17:
In PR #146, the issue was raised that itcl/itk libraries may not be available in the same place as tcl/tk (since itcl/itk is intended to be dynamically loaded in tcl).  The issue was deferred from that PR, so I'm opening this ticket to record it.

The tcl interpreter has a search path where index files (`pkgIndex.tcl`) can be found, that declare the available packages, and give arbitrary code for how to load them.  For example, as @bpfoley pointed out in the PR thread:

    $ tclsh
    % package require Itcl
    3.4
    % package ifneeded Itcl 3.4
    load /usr/lib/x86_64-linux-gnu/libitcl3.4.so.1 Itcl

And here is the search path on that same system:

    % puts $auto_path
    /usr/share/tcltk/tcl8.6 /usr/share/tcltk /usr/lib /usr/local/lib/tcltk /usr/local/share/tcltk /usr/lib/tcltk/x86_64-linux-gnu /usr/lib/tcltk /usr/lib/tcltk/tcl8.6

On this system, under `/usr/lib/tcltk/x86_64-linux-gnu/` are `itcl3.4` and `itk3.4` directories that contain `pkgIndex.tcl` files declaring `Itcl` and `Itk`:

    if {![package vsatisfies [package provide Tcl] 8.6]} {return}
    package ifneeded Itcl 3.4 [list load [file join /usr lib x86_64-linux-gnu "libitcl3.4.so.1"] Itcl]

(I would guess that the search of `auto_path` must recurse into subdirectories?)

In addition to this kind of searching, the tcl interpreter can also declare that some packages are statically available.  (This is from the point of view of the interpreter; the library might be dynamically loaded by the executable when run, but that's still considered "static" because the library is already loaded by the time the interpreter executable is running.)  In `bluetcl_Main.hsc` (and `bluewish_Main.hsc`), we declare that itcl (or itk) is available:

    // Itcl startup
    if (Itcl_Init(interp) != TCL_OK) {
      fprintf(stderr,"Unable to start itcl -- %s\n", Tcl_GetStringResult(interp));
      exit (-1);
    }
    Tcl_StaticPackage(interp, "Itcl", Itcl_Init, Itcl_SafeInit);

**NOTE:** At some point (maybe when we unvendored tcl/tk and itcl/itk?) we seem to have also changed the linking of those libraries to be dynamic instead of static.  So we are now requiring that itcl/itk be available in the user's environment, not just the build environment!

[...]

PR #151, opened 2020-04-14
Remove compile-time dependency on itcl/itk (issue #149)

quark17:
If we're not statically linking in Itcl/Itk, then we might as well not declare it as static, and load it when it's required -- as tclsh/wish do.

With this change, I now see (on macOS 10.15.4):

    $ bluetcl 
    % package require Itcl
    3.4
    % package ifneeded Itcl 3.4
    load /System/Library/Tcl/8.5/itcl3.4/libitcl3.4.dylib Itcl
    % package ifneeded Itk 3.4
    load /System/Library/Tcl/8.5/itk3.4/libitk3.4.dylib Itk

    $ bluewish
    % package ifneeded Itcl 3.4
    load /System/Library/Tcl/8.5/itcl3.4/libitcl3.4.dylib Itcl
    % package ifneeded Itk 3.4
    load /System/Library/Tcl/8.5/itk3.4/libitk3.4.dylib Itk

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -60,17 +60,12 @@ HTCL_HS = ../vendor/htcl
 HTCL_INC_FLAGS = -L$(HTCL_HS)
 HTCL_LIB_FLAGS = -lhtcl
 
-# TCL/TK versions
-# TCL and TK are always the same, but ITCL/ITK may have different minor versions
+# TCL/TK version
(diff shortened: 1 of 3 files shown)


bpfoley:
Seems entirely reasonable to me. We took a very circuitous route to get here, but at least we finally got here! :)

PR #152, opened 2020-04-14
Remove itcl/itk/xvfb from the build requirements

quark17:
This updates the documentation and install scripts to match the changes in commit 1f956961.  It removes `xvfb` from the requirements and demotes `itcl` and `itk` to a mention.  It also separately specifies the `bluetcl` requirements from the `bluewish` requirements (in anticipation that we might support not building `bluewish` per issue #116 ).

Issue #116, opened 2020-03-11
Tk dependencies should be optional


thoughtpolice:
The dependencies for the Bluespec package should be *very* minimal at runtime, but because there's a hard dependency on Tk (from what I can tell), you need to pull in other things like font-config, X11, etc. This percolates into a lot of unneeded things. I think all we should really need is `libc.so`, `libstdc++.so`, `libgmp.so`, and perhaps one or two others...

My use case is that I have a [tool](https://github.com/thoughtpolice/nix-makeself) that can create distro-agnostic binary tarballs, similar to the way Bluespec was packaged when it was proprietary (but with a much more fancy usage mechanism so it works anywhere.) Right now adding Bluespec would be a pretty big overhead -- other tools like NextPNR also have this problem when packaging them (they use Qt5, not Tk, however.)

quark17:
The only tool that really needs `libX11` etc is `bluewish`, and I think it's entirely reasonable to have a compile option that avoids building that and therefore avoids the need for X11 etc.  I believe, though, that currently `bluetcl` unnecessarily has the same requirements as `bluewish`, probably because they share the same compile flags (perhaps in an effort to avoid recompilation of the Haskell modules when the the flags change).

bpfoley:
Yeah, this makes sense to me. If you want to use bluesim, tcl is currently required, but Tk should be a (build time) option.

I suspect the best way to do this would be to split the entire bluespec codebase up into several Haskell modules that can be compiled independently, and only have tk deps for bluewish, and tcl deps for bluetcl.

thotypous:
As a related question, are there plans for releasing the workstation (BDW) as open source? I was curious as to whether bluewish is used outside of BDW. Do you have any customers who wrote in-house tools using Tk?

quark17:
I've created a version of the Makefile that removes the Tcl/Tk/X11 dependencies from BSC and the TK/X11 dependencies from Bluetcl.  It's attached to PR #144 and just waiting for @bpfoley to review.

I don't know if anyone is using Bluewish aside from the Workstation GUI.  But hopefully that's moot with the above Makefile fix.

We do plan to release the GUI as open source, but I'm not sure when.  Issue #69 is open for that.

quark17:
[...]

PR #153, opened 2020-04-14
Add checks for binaries, GHC packages & submodules

bpfoley:
Helps new developers get their built environment set up without
lots of inscrutable errors.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,7 +1,46 @@
 PWD := $(shell pwd)
 TOP := $(PWD)/..
 
+include $(TOP)/platform.mk
+
 PREFIX?=$(TOP)/inst
+
+ifndef NO_DEPS_CHECKS
+CC_TOOLS=$(CC) $(CXX) $(LD)
+BSC_TOOLS=ghc ghc-pkg
+BLUETCL_TOOLS=tclsh
+BLUEWISH_TOOLS=wish
+YICES_TOOLS=gperf autoconf
+STP_TOOLS=flex bison
+
+TOOLS=$(CC_TOOLS) \
+      $(BSC_TOOLS) \
+      $(BLUETCL_TOOLS) \
+      $(BLUEWISH_TOOLS) \
+      $(YICES_TOOLS) \
+      $(STP_TOOLS) \
+
+GHC_PKGS=regex-compat syb old-time split
+
+SUBMODS=vendor/yices/v2.6/yices2
+
+# Check all the necessary build tools are present
+pathsearch=$(firstword $(wildcard $(addsuffix /$(1),$(subst :, ,$(PATH)))))
+checktool=$(if $(call pathsearch,$(1)),,$(error Tool $(1) could not be found))
+$(foreach tool,$(TOOLS),$(call checktool,$(tool)))
+
+# Check Haskell package dependencies
+checkghcpkg=$(if $(shell ghc-pkg list --simple-output $(1)),,$(error GHC package $(1) not installed))
+$(foreach pkg,$(GHC_PKGS),$(call checkghcpkg,$(pkg)))
+
+# Check all submodules are initialized
+define SUBMOD_MSG
+Submodule $(1) missing. Initialize with
+    git submodule update --init --recursive
+endef
+checksubmod=$(if $(wildcard $(1)/.git),,$(error $(SUBMOD_MSG)))
+$(foreach submod,$(SUBMODS),$(call checksubmod,$(submod)))
+endif # NO_DEPS_CHECKS
 
 .PHONY: all
 all: install


src/Makefile
```diff
@@ -1,8 +1,48 @@
...
+
 PREFIX?=$(TOP)/inst
 
+# Check all the necessary build tools are present
+CC_TOOLS=$(CC) $(CXX) $(LD)
+BSC_TOOLS=ghc ghc-pkg
+BLUESIM_TOOLS=tclsh
+BLUEWISH_TOOLS=wish
+ifneq ($(OSTYPE), Darwin)
+BLUEWISH_TOOLS+= xvfb-run
```
quark17, line 14: xvfb won't needed after PR #152 is merged

src/Makefile
```diff
@@ -1,8 +1,48 @@
...
 TOP := $(PWD)/..
 
+include $(TOP)/platform.mk
+
 PREFIX?=$(TOP)/inst
 
+# Check all the necessary build tools are present
+CC_TOOLS=$(CC) $(CXX) $(LD)
+BSC_TOOLS=ghc ghc-pkg
+BLUESIM_TOOLS=tclsh
```
quark17, line 11: Did you mean to name this BLUESIM and not BLUETCL?  Both are correct.

src/Makefile
```diff
@@ -1,8 +1,48 @@
...
+      $(STP_TOOLS) \
+      $(CHECK_TOOLS) \
+
+pathsearch=$(firstword $(wildcard $(addsuffix /$(1),$(subst :, ,$(PATH)))))
+checktool=$(if $(call pathsearch,$(1)),,$(error Tool $(1) could not be found))
+$(foreach tool,$(TOOLS),$(call checktool,$(tool)))
+
+# Check Haskell package dependencies
+GHC_PKGS=regex-compat syb old-time split
+checkghcpkg=$(if $(shell ghc-pkg list --simple-output $(1)),,$(error GHC package $(1) not installed))
```
quark17, line 34: What provides `ghc-pkg`? Is it specific to `cabal`?  I'm just wondering if all installations of GHC will have it.

src/Makefile
bpfoley: Removed.

src/Makefile
bpfoley: Haskell itself. When you install GHC, there's a bunch of 'default' packages (eg base, containers etc) installed, and ghc-pkg is a tool that lets you manipulate the package registry.

src/Makefile
bpfoley: I can go with either. I'll rename it to BLUETCL.

bpfoley:
> Technically, `iverilog` isn't required to build a release, so we may not want to require it here. Would it defeat the purpose to put the check in the smoke test directory?

In that case, maybe we should just drop the iverilog check. I was hoping that all the tests would happen almost immediately when you run make, so you don't end up fixing one dependency, then waiting, then fixing up another. But if remove the iverilog check here, the smoke test failure itself will produce an easily understood error message -- iverilog not found.

> Also, if we support building without tk/X11 (per issue #116 ), then there'll need to be a way to not include those in the requirements.

Indeed. We can wrap BLUEWISH_TOOLS in NO_X once we do it in src/comp/Makefile.
 
> Should we add some flag for disabling the check, so that people don't have to edit the Makefile when they know what they're doing? Would it make sense to put these checks into a separate script, that prints all the warnings and returns an error code, which the Makefile can use to choose to abort or not? I have no objection to how it is now, I'm just unsure if this will break things for someone.

Sure. We can do this without requiring a separate script. I've added a `NO_CHECKS` variable. Let me know what you think.

src/Makefile
```diff
@@ -1,8 +1,47 @@
 PWD := $(shell pwd)
 TOP := $(PWD)/..
 
+include $(TOP)/platform.mk
+
 PREFIX?=$(TOP)/inst
 
+ifndef NO_CHECKS
```
quark17, line 8: Can we use a more precise name, like `NO_TOOL_CHECKS` or something?  And any reason not to put the ifdef around the the checks, too, not just the assignments?

bpfoley: How about `NO_DEPS_CHECKS`? We're not just checking for tools, but other dependencies too.

I can move the ifdef to cover the checks too, but technically it's a bit misleading, because the `$(foreach ...)` expressions (with those nasty terminating `$(error ...)`s get expanded  even if they're inside an `ifdef`. What actually matters is  the `TOOLS` etc variables not being assigned to so that the `$(foreach ...)` expressions expand to nothing.

quark17: In my experimentation it's the opposite: `TOOLS` can be defined outside the `ifndef`, and whether we see an error can be enabled/disabled when only the checking code is inside the `ifndef` block.  (I'll check in this version that everything inside the `ifndef`.)

PR #155, opened 2020-04-15
Update bluesim runtime lib to build using -std=c++11

bpfoley:
While here
* Turn on more warnings and make compile warnings into errors.
* Remove obsolete portability code for strdup, strndup, ftello.
  All of these are present even in CentOS 6, and the strndup code
  could read arbitrary memory.
* Remove some unused dead code detected by Clang


bpfoley:
(NB, still passes bsc.bluesim/to_systemc tests)

PR #156, opened 2020-04-15
Don't specify ghc -rtsopts for programs with custom main

bpfoley:
In this case, the argument does nothing, and prints a compiler warning.

The htcl custom main already does the equivalent operation by
calling hs_init_ghc(...) from htcl_initHaskellRTS.

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -177,6 +177,8 @@ GHCINCLUDES =  \
 	-i$(YICES_HS) \
 	-i$(HTCL_HS)
 GHCTMP = '-tmpdir $(TMPDIR)'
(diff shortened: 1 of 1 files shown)


src/comp/Makefile
```diff
@@ -177,6 +177,8 @@ GHCINCLUDES =  \
 	-i$(YICES_HS) \
 	-i$(HTCL_HS)
 GHCTMP = '-tmpdir $(TMPDIR)'
+# Default RTS flags for programs build with a Haskell main
```
quark17, line 180: build -> built?

bpfoley: Spelling hard. Thanks!

PR #157, opened 2020-04-15
Slim down Docker image

thotypous:
This patch disables building `bluewish`, which will hardly be used from inside a container, and removes Tk from runtime dependencies. As a result, the Docker image size is slimmed down from 663MB to 440MB.

The runtime dependencies we keep (build-essential iverilog tcl) are enough for carrying out Verilog and Bluesim simulations, which besides compilation are the main use case for a Docker container in a CI system.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,12 @@ RUN DEBIAN_FRONTEND=noninteractive \
     /build/install_dependencies_ubuntu.sh
 ADD . /build/
 RUN make -C /build -j2 GHCJOBS=2 GHCRTSFLAGS='+RTS -M5G -A128m -RTS'
-RUN make -C /build check
 
 FROM ubuntu:18.04
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


thotypous:
Please hold until #116 is resolved. Maybe we can remove `TCLEXES=bluetcl BLUEWISH_TOOLS=` from make arguments.

bpfoley:
> Please hold until #116 is resolved. Maybe we can remove `TCLEXES=bluetcl BLUEWISH_TOOLS=` from make arguments.

I like that solution though. It's straightforward, and it works... :)

thotypous:
> I like that solution though. It's straightforward, and it works... :)

But if @quark17 removes `bluewish`, we won't need that ;) Then I would force push a patch just adjusting the runtime dependencies.

quark17:
PR #160 has been accepted, so feel free to force push a new commit.

thotypous:
Done, thank you @quark17 !

PR #160, opened 2020-04-18
Remove bluewish, and all dependencies (including tk, x11, fontconfig)

### LINKED ISSUE #116: Tk dependencies should be optional

bpfoley:
Update the docs to reflect this, and while here, update a slightly
bitrotted vagrant provisioning script.

NB this resolves #116


README.md
```diff
@@ -110,19 +110,6 @@ you will need the `tcl` library:
...
-For building and using the Bluespec Tcl/Tk windowing shell (`bluewish`),
-you will also need the `tk`, `fontconfig`, `X11`, and `Xft` libraries:
-
-    $ apt-get install \
-        tk-dev \
-        libfontconfig1-dev \
-        libx11-dev \
-        libxft-dev
-
-Some applications using the Bluespec Tcl/Tk shells will also require
```
quark17, line 122: I had left in the mention of itcl/itk because I had a sense that some of the commands defined by `bluetcl` were returning data in a form that was consumed directly by itcl/itk widgets.  (For example, commands like `browsepackage` and `browsemodule` return a hierarchical view of packages/modules for displaying with the iwidgets `hierarchy` class.)  If that was the case, it would be worth mentioning that (while not required to build the tools) some of the commands in the tool might only be useful if itcl/itk are installed.  However, in looking now, I think the output of those commands is generic enough that we don't need to mention itcl/itk.  (And looking at the GUI tcl code, it seems that the bluetcl commands don't plug directly into the widgets, but need a fair amount of wrapping; which is further evidence that the bluetcl commands are not specific to iwidgets.)

So I'm OK with removing this text.

src/comp/Makefile
```diff
@@ -60,29 +60,26 @@ HTCL_HS = ../vendor/htcl
...
 $(error TCL version check failed)
 endif
 
-# TCL/TK include flags
-TCLTK_INC_FLAGS = $(shell pkg-config --silence-errors --cflags-only-I tcl tk || echo -I/usr/include/tcl)
+# TCL include flags
+TCL_INC_FLAGS = $(shell pkg-config --silence-errors --cflags-only-I tcl || echo -I/usr/include/tcl)
 ifeq ($(OSTYPE), Darwin)
-TCLTK_INC_FLAGS += -I$(shell xcrun --show-sdk-path)/System/Library/Frameworks/Tk.framework/Headers
+TCL_INC_FLAGS += -I$(shell xcrun --show-sdk-path)/System/Library/Frameworks/Tcl.framework/Headers
```
quark17, line 78: I was always confused why we had this path for Tk and didn't have an equivalent one for Tcl.  Now that we're removing Tk, it doesn't seem obvious that we should change Tk to Tcl here versus just remove this line entirely.  Does it not work if you remove this line?  Was the Tk path somehow also including the Tcl code as well?

bpfoley: I think the reason we specifically needed the Tk path was to guarantee we got the system Tk and avoid picking up a `tk.h` that referred to X11. Almost the first thing tk.h does is `#include <tcl.h>`, and as far as I can see it's getting it from `$(shell xcrun --show-sdk-path)/usr/include/tcl.h`

But it looks like now that we've got rid of the Tk dependency this isn't an issue anymore, and it looks like it compiles just fine without the extra path, so I'll drop it.

PR #161, opened 2020-04-18
Flag-controlled optimizations for compiler-derived Bits instances

nanavati:


quark17:
Thanks for exploring this!  It seems good, but I still need to dig deeper and check it on the test suite.  I also want to make sure I understand a few things.  (It's been a couple weeks since I looked over the code, so apologies if my comments here could be answered by re-reading the commits.)

If I understand correctly: The derived Bits instance for enums did not have an explicit default clause, which implicitly creates a `UNoMatch` don't-care; and the `improveIf` code would exempt `UNoMatch` from optimization, which is why pack/unpack might not reduce to identity.

One of the changes explored here is to add an explicit default clause to the Bits instance.  There are also changes to `improveIf`.  Does the explicit default clause, by itself, fix the pack/unpack issue?  Do the `improveIf` changes, by themselves, fix the pack/unpack issue?

One question I had while reading the code was whether why we're treating `UNoMatch` specially.  That is, why should an explicit don't-care make a difference?  I assume it's because, for case-statements inside an Action block (for example), we want the default to be `noAction` -- so, for example, a case-statement of different display statements doesn't do one of the displays in the fall-through case.  Is that it?  Anyway, based on that, I started to wonder if the check for `UNoMatch` was being overly conservative when we know the type isn't `Action` (or the like).  It looks like your changes are more targeted than that: looking for specific constructs (like `PrimChr`), although you do allow `ICInt` reduction when the type is `Bit`.

(But I'm not even sure that distinguishing UNoMatch based on type is right.  I kind of wonder whether the check for it ought to be removed from `improveIf` and whatever purpose it was serving there could be implemented in a more precise way.)

quark17:
You added a (flag-controlled) explicit default clause to the derived `pack` for non-enum data types (of more than one constructor).  Should we not do the same for enum data types?  (And for non-enum types of just one constructor?)

src/comp/FlagsDecode.hs
```diff
@@ -1331,6 +1332,10 @@ externalFlags = [
          (Toggle (\f x -> f {optUndetExpand=x}) (showIfTrue optUndetExpand),
           "internal optimization of undetermined values during static elaboration", Hidden)),
 
+        ("derive-pack-last-clause",
```
quark17, line 1335: This list of flags is mostly in alphabetical order, although some flags aren't.  Should this one be placed alphabetically?

quark17:
This causes problems for some tests in the test suite.  Most notably, `bsc.lib/FloatingPoint/FloatTest.bsv` compiled for hours overnight, until I killed it.

In `bsc.bugs/b1490/`, there are tests for resource use, and `Bug1490MyUnion.bsv` now exceeds the heap.

In `bsc.evaluator/undefined/sysListReg3D.v`, the order of subexpressions has changed.  (Can we figure out why and decide if something needs to be done?)

In `bsc.bugs/b1894/Bug1894.bsv`, there's a change in the generated Verilog -- presumably an improvement, where an if-expression with AAAA for the else-clause is reduced to just its then-clause.  This also happens in `bsc.names/signal_names/Method.bsv`.

For `bsc.scheduler/relax-schedule/IBuffer2.bsv`, there is a difference in signal names, which could be due to improved Verilog (but it'd be good to confirm).

And there appears to be an improvement in `bsc.verilog/undet/MaybeMux.bsv`, where the generated Verilog had unnecessary occurrences of `[31:0]` on 32-bit value (presumably because concats were combined late?) and now it doesn't.

bpfoley:
I've done some further testing of this patchset, and pulled several safe patches out into their own PRs. These are #250, #251, #252, #253.

I noticed that if you apply 0b01275 (Restore if / undet  optimizations) to a clean repository, it causes a number of testsuite test failures. See attached tarball, and the summary below.

```
  ./bsc.bugs/b1753/bsc.sum: Running ./b1753.exp ...
Error FAIL: expected `1' copies of "assign lastdata$D_IN = { ending[5], ending[4:0] + 5'd1 } ;" in `sysShadowInExpr.v', found `0'
  ./bsc.bugs/b1894/bsc.sum: Running ./b1894.exp ...
Error FAIL: did not match {2047u \& \(\(\(\(\(tUInt32\)\(\(tUInt8\)0u\)\) << 3u\) \| \(\(\(tUInt32\)\(DEF_cond__h[0-9]+\)\) << 2u\)\) \| \(tUInt32\)\(DEF_cond__h[0-9]+ \? DEF_AVMeth_s_m : \(tUInt8\)2u\)\);} in `mkTop.cxx'
  ./bsc.lib/FloatingPoint/bsc.sum: Running ./FloatTest.exp ...
Error FAIL: `sysFloatTest.v.out' differs from `sysFloatTest.out.expected'
Error FAIL: Bluesim simulation `sysFloatTest' should execute: child process exited abnormally
Error FAIL: `sysFloatTest.c.out' differs from `sysFloatTest.out.expected'
Error FAIL: Bluesim simulation `sysFloatTest' should execute: child process exited abnormally
  ./bsc.names/signal_names/bsc.sum: Running ./signal_names.exp ...
Error FAIL: `Method.bsv.bsc-vcomp-out.filtered' differs from `Method.bsv.bsc-vcomp-out.expected.filtered'
  ./bsc.scheduler/relax-schedule/bsc.sum: Running ./relax-schedule.exp ...
Error FAIL: `IBuffer2.bsv.bsc-sched-out.filtered' differs from `IBuffer2.bsv.bsc-sched-out.expected.filtered'
  ./bsc.verilog/opt/bsc.sum: Running ./opt.exp ...
Error FAIL: expected `1' copies of "assign RDY_consume = 1'b1 ;" in `mkAOptTwoMethodsUsingSamePort.v', found `0'
  ./bsc.verilog/undet/bsc.sum: Running ./undet.exp ...
Error FAIL: did not match {b \? data\[31\:0\] \: incoming\[31\:0\]} in `sysMaybeMux.v'
```

In particular it causes several tests in /bsc.lib/FloatingPoint/bsc.sum to run forever (or at least for 10+h in one case) after only generating a few lines out output.

[logs.tar.gz](https://github.com/B-Lang-org/bsc/files/5076219/logs.tar.gz)

quark17:
This PR is attempting to solve an issue -- are there source files that demonstrate what is trying to be fixed?

I will need some time to look over each of the individual PRs.  Some do look straightforward.  Some I'm less sure about.  For example, PR #253 might not break the testsuite, but I don't have a sense of the suite's completeness when it comes to testing quality of the generated Verilog.  Yes, that optimization is useful when applied to the if-else tree generated by matching over all the arms of a data type ... but in other situations, could that reduction rule actually make the generated code worse or even prevent other optimizations from applying?  These optimization rules look only at the arms of one if-else expression, and the tree (which ultimately represents a case-expression) is optimized pair by pair.  But if we happen to have an if-else tree where only one pair matches this form, is it still worth applying in just that one pair?  I'd feel safer if our optimizations could look for a larger structure, or if our optimizations could look ahead to see if this rule enables further rules to apply, and only apply this rule if it does.  (Some parts of ITransform/IExpand do this, by returning a bool, so that callers can know if further optimization occurred.)

For example, the reason that we want to apply the rule from PR #253 is to turn a single if-else tree of concatenations:
```
if (x matches Cons0) then
  0 ++ data
else if (x matches Cons1) then
  1 ++ data
else if (x matches Cons2) then
  2 ++ data
else ...
```
into a concat of two if-else trees, because then we notice that the first tree can be reduced:
```
if (x matches Cons0) then 0
else if (x matches Cons1) then 1
else if (x matches Cons2) then 2
else ...
```
or, perhaps more the point of this entire PR, is because we can handle the don't-care of both case-expressions individually:
```
...
else if (x matches ConsN) then n
else _
```
This first case-statement can become simply wires (or can reduce to `PrimChr` -- see Issue #205 ). This is all great when we're optimizing these kinds of expressions.  But what about other expressions?  If we convert the single case-statement of concats into a concat of two case-statements ... but can't then reduce either case-statement ... have we grown the size of the generated code without adding anything? (And that's the best outcome; at worst, have we possibly prevented another optimization?)

Certainly we need rules to optimize these kinds of expressions, when written by the user.  But for the derived Bits instances ... could we perhaps start by just writing the code in the form that we want?  That is, could we write is a concat of two case-statements, to begin with?  Perhaps not even as two case-statements, but a concat of `PrimChr` and a case-statement of the data? (avoiding the work of reconstructing `PrimChr`).  (Ravi's comment on Issue #205 is already suggesting that we could implement the Bits for enum this way.)

Anyway, to return back to the high-level points: Thank you for splitting this into pieces.  Some pieces (such as recording when a type is an enum) make sense to commit on their own.  Some pieces (like individual reduction rules) are less obvious to add on their own merely based on whether they break anything -- they might still be worth committing on their own, but I would need to see example test cases and an argument that this rule always makes forward progress, when considered on its own.  (I'm also wondering whether tweaks of the current system are going to be successful or whether we really need some rearchitecting; for example, so that instead we optimize by looking at the whole if-else tree, or by doing more to only apply reduction rules when they enable other rules.)

I'll take some time to look more specifically at these patches, to give more concrete comments.  As for possible infinite loops: that can happen when reduction rules are at odds.  For example, there could be a rule that says: a concatenation of two case-statements with constants in all the arms can be reduced into a single case-statement of concats of those constants, because concat of constants reduces to constants.  But if we have that rule, and we also have the rule that splits case-of-concat into concat-of-case, and we're not careful, then these optimizations could ping pong.  (That's why single optimizations like PR #253 make me nervous, without any argument for why it's safe to always apply the rule; and why I'd tend to prefer rules that only apply when they know they can make progress beyond their simple scope.)

quark17:
I realize that I might be being too harsh on PR #253 , and it might be worth committing, but I'd just like to see some example test cases and an argued reasoning for it.

I do notice that my original pseudocode was perhaps too simplified, when I showed a single concat of a constructor tag and a data value.  That `dataN` value could be a concatenation of several fields.  And a few constructors could have aligned fields, which might be worth optimizing by splitting the data case-of-concats into concat-of-cases; but if only two (adjacent) constructors have aligned fields, could optimizing that one pair (not the whole tree) prevent cleanly generating a case-statement for the whole tree? or would reducing that one pair still be useful?  It'd help to have concrete test cases that demonstrate these things (constructors with multiple fields, adjacent constructors with aligned fields, non-adjacent constructors with aligned fields) so that we can observe the effects of any individual optimizations.  (And these should use explicit code, not derived Bits instances, so that what they are testing doesn't change as we change the derived instances.)

quark17:
> but I don't have a sense of the suite's completeness when it comes to testing quality of the generated Verilog

There are some tests that diff the generated Verilog against expected, so if you make a change, you might observe a failure if it causes a change in those speciifc tests.  If you want to check over a larger sample, here's a possible idea: Run the testsuite with a fresh BSC and save that aside; then run another testsuite with your changed BSC; find all `.v` files and diff them between the two runs.  You'll probably need to diff in the way that the testsuite does, to ignore changes in the names of signals, so you can just see the structural changes.  Anyway, that might show you some places where the generated Verilog changed, but didn't affect whether the test passed.  Another check you could do is to compare the runtime (of the codegen steps, over the whole suite or individual tests) to see if the changed BSC took more or less time (while not affecting whether the test passed).

bpfoley:
Yup. A couple of the logs I attached to #161 are like that: just changes in
the Verilog that I think don't matter.

Cheers,
B.

quark17:
> Yup. A couple of the logs I attached to #161 are like that: just changes in the Verilog that I think don't matter.

I'm unsure how this responds to my concerns.  Let me try to be more specific:

1. For PRs #252 and #253 , please give:
   1. An argument why this patch makes sense by itself, and that applying it is always forward progress
   2. An example source file showing how the generated Verilog is improved by this patch
   3. Some proof that Verilog for other examples is not made worse by this patch
      1. You've said that the testsuite passed, but that is not enough, since most generated Verilog is not compared against an expected output.  I'd feel somewhat better if you actually diff'd all the generated Verilog files, and determined that none of the differences reduce the quality
      2. For #253, I specifically listed some examples to try: examples of data types with multiple constructors where some constructors have multiple fields where some fields are aligned (in various combinations with other constructors whose fields don't align)

2. For this PR #161 , please attach a source file (or files) which produces Verilog that is "bad" and which this PR is attempting to improve.  Show what the improved Verilog ought to look like.  If possible, also show what the poor Verilog is, and the command-line to generate it, so that I can guarantee that I am reproducing what you see.

nanavati:
This is obsolete. All of the interesting features have been re-implemented with cleaner code.

PR #168, opened 2020-04-29
Fix width mismatch in FIFO verilog to make Verilator happy

mtikekar:


quark17:
Can you provide some context for this?  What is the behavior that you're seeing from Verilator? and what version of Verilator?  (And version of BSC, if different.)

Verilator has been used to simulate BSV designs with FIFOs, so I'm wondering what the issue is here.  For example, the Piccolo processor has Makefiles for building Verilator sims: https://github.com/bluespec/Piccolo/tree/master/builds/RV32ACIMU_Piccolo_verilator

mtikekar:
I'm on verilator 4.034. It prints warnings about width mismatch between `guarded` (32 bits) and `DEQ` (1 bit) and finally exits with an error due to those warnings. In the Piccolo repo, the warnings are being turned off [here](https://github.com/bluespec/Piccolo/blob/master/builds/Resources/Verilator_resources/verilator_config.vlt), but I would like to keep them around as I have some hand-written Verilog for which the linting would be useful.

To be clear, this does not solve the problem completely. In bsc-generated Verilog, uninitialized state elements are initialized with the 101010 pattern as `{((width + 1)/2){2'b10}}` which has a width mismatch for odd values of width. Another instance is the use of `$test$plusargs` (32-bit output is being assigned to 1-b variable).

By the way, the lint disabling command used in Piccolo was deprecated at some point. `lint_off -msg <RULE>` needs to be replaced with `lint_off -rule <RULE>`.

quark17:
Thanks!  (We observed a segfault in simulation when using 4.008, so we've been holding back to 3.916.)

quark17:
Instead of changing the uses of `guarded` to expect it as a 32-bit value, I wonder if the right fix here is to make `guarded` be a 1-bit value (by only assigning 1-bit values to it).

I notice that, in `FIFOF_.bsv` where the Verilog modules are imported, that the `guarded` parameter is `Integer` (and that the various BSV wrappers around the module continue to take an `Integer` argument `g`).  This means that BSC will generate module instantiations which provide 32-bit values.  Originally, BSC may not have supported parameters of other types, but it does now, so maybe we should use it.  (I believe that, in the absence of size information, Verilog will default integer numbers to at least 32-bit, which may be why that was used originally.)

If `guarded` only ever encodes true or false, then it'd be safer to give the type `Bool` anyway.

Doing this would involve changing all the BSV import-BVI modules (and all the wrappers around them) to take a `Bool` argument instead of `Integer`.  And in the Verilog files, we probably should change the default assignment for the parameter to be `1'b1` instead of just `1`.  (Instantiations inside BSC-generated modules should never use the default, though.)

mtikekar:
I agree that `guarded` should be made a 1-bit value instead. I wasn't sure why it was 32-bit in the first place and changing it is a more extensive change, so I proposed a minimal change in this PR.

quark17:
I will submit a PR that changes `guarded` to `Bool`.  I've made the changes, and it doesn't appear to break anything.  (And most users won't be exposed to the change, since the exported modules are wrappers around the core primitives, and those wrappers don't expose the Integer argument.)

One thing I notice is that Verilog-2001 allows a parameter to have a declared type (rather than getting the type from the last assigned value).  In the Verilog primitives, we could declare `guarded` as one bit.  I don't imagine that anyone still needs BSC to support only V95, but I'll leave that as a later possibility, since it's not needed to address the current issue once the BSV knows to assign the parameter a 1-bit value in instantiations.

quark17:
I've created PR #264 which seems to avoid the Verilator WIDTH warnings related to the `guarded` parameter:
```
%Warning-WIDTH: ../../src_bsc_lib_RTL/FIFO2.v:167: Logical Operator LOGOR expects 1 bit on the RHS, but RHS's VARREF 'guarded' generates 32 bits.
%Warning-WIDTH: ../../src_bsc_lib_RTL/FIFO20.v:124: Logical Operator LOGOR expects 1 bit on the RHS, but RHS's VARREF 'guarded' generates 32 bits.
%Warning-WIDTH: ../../src_bsc_lib_RTL/SizedFIFO.v:248: Logical Operator LOGOR expects 1 bit on the RHS, but RHS's VARREF 'guarded' generates 32 bits.
```

Can you confirm that this PR does what you want?  If so, I'll merge it and close this PR.

mtikekar:
Hi @quark17, I checked that PR #264 fixes the verilator issue. This PR can be closed in favor of that one.

PR #171, opened 2020-05-06
Use ccache to cache C++ compilation across builds

bpfoley:
ccache will use hashes of the full compiler command line and input
files to determine whether to use previously compiled object files
or not, so this is safe.

Cumulatively before caching stp and yices take about 2.5mins to
compile on Ubuntu.

Store an updated cache after every build, and rely on github's cache
pruning to remove the oldest caches.

Currently <100MB of uncompressed object files per OS.

### DIFF
--- a/.github/workflows/install_dependencies_macos.sh
+++ b/.github/workflows/install_dependencies_macos.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
-brew install autoconf cabal-install gperf icarus-verilog pkg-config && \
+# ccache is not required to buid bsc, but we use it in build.yml to improve
+# the build performance by caching C++ obj files across multiple builds.
+brew install ccache autoconf cabal-install gperf icarus-verilog pkg-config && \
     cabal update && \
     cabal v1-install old-time regex-compat split syb
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,10 +30,28 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: "sudo .github/workflows/install_dependencies_ubuntu.sh"
+      # Restore previous ccache cache of compiled object files. Use a SHA
+      # in the key so that a new cache file is generated after every build,
+      # and have the restore-key use the most recent.
+      - name: CCache cache files
+        uses: actions/cache@v1
+        with:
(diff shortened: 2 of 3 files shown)


.github/workflows/install_dependencies_macos.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-brew install autoconf cabal-install gperf icarus-verilog pkg-config && \
+brew install ccache autoconf cabal-install gperf icarus-verilog pkg-config && \
```
quark17, line 3: Note that this is the first time that we're adding a dependency to the scripts that isn't required to build BSC, it's only required by the CI.  I think that some people might have been using these scripts to grab dependencies for their own systems, and now they'd be getting unnecessary packages.

I have no object to the files under `.github/` being specific to the GitHub flows.  But maybe we should find an area outside `.github/` where install scripts for various OSes can sit; and then the CI can call out to those shared files, to install the build dependencies, while separately having some command here to install the additional packages that it needs.

bpfoley: Yeah, that's true. On the other hand, `.github/workflows` is pretty clearly GitHub specific scripting, and I'm not sure there's any need for us to do the distro maintainers' jobs for them (unless we become those maintainers of course :))

We could also move these scripts (or even just the ccache install) inline into the build.yaml if you like.

quark17: I was just noting the issue, and listing possible options.  I don't feel strongly enough to tell you to do anything different.  I think it's slightly more self-documenting / less error-prone to separate the ccache install, which could be done in the build.yaml file since that's the place that uses ccache.  Or, at the very least, the install script could have two install commands in it, with comments above the one for CI and the one for the build. (I wasn't thinking of just distro maintainers, but also of individual users who want a convenient place to copy a command-line before building.  But that becomes even less important once we get pre-built versions up and into distress.)

bpfoley: Fair enough. How about I add a comment noting the use of ccache? Running brew/apt-get twice adds a little overhead to the setup (on the order of a couple of seconds) as they need to scan the package database each time they run.

PR #172, opened 2020-05-19
Silence false positive overflow error on gcc 10

thotypous:
`0 <= numDigits(v, base) <= 64` for any `v`, `base`. However, since the value is returned as an `unsigned int`, gcc 10 alerts for a possible overflow on `char buf[value_width+1]`, although it is a false positive. By returning an `unsigned short`, the warning (which is actually treated as an error due to the CXXFLAGS) is silenced.

```
dollar_display.cxx: In function ‘const char* print_binary(tFieldDesc&, ArgList*, Target*)’:
dollar_display.cxx:741:22: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
  741 |     buf[value_width] = '\0';
      |     ~~~~~~~~~~~~~~~~~^~~~~~
dollar_display.cxx:740:10: note: at offset [-9223372036854775808, 9223372036854775807] to an object with size at most 4294967295 declared here
  740 |     char buf[value_width+1];
      |          ^~~
```

Please note gcc 10 is already the default compiler on some distributions, such as Arch Linux.

### DIFF
--- a/src/bluesim/dollar_display.cxx
+++ b/src/bluesim/dollar_display.cxx
@@ -91,7 +91,7 @@ static unsigned int maxWidth(unsigned int nBits, bool isSigned,
 
 // Return the number of digits required to represent a given value
 // in the specified base.
-static unsigned int numDigits(const tValue& v, unsigned int base=10)
+static unsigned short numDigits(const tValue& v, unsigned int base=10)
 {
   if (v.bits < 2)
     return 1;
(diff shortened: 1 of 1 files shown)


bpfoley:
I think `unsigned short` isn't really the right type here. What you want to use update `numDigits()` to return`size_t`, just like `strlen` does, and make `pad()` use `size_t` for its widths too. That seems to make gcc 10 happy.

bpfoley:
Hi @thotypous, any chance you could update your patch to use `size_t` instead of `unsigned short`?

bpfoley:
Closed, and replaced by #204

PR #173, opened 2020-06-02
Fix install-utils typo in src/comp/Makefile

nanavati:
Fix a typo in the install-extra target (for installing bsv2bsc and other utilities from src/comp)


quark17:
Thanks!  I'd run into this before and thought that I'd fixed it, but guess not.

PR #175, opened 2020-06-04
[WIP] makedepend.tcl: Generate dependency rule for the Makefile itself

jrtc27:
This ensures that the dependency file is regenerated automatically when
the source files change rather than becoming stale.


jrtc27:
Never mind, this doesn't work due to the file having been included by the point of reading the dependencies in it, and it would still break if a source file got removed.

PR #179, opened 2020-06-16
Fixes to .ghci

krame505:
This fixes the configuration in `.ghci` to allow source files to be loaded in ghci.  

Also renamed variables named `rec` to avoid conflicting with `-XDoRec`, which is apparently enabled by default in ghci.

### DIFF
--- a/src/comp/Balanced.lhs
+++ b/src/comp/Balanced.lhs
@@ -348,12 +348,12 @@ Folding a list in a binary-subdivision scheme.
 > foldm                         :: (a -> a -> a) -> a -> [a] -> a
 > foldm (*) e x
 >   | null x                    =  e
->   | otherwise                 =  fst (rec (length x) x)
->   where rec 1 (a : as)        =  (a, as)
->         rec n as              =  (a1 * a2, as2)
+>   | otherwise                 =  fst (r (length x) x)
+>   where r 1 (a : as)          =  (a, as)
+>         r n as                =  (a1 * a2, as2)
 >           where m             =  n `div` 2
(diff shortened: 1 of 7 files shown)


src/comp/Balanced.lhs
```diff
@@ -348,12 +348,12 @@ Folding a list in a binary-subdivision scheme.
 > foldm                         :: (a -> a -> a) -> a -> [a] -> a
 > foldm (*) e x
 >   | null x                    =  e
->   | otherwise                 =  fst (rec (length x) x)
->   where rec 1 (a : as)        =  (a, as)
->         rec n as              =  (a1 * a2, as2)
+>   | otherwise                 =  fst (r (length x) x)
+>   where r 1 (a : as)          =  (a, as)
```
quark17, line 352: I think a single-letter name gets lost here, particular as a function.  What about `recFn`?

src/comp/IExpand.hs
```diff
@@ -1215,15 +1215,15 @@ methodResetName i ws = do
 -- "run" the module monad, collecting rules, interfaces, and state
 iExpandModule :: Bool -> (HClock, HReset) ->
                  IStateLoc -> HPred -> HExpr -> G PExpr
-iExpandModule rec curClkRstn ns p e = do
+iExpandModule r curClkRstn ns p e = do
```
quark17, line 1218: I'd prefer not to use a single-letter here.  While `isRec` might work, this variable indicates whether we're evaluating inside a use of `PrimModuleFix`, so perhaps `isMFix`?

krame505:
Names updated as requested.

PR #180, opened 2020-06-18
Add CI for Ubuntu 20.04

bpfoley:
The Ubuntu 20.04 environments are preview only at the moment, but appear to Just Work™.


bpfoley:
The macos build randomly timed out after a successful smoketest run, I suspect it was a VM issue outside our control. Besides this PR doesn't touch macOS at all.

PR #183, opened 2020-06-20
Allow constructors to be partially applied

krame505:
Fairly straightforward change, also adds a better error message when a constructor is applied to too many arguments.

### DIFF
--- a/src/comp/Error.hs
+++ b/src/comp/Error.hs
@@ -756,8 +756,9 @@ data ErrMsg
 
         | EFuncMismatchResult  String String String  String String
         | EFuncMismatchNumArgs String String String  Integer Integer (Maybe Integer)
-        | EConMismatchNumArgs  String String         Integer Integer
-        | EPartialConMismatchNumArgs  String String String  Integer Integer Integer
+        -- XXX these should contain the type of the constructor
+        | EConMismatchNumArgs  String{-String-}      Integer Integer
+        | EPartialConMismatchNumArgs  String String{-String-}Integer Integer Integer
         | EFuncMismatchArgN    String String String  Integer String String
         | EFuncMismatchNoArgsExpected String String String
         | EFuncMismatchArgsExpected   String String String  Integer
@@ -2451,19 +2452,21 @@ getErrorText (EFuncMismatchNumArgs e t1 t2 n1 n2 maybe_startnum) =
      s2par "Inferred type:" $$
      nest 2 (text t2))
 
-getErrorText (EConMismatchNumArgs e t n1 n2) =
+-- XXX Errors about consturctor argument mismatches should contain the type of
+-- the constructor, but this is overly hard to actually compute.
+getErrorText (EConMismatchNumArgs e {-t-} n1 n2) =
     (Type 81, empty,
      s2par "Too many arguments in the use of the following constructor:" $$
      nest 2 (text e) $$
      s2par ("The constructor expects " ++ show n1 ++
-            " arguments but was used with " ++ show n2 ++ " arguments.") $$
+            " arguments but was used with " ++ show n2 ++ " arguments.") {-$$
      if n1 > 0
      then
        s2par "Constructor has type:" $$
        nest 2 (text t)
-     else text "")
+     else text ""-})
 
-getErrorText (EPartialConMismatchNumArgs e t1 t2 n1 n2 n3) =
+getErrorText (EPartialConMismatchNumArgs e t1 {-t2-} n1 n2 n3) =
     (Type 81, empty,
      s2par "Wrong number of arguments in the partial application of the following constructor:" $$
      nest 2 (text e) $$
@@ -2475,12 +2478,12 @@ getErrorText (EPartialConMismatchNumArgs e t1 t2 n1 n2 n3) =
             then  "with " ++ show n3 ++ " arguments was expected."
             else "was not expected.") $$
      s2par "The expected type is:" $$
-     nest 2 (text t1) $$
+     nest 2 (text t1){- $$
      if n1 > 0
      then
        s2par "Constructor has type:" $$
        nest 2 (text t2)
-     else text "")
+     else text ""-})
 
 getErrorText (EFuncMismatchArgN e t1 t2 num argt1 argt2) =
     (Type 82, empty,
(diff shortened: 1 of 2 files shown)


quark17:
I still need to take a closer look at your changes, but in the meantime, here are the 3 testsuite changes that are caused by this commit.  Can you submit a PR to bsc-testsuite to resolve these?  Also, can you look at the error messages for the 2nd and 3rd examples, and see if you think the error messages could be improved, as part of this bsc PR?

```
Running ./bsc.syntax/syntax.exp ...
XPASS: `ConstructorDot.bs' compiles (PRMS 1062)

Running ./bsc.typechecker/constructors/constructors.exp ...                     
FAIL: expected `1' copies of Error T0020 in `bsc.typechecker/constructors/UnionOneArgGivenNone.bsv.bsc-out', found `0'                                         
FAIL: `UnionOneArgGivenNone.bsv.bsc-out' differs from `UnionOneArgGivenNone.bsv.bsc-out.expected'

Running ./bsc.typechecker/kind/kind.exp ...
FAIL: expected `1' copies of Error T0007 in `bsc.typechecker/kind/DataConstrOfNoArgsAppliedToMultipleArgs.bsv.bsc-out', found `0'
```

The first example (`ConstructorDot.bs`) was written with partial-application in mind, so it was an "expected failure" that BSC wouldn't compile it.  Now it compiles, so the test command should change to expect the compile to succeed.

The second example (`UnionOneArgGivenNone`) is also good!  As the exp-file says, the check for T0020 is just confirming that the current output is what we expect, but we'd really like to see a different error.  The code is this:
```
typedef union tagged { Bool Foo; } Bar;

Bar x = Foo;
```
The expected behavior of BSC was this:
```
Error: "UnionOneArgGivenNone.bsv", line 3, column 9: (T0020)
  Type error at:
    ()

  Expected type:
    Bool

  Inferred type:
    void
```
and now BSC (with your changes) gives this:
```
Error: "UnionOneArgGivenNone.bsv", line 3, column 9: (T0083)
  The following expression does not expect arguments, but was used with
  arguments:
    \ _tctemp1003 -> (Foo)(_tctemp1003)

  The expected type is:
    UnionOneArgGivenNone::Bar

  But it was used with the following type:
    function b__ f(a__ x1)
```
This is an improvement, but I wonder if this is the correct error message.  If we know that the context is expecting more arguments to the constructor than what have been applied, we should report that.  In your code for `tiExpr`, you only compare the number of given arguments with the total number of arguments.  I wonder if it's worth also comparing against `td`, for those situations when `td` does provide further information?

We may not always be able to detect this situation at the point of typechecking in the constructor (for example, when `td` does not provide further info).  In that case, I still wonder if we can report this specific error message better?  Can we catch the situations where unification fails on an expression of this form (or with more lambdas):
```
\ _tctemp1003 -> (Foo)(_tctemp1003)
```
and (based on the given and expected types) report a better message -- preferably one that doesn't mention the internal name `_tctemp1003`.  This work could arguably be recorded as a future enhancement, if it's not worth your doing now (but I'd prefer to commit it all together).

Finally, the third example (`DataConstrOfNoArgsAppliedToMultipleArgs.bsv`) is also good!  It's again a situation where the suite is testing for the expected behavior, but we'd really like to see different behavior.  The expected output is:
```
Error: "DataConstrOfNoArgsAppliedToMultipleArgs.bsv", line 5, column 12: (T0007)
  Unbound type constructor
  `DataConstrOfNoArgsAppliedToMultipleArgs::Color_$Red'
```
And with your change to BSC, we now get:
```
Error: "DataConstrOfNoArgsAppliedToMultipleArgs.bsv", line 5, column 12: (T0081)
  Too many arguments in the use of the following constructor:
    DataConstrOfNoArgsAppliedToMultipleArgs::Red
  The constructor expects 0 arguments but was used with 2 arguments.
  
  Constructor has type:
    function DataConstrOfNoArgsAppliedToMultipleArgs::Color f(void x1)
```
which is a significant improvement!  My one complaint is that additional lines at the end are unhelpful (and in fact incorrect) when the type is an enum and/or takes zero arguments.  (Showing the constructor's type is only helpful when the constructor has arguments -- because the type is for showing the type of those arguments.  And when the constructor is an enum tag, displaying the type as a function on a void argument is exposing an internal representation.)  It's arguable that these improvements can be filed as future work, but if you feel like addressing it here, that'd be great.  In any case, the exp-file needs to be updated to expect this error output now.

Thanks!

quark17:
In the `UnionOneArgGivenNone.bsv` example, maybe this construction:
```
\ _tctemp1003 -> (Foo)(_tctemp1003)
```
can be avoided by noticing that the constructor only takes 1 argument (`numExpected` is 1) and has been given no arguments (`numGiven` is 0), and so there's no need to wrap it in a lambda, we can just return the constructor itself.

It would also be worth making a similar example of a constructor missing all its arguments, but with a constructor that takes 2 (or more) arguments, to see how the error message looks in that case.  (It might need to be written in BH/Classic, since BSV doesn't support data constructors of that form, I think.)

krame505:
I added a much nicer error for cases involving partial application of constructors.  

Also, it turns out that it is very difficult to properly display the type of a constructor - the cases with 0 or 1 arguments are fine, but when there are 2 or more we must somehow manually lookup the type of field of the generated struct in the symbol table.  I really didn't feel like diving into this right now (although the type of the constructor in question would be a nice addition to error messages about constructor application), so I simply disabled this for now and added comments to that effect.

src/comp/Error.hs
```diff
@@ -2449,6 +2452,39 @@ getErrorText (EFuncMismatchNumArgs e t1 t2 n1 n2 maybe_startnum) =
      s2par "Inferred type:" $$
      nest 2 (text t2))
 
+-- XXX Errors about consturctor argument mismatches should contain the type of
+-- the constructor, but this is overly hard to actually compute.
+getErrorText (EConMismatchNumArgs e {-t-} n1 n2) =
```
quark17, line 2457: The first field in the return value of `getErrorText` is a unique ID.  You've inserted two new messages with the same ID as an existing error (`Type 81`).  Instead, you'll need to insert these messages after `Type 142` and give them the IDs `Type 143` and `Type 144`.  (Related messages can grouped in the constructors of `ErrMsg` but the associated arms of `getErrorText` may not be together, if the messages are added at different times.)

krame505: Ah, that's what I get for copy/pasting without looking too closely :-) Should be fixed.

PR #186, opened 2020-06-27
Make DefaultValue derivable

krame505:
This moves the `DefaultValue` type class from a library into the Prelude, and makes it derivable on struct and data declarations.


src/Libraries/Base1/Prelude.bs
```diff
@@ -4044,3 +4038,32 @@ instance (FShow a) => FShow (List a)
...
+-- ModIfc <- mkMod (modParams) ; -- construct the module
+
+
+class coherent DefaultValue a where
+   defaultValue :: a
+
+-- Any type in the Literal class can have a default value -- simply 0
+-- This picks up Bit#(n), Int#(n), UInt#(n), Real, Integer
+-- as well as FixedPoint, Complex
+instance (Literal t) => DefaultValue t where
```
quark17, line 4068: It's worth noting that the provisos/contexts on an instance are not used when testing whether an instance matches!  So this instance will be a catch-all for any type that doesn't match a more specific instance -- regardless of whether that type has an instance of Literal or not.  When the base type of the instance matches, then the provisos/contexts are added as additional requirements to be met.  (So a type T that falls through but doesn't have a Literal instance will probably trigger an error that there are no instances of Literal for T.)

I know that you didn't introduce this behavior, but I wanted to mention it.  I'll still accept this PR (as the behavior has not changed).

quark17:
I added a testcase to bsc-testsuite (`DeriveDefaultValue.bsv` in `bsc.typechecker/deriving/`)

krame505:
Thanks, I am aware of that behavior. 

FWIW, we anticipate that this deriving code will soon be replaced by the more general generics mechanism that we are planning to build.  This may call for the introduction of a different default instance that conflicts with this one, but we still need to discuss those details.

PR #188, opened 2020-06-29
FixedPoint with 0 integer bits

mtikekar:
There are two fixes pertaining to FixedPoint#(0, f):

1. Use the msb of fractional part as sign bit for fxptSignExtend
2. Fix `1 is not a valid Int#(1) literal` error in fxptWrite

FixedPoint#(0, f) is subtly broken currently. In principle, it could represent numbers as either of two ranges:

1. [0 to 1 - 2^-f] - if the sign bit is always 0
2. [-0.5 to 0.5 - 2^-f] - if the sign bit is the msb of the fractional part

All the Arith and Ord functions first convert FixedPoint#(i, f) to Int#(TAdd#(i, f)) so FixedPoint#(0, f) gets implicitly treated as case 2. But fxptSignExtend and fxptWrite implicitly treat it as case 1. The first commit makes them all consistent as case 2.

fxptWrite also need a fix for an invalid literal error that only shows up for FixedPoint#(0,f)

### DIFF
--- a/src/Libraries/Base3-Math/FixedPoint.bsv
+++ b/src/Libraries/Base3-Math/FixedPoint.bsv
@@ -518,8 +518,9 @@ function FixedPoint#(ri,rf) fxptSignExtend( FixedPoint#(ai,af) a )
    provisos( Add#(xxA,ai,ri),      // ri >= ai
              Add#(fdiff,af,rf)    // rf >= af
             )  ;
-   return FixedPoint {i:signExtend(a.i),
-                      f:{a.f,0} } ;
+   Bit#(TAdd#(ai, 1)) i_a = {a.i, msb(a.f)};
+   Bit#(TAdd#(ri, 1)) i_r = signExtend(i_a) >> 1;
+   return FixedPoint {i: truncate(i_r), f: {a.f, 0}};
 endfunction
 
 
@@ -654,7 +655,8 @@ function Action fxptWrite( Integer fwidth,
             );
    action
       // this can be i bits, but iverilog gets confused!
-      Int#(TAdd#(1,i))  ipart = extend(fxptGetInt(a));
+      FixedPoint#(TAdd#(i,1), f) a_ext = fxptSignExtend(a);
+      Int#(TAdd#(1,i))  ipart = fxptGetInt(a_ext);
       Bit#(f)           fpart = pack( fxptGetFrac( a ) ) ;
 
       // negate the fractional part if the integer part is less than 0
--- a/src/Libraries/Base3-Math/FixedPoint.bsv
+++ b/src/Libraries/Base3-Math/FixedPoint.bsv
@@ -661,7 +661,7 @@ function Action fxptWrite( Integer fwidth,
       if (( ipart < 0 ) && (fpart != 0))
          begin
             fpart = 0 - fpart ;
-            ipart = ipart + 1;
+            ipart = ipart - (-1);
             if ( ipart == 0 )
                $write( "-0." ) ;
             else


darius-bluespec:
I agree that there is an issue with ```FixedPoint#(0, f)```, but this is not the correct solution.  Problems would remain with other function that processes the fractional bits (e.g., ```epsilon```, ```fxptGetFrac```).

A better solution is to disallow zero integer bits.  This is consistent with the documentation in the code and assumptions throughout the package.

quark17:
@mtikekar , are you okay with the alternate proposal, to disallow zero integer bits?  PR #249 has been opened for adding provisos to specify this restriction.

mtikekar:
I need `FixedPoint#(0, f)` in my code and I cannot replace it with `FixedPoint#(1,f)` as I rely on the wrap-around behavior. I'm using it to represent angles with the range [-pi, pi) mapped to [-0.5, 0.5). This way, wrap-around in binary nicely coincides with wrap-around in angles.  

If #249 is merged, I can just maintain a copy of FixedPoint.bsv in my project, but I think it would be better to support `FixedPoint#(0,f)` if possible. Perhaps we can add the provisos only on the functions that don't work? Btw, I don't see how `epsilon` and `fxptGetFrac` are broken. Maybe `fxptGetInt` is broken, but only if `Int#(0)` is not a valid type.

I was even thinking of generalizing `FixedPoint#(i,f)` to negative i and f. For example, if I want to represent  the range [-0.125,0.125), I would use `i = -2`. Or, if I want an epsilon of 4, I would use `f = -2`. These look like perfectly valid generalizations, although they would break the `fxptGetInt` and `fxptGetFrac` functions. Also, numeric types don't support negative numbers (I think), so we would need to rethink how to represent the data-type which would mean a complete redesign of FixedPoint.bsv.

quark17:
@mtikekar , I agree with you that it's possible to imagine a data type representing N binary digits at a fixed location P places from the decimal point (anywhere before, after, or straddling the decimal).  But that is not what the current `FixedPoint` package is.  You'd need to create a new package.

The current package defines a type with `i` bits before the decimal, `f` bits after the decimal, and with a sign bit.  PR #249 clarifies that the sign bit is always included in `i`, which must therefore always be at least 1.

You're asking, instead, to support the idea that, when `i` is zero, the sign bit is part of `f`.  That would at least require adding provisos to say that `i + f` is at least 1.  (Or do you also want to support a meaning for `FixedPoint#(0,0)`?)

For example, the function `epsilon` calls `_fromInternal` on `'b01`.  Hm, perhaps this argues that `i + f` must be at least two?  Or do `FixedPoint#(1,0)` and `FixedPoint#(0,1)` have a meaning, and would `epsilon` as written give the right answer for them?  (@darius-bluespec , does PR #249 also need this proviso, at least on the `epsilon` function? that not only is `i` at least 1, but that `i + f` is at least 2?)

In any case, it does seem to me that the assumption about the sign bit being part of `i` is baked in to the current package.  For one, there's no `fxptGetSign`.  I also notice that the `FShow` instance prints out the raw bits -- so a negative number is shown in two's complement form, rather than as a positive number with a minus sign in front.  This definition will break when `i` is 0 because the output will appear as `0.` followed by two's complement fractional digits.  We could fix this, but...

At this point, I think it's safest to just guard the package with the provisos that the designer was obviously assuming.  Making the package support one additional feature (but not the fully generalized range) seems wrong, unless it can be shown that it's really generalizing the current type to one more place -- and I'm not convinced that it is.  (I think one would also need to show that it's not reducing the quality for the existing users.)  I think it would be better to define a new type/package, that represents what you want.

(As an example, I could imagine a variant of the current package where the sign bit is an implicit additional bit, that isn't part of either `i` or `f`.  So the bit width of the representation would be `1 + i + f`.  The `fxptGetInt` function would return a value without the sign bit, and a separate `fxptGetSign` would return the sign.  Etc.)

darius-bluespec:
>  ... does PR #249 also need this proviso, at least on the `epsilon` function? that not only is `i` at least 1, but that `i + f` is at least 2?

Indeed.  I have updated #249 to address that.

quark17:
Hi @mtikekar.  I am going to close this old PR, because I did merge the alternative PR.  I want to thank you for contributing, though, and I also want to check-in, to ask if there had been any further developments?  I assume that you worked with a local copy of the FixedPoint library, and I'm curious if you discovered anything else about the library or its assumptions.  Or, for example, whether you ended up writing a new implementation of FixedPoint that's worth contributing back (to the `bsc-contrib` repo, say).  In any case, thank you again and apologies for leaving this open so long.

mtikekar:
I think I just made a local copy of this file in my project. Don't remember if I made any other changes but I would guess not.

PR #197, opened 2020-07-07
Sprinkle some documention for the Deriving fns

bpfoley:


nanavati:
There's a tricky interplay going on between the "raw" primitive variants of undefined / uninitialized values vs the typeclass "make" primitives.

In both cases, the compiler is building out the internal structure of an undefined / uninitialized value for elaboration performance (undefined values) or better error messages with imperative initialization (uninitialized values).

Reaching the "raw" primitive means that there's no more interesting internal structure. This is the case for primitive types (except for a Bit n special case handled with an explicit instance in the prelude) and non-newtype data declarations (newtype data declarations use newtype deriving).

The struct case is when we do want to build out the internal structure (which is done field-by-field), which is why it calls the typeclass method (not the raw primitive).

src/comp/Deriving.hs
```diff
@@ -795,6 +832,10 @@ doSUndefined i vs fs = Cinstance (CQType ctx (TAp (cTCon idUndefined) ty)) [unde
         undef = --trace ("ctx: " ++ ppReadable ctx) $
                 CLValueSign (CDef idMakeUndefinedNQ (CQType [] aty) [CClause [CPVar id_x, CPVar id_y] [] str']) []
 
+-- | Derive Uninitialized for a struct (product type), and return the
```
quark17, line 835: As with `doDUninitialized`, consider calling the class `PrimMakeUninitialzed` and not just `Uninitialized`.

Your description of the construction here is also wrong.  There is no "Raw" used here.

For `doSUndefined` there was a "Raw" primitive used for structures with no fields, otherwise a struct was returned with fields that used one of two functions (either a direct call to the class member or a call to "build" which is a delayed call).

In this case, there is no "Raw" primitive for a struct with no fields -- because a struct with no fields can't be uninitialized -- we just return an empty struct value.  But for any fields, there's still the choice between two possible values: a call to typeclass member (`primMakeUninitialized`) or a call to an elaboration primitive that delays the real call to the typeclass member.  For undefined, we put "build" in the name; here, we did something different and just called it `primUninitialized`.  (Would it be less confusing if we renamed it to also have "build" in the name?)

src/comp/Deriving.hs
```diff
@@ -956,6 +997,9 @@ addRequiredDeriv flags r i tvs clsId derivs =
   -- trace ("auto-derive: " ++ ppReadable (cls, i))
   (CTypeclass clsId) : derivs
 
+-- All types are automatically given instances for the typeclasses in
+-- requiredClasses. Implement this by adding the classes to the derive
```
quark17, line 1001: This only adds deriving if an explicit instance is not already given.  All types are required to have these instances, but the user (or library) can define their own instance.  By writing an explicit instance, they preempt BSC from deriving one.

bpfoley:
> Hopefully these comments are helpful?
> 
> GitHub won't let me tag lines that aren't in the diff, so I'll also mention here: You skipped over adding a comment to`doDDeepSeqCond` and to `doSDeepSeqCond`. Was that intentional?

More because I didn't really know what these do, and I didn't want to do the research at the time. I'll add something.

src/comp/Deriving.hs
```diff
@@ -840,6 +920,10 @@ doSUninitialized i vs fs = Cinstance (CQType ctx (TAp (cTCon idClsUninitialized)
         uninit = --trace ("ctx: " ++ ppReadable ctx) $
                 CLValueSign (CDef idMakeUninitializedNQ (CQType [] aty) [CClause [CPVar id_x, CPVar id_y] [] str]) []
 
+-- | Derive the PrimDeepSeqCond typeclass for struct (product) types.
+-- Fully evaluate the struct by, for each field calling primDeepSeqCond
```
quark17, line 924: Perhaps a comma after "for each field" would make this wording read less awkwardly?  And remove the comment before "or" in the next line?

bpfoley:
Thanks for the detailed comments. Applied. And I've used your rewrites wholesale. Thanks!

PR #202, opened 2020-07-13
Show version

bpfoley:


quark17:
When we deleted an error message, we wanted to retire its identifer (`Parse 6`, `Type 85`,  `Generate 51`, etc) to avoid future confusion.  For that reason, we wanted a record in the source code as a reminder not to use that identifer.  So that's why we left in commented out code.

We can certainly delete the arms of the `ErrMsg` type, like this:
```
--        | EBSVReservedWord              -- BSV reserved word in a Classic program
```
And we can certainly delete the commented out blocks of code from `getErrorText`, like this:
```
--getErrorText (ESyntaxUnknown) =
 --    (Parse 6, empty, s2par "unknown parse error (BSC BUG)")
```
But I would like to put, in its place, a single comment that says the identifier has been retired.  For example, there's this line (which you missed deleting, probably because you were only looking for block quotes?):
```
-- removed Type 86, EWeakCtxPrimIndexNeedsAddCtx
```
I see that you did delete some that were expressed as block comments:
```
{-
 Parse 61 obsoleted by EBadAttributeName
 -}
```
If it helps, those can be made into single-line comments.

bpfoley:
Ah, OK, that makes sense. I'll go back re-add placeholders.

Were there ever any cases where errors/warnings were un-retired?

src/bluetcl/makedepend.tcl
```diff
@@ -55,7 +56,7 @@ if { $rootfiles == "" } {
 }
 
 # scan options looking for -o
```
quark17, line 58: Should this comment also say that we're scanning for `-no-show-timestamps`?

src/comp/FlagsDecode.hs
```diff
@@ -1143,14 +1138,6 @@ externalFlags = [
          (Toggle (\f x -> f {genABinVerilog=x}) (showIfTrue genABinVerilog),
           "include generated Verilog in .ba files", Hidden)),
 
-        ("esecomp",
-         (Toggle (\f x -> f {esecomp=x}) (showIfTrue esecomp),
-          "source is from eseparse", Hidden)),
-
-        ("expand-ATS-def",
```
quark17, line 1150: In `AExpand.hs`, the exported functions have an argument called `expnond`.  In the past, the `expand-ATS-def` flag was fed in as that argument.  At the moment, the few callers of this function only pass in `False`.

There's a single function in `AExpand` (`aoptExpandTest`) that defines the test for whether an identifier should have its definition inlined.  When `expnond` is `False`, the function only considers inlining identifiers that have "bad names" (usually an arbitrary identifier created by the compiler) or names that the compiler created based on the assigned expression.  When `expnond` is set to `True`, the test considers inlining all identifiers.  (The name `expnond` comes from "expand non-d names", when it used to be that the definition of "bad names" were names that the compiler generated starting with the prefix `__d` and a unique number.)

So the `expand-ATS-def` flag, at one time, could be turned on to allow inlining for all identifiers.  That was necessary (more than 14 years ago) because of the way that backend optimizations happened back then.  It used to be that `AOpt` could only optimize if something first expanded identifiers to reveal the expressions that the `AOpt` rules matched on.  (Experiments have also shown that there's less need to do a lot of optimization in the backend, since the Verilog tools that take the output are equally good at applying optimizations at that level.)

It's unclear when this flag or the `expnond` argument stopped being used, but it could be as much as 14-15 years ago.  If we're removing the flag, should we also consider removing the `expnond` argument from those functions?  I can't decide.  It may be safer to leave it (but better document it and maybe give it a better name); if we do remove it, we may still want to leave a comment in `AExpand` about inlining, so that if someone does go investigating that module in the future, they know that it's a possible place of exploration.

src/bluetcl/makedepend.tcl
bpfoley: Done.

src/comp/FlagsDecode.hs
bpfoley: Interesting! Thanks for the history lesson. (So much institutional knowledge, so little time!) If `expnond` isn't being used any more, it sounds like it would be worth removing. Do you mind if I do it in a separate PR in case it causes problems?

quark17: Yes, a separate PR is good.  In fact, I was in the middle of a review of `Error.hs` and was going to say this at the end of that review (still to come):

I think it would have been easier if this PR had been three separate PRs: Remove dead flags; remove dead messages; add new flag.  I would have already merged the new flag commits, but there were edits to be made regarding flags/messages; now you've added more commits related to the new flag and that's now holding up being able to merge the changes for removing dead flags.  (Could it even have been worth doing separate commits/PRs for individual groups of flags? for instance, a PR to just remove the BDD-related flags, that ought to have been removed when CUDD was removed?)

src/comp/Error.hs
```diff
@@ -2984,8 +2918,7 @@ getErrorText (EClockDomain object name clock_domains) =
             quote name ++ ".") $$
      prMethodsByClockDomain clock_domains)
 
--- getErrorText (WInterfaceArg f) =
---  (Generate 8, empty, "Code generation for module " ++ f ++ ", which has an interface argument.  This is unreliable.")
+-- removed Generate 8, WInterfaceArg
```
quark17, line 2921: You could add, "since synthesis of ifc module args is no longer an experimental feature (to warn when used) but is now outright unsupported (an error)."

src/comp/Error.hs
```diff
@@ -3303,27 +3225,9 @@ getErrorText (WOutputResetNoBoundaryClock reset) =
...
--- order to determine if a rule executes between two methods.
-getErrorText (WMethodAnnotChangeArb method1 method2 rules) =
-    (Generate 51, empty,
-     s2par ("The scheduling relationship between method " ++ ishow method1 ++
-            " and method " ++ ishow method2 ++ " was changed to SBR " ++
-            "because rules were arbitrarily placed between them in the " ++
-            "execution order.")
-    )
--}
+-- Removed Generate 51, WMethodAnnotChangeArb
```
quark17, line 3228: Did you want to include the explanation for this (from the deleted comment)?  Or simply say: method-method scheduling annotations are no longer conservatively computed (no longer needing a warning when the compiler has unnecessarily restricted the scheduling)

src/comp/Error.hs
```diff
@@ -3341,25 +3245,9 @@ getErrorText (EPortNamesClashFromMethod m1 m2 port loc ) =
...
--- This feature is now supported by both backends, so no need for the error
-getErrorText (EUnsupportedRuleUrgency target methods rules) =
-    (Generate 56, empty,
-     s2par("Rules more urgent than methods are not supported for " ++
-           target ++ " generation.  The rules:") $$
-     nest 2 (sepList (map text rules) comma) $$
-     s2par("are more urgent than the methods:") $$
-     nest 2 (sepList (map text methods) comma))
--}
+-- Removed Generate 56 EUnsupportedRuleUrgency, now supported by both backends
```
quark17, line 3248: Don't say "now supported by both backends".  That was an incremental update: originally the error could occur in either backend, then we improved things to work in one backend, then we finally removed the error when neither backend could trigger it.

Instead, say: now that rules can be more urgent than methods

src/comp/Error.hs
```diff
@@ -3341,25 +3245,9 @@ getErrorText (EPortNamesClashFromMethod m1 m2 port loc ) =
...
 
-{-
--- This feature is now supported by both backends, so no need for the error
-getErrorText (EUrgentRuleConflict method conflicting_rules) =
-    (Generate 57, empty,
-     s2par("The method " ++ quote method ++
-           " conflicts with rules which are more urgent than it: ") $$
-     nest 2 (sepList (map text conflicting_rules) comma))
--}
+-- Removed Generate 57 EUrgentRuleConflict, now supported by both backends
```
quark17, line 3250: Same reason as for 56 above:  now that rules can be more urgent than methods

src/comp/Error.hs
```diff
@@ -3490,18 +3378,10 @@ getErrorText (EClkGateNotUnused clk) =
...
-    s2par ("Attempt to update an undetermined list/array"))
-getErrorText (EUpdateListOrArrayUndeterminedIndex) =
-   (Generate 79, empty,
-    s2par ("Attempt to update a list/array at an undetermined index"))
-getErrorText (EUndeterminedListOrArrayLength) =
-   (Generate 80, empty,
-    s2par ("Attempt to get the length of an undetermined list/array"))
--}
+
+-- Removed Generate 78 EUpdateUndeterminedListOrArray, handled by primGenerateError
```
quark17, line 3382: These errors still exist!  But instead of being created inside in the Haskell source for BSC, they are created by BH/BSV code for libraries, calling the primitive `primGenerateError`.

We need a comment here in `Error.hs` that tells us that these Error numbers are reserved, but instead of saying that they are reserved because of being retired, we should say that they are reserved for use in library code.

Look for calls to `primGenerateError` in the libraries, for example in `Prelude.bs` in the functions `listPrimUpdate`, `listLength`, `stringLength`, and `arrayLength`.

src/comp/Error.hs
```diff
@@ -3969,9 +3849,7 @@ getErrorText (EMissingPackage p) =
     (System 0, empty, s2par ("Cannot find package " ++ ishow p))
 getErrorText (WFilePackageNameMismatch fname pname) =
     (System 1, empty, s2par ("File name " ++ ishow fname ++ " does not match package name " ++ ishow pname))
---getErrorText (EFailOpenSrcFile fname err_descr) =
---    (System 2, empty,
---     s2par ("Cannot open source file " ++ ishow fname) $$ s2par err_descr)
+-- Removed System 2 EFailOpenSrcFile
```
quark17, line 3852: Add: obsoleted by EFileReadFailure

bpfoley:
Thanks for all the comments. The context is useful and I've added explanations as you suggest.

I'll do the TIMonad cleanup as a separate PR.

And yes, this patchset did metastasize somewhat and would have been better as a separate PR. I thought it was just going to be a quick cleanup, in and out in 20 minutes </rick-and-morty>, but obviously I misjudged.

src/comp/Error.hs
```diff
@@ -3969,9 +3863,8 @@ getErrorText (EMissingPackage p) =
     (System 0, empty, s2par ("Cannot find package " ++ ishow p))
 getErrorText (WFilePackageNameMismatch fname pname) =
     (System 1, empty, s2par ("File name " ++ ishow fname ++ " does not match package name " ++ ishow pname))
---getErrorText (EFailOpenSrcFile fname err_descr) =
---    (System 2, empty,
---     s2par ("Cannot open source file " ++ ishow fname) $$ s2par err_descr)
+-- Removed System 2 EFailOpenSrcFile. Obsolted by EFileReadFailure.
```
quark17, line 3866: Do you mind fixing the type in "Obsoleted"?

src/comp/Error.hs
```diff
@@ -1665,7 +1641,7 @@ getErrorText (EDuplicatePortRenameAttribute attrname method) =
                               quote method ++ "." ))
 getErrorText (EAttributeArgNotAllowed name) =
     (Parse 159, empty, s2par ("Attribute " ++ quote name ++ " does not support an argument in this context." ))
--- Parse 160 (EActionSelfSB has been moved out of parse errors
+-- Removed Parse 160 EActionSelfSB has reclassified as Type 94
```
quark17, line 1644: Do you want to either remove "has" or make it "has been"?  And maybe a comma between the two clauses "Removed Parse 160" and "EActionSelfSB has been..."?

src/comp/Error.hs
```diff
@@ -3969,9 +3863,8 @@ getErrorText (EMissingPackage p) =
     (System 0, empty, s2par ("Cannot find package " ++ ishow p))
 getErrorText (WFilePackageNameMismatch fname pname) =
     (System 1, empty, s2par ("File name " ++ ishow fname ++ " does not match package name " ++ ishow pname))
---getErrorText (EFailOpenSrcFile fname err_descr) =
---    (System 2, empty,
---     s2par ("Cannot open source file " ++ ishow fname) $$ s2par err_descr)
+-- Removed System 2 EFailOpenSrcFile. Obsolted by EFileReadFailure.
+
```
quark17, line 3867: Did you intend an empty line after, but not an empty line before?  I'm not sure there's any consistency -- some of these comments have empty lines around them and some don't -- but it should either have both or none :)

bpfoley: :) Done

src/comp/Error.hs
bpfoley: Argh. Written in too much of a hurry. Done.

src/comp/Error.hs
bpfoley: Woops. Done.

PR #204, opened 2020-07-15
 Silence false positive overflow error on gcc 10

bpfoley:
`0 <= numDigits(v, base) <= 64` for any `v`, `base`. However, since the
value is returned as an `unsigned int`, gcc 10 alerts for a possible
overflow on `char buf[value_width+1]`, although it is a false positive.
By returning a `size_t` (which matches the return type of strlen et al),
the warning is silenced.

Based on a patch by thotypous.

### DIFF
--- a/src/bluesim/dollar_display.cxx
+++ b/src/bluesim/dollar_display.cxx
@@ -737,7 +737,7 @@ const char* print_binary(tFieldDesc& spec, ArgList* args, Target* dest)
   {
     unsigned int value_width = numDigits(v,2);
     pad(spec.width, field_width, value_width, '0', dest);
-    char buf[value_width+1];
+    char buf[(size_t)value_width + 1];
     buf[value_width] = '\0';
     for (unsigned int bit=0, idx=value_width-1; bit < value_width; ++bit,--idx)
       buf[idx] = ((v.data.uVal & (1llu << bit)) != 0llu) ? '1' : '0';


bpfoley:
Based on #172 with `unsigned short` changed to `size_t`. I requested this in the PR review, but the author has been unresponsive for a while, so I've decided to put an amended patch here.

quark17:
The Bluesim code uses `unsigned int` throughout, to represent the bit width of a signal.  The `dollar_display.cxx` file continues to use `unsigned int` for widths, not just in the result of `numDigits` but also in the multiple arguments to `pad`.

If I understand, the problem comes with this code:
```
    unsigned int value_width = numDigits(v,2);
    pad(spec.width, field_width, value_width, '0', dest);
    char buf[value_width+1];
    buf[value_width] = '\0';
```
where it seems that a character buffer has a max size.  If the max size of a character buffer is less than the maximum value of `unsigned int`, then you get the overflow error (if I understand correctly).

Your fix is to have `numDigits` return `size_t`, under the idea that the maximum value of `size_t` should fit within the maximum character buffer size.

But there's no guarantee that `size_t` isn't larger than `unsigned int`, right?  So when you pass `value_width` to `pad`, won't that cause errors/warnings on systems where `size_t` is bigger than `unsigned int`?

(And then, if you change one argument of `pad` to be `size_t`, do you have to start changing all of the arguments to `size_t` and possibly `ssize_t`?)  (And, for that matter, inside the definition of `numDigits`, I think you need to declare the variable `digits` as `size_t` since that's the returned value.)

The result of `numDigits` is being used in two places.  You've fix a type-mismatch in one place, but then introduced a possible mismatch in the other.

The benefit of using `unsigned short` in the original proposal is that you know its value fits within `unsigned int`, so you don't actually need to change `pad` or any other `unsigned int` types where the value is eventually used.  It's still not guaranteed that `unsigned short` won't exceed the character buffer size, it just becomes very unlikely.  If it helps, `numDigits` can even return `unsigned char`, since the value is specified as never being larger than 64.

jrtc27:
> If I understand, the problem comes with this code:
> 
> ```
>     unsigned int value_width = numDigits(v,2);
>     pad(spec.width, field_width, value_width, '0', dest);
>     char buf[value_width+1];
>     buf[value_width] = '\0';
> ```
> 
> where it seems that a character buffer has a max size. If the max size of a character buffer is less than the maximum value of `unsigned int`, then you get the overflow error (if I understand correctly).

I think the issue is the other way round. If `value_width` is `UINT_MAX`, `value_width+1` gets allocated to 0, but you then index it with the non-zero `value_width`, despite the fact that a `UINT_MAX+1` buffer can be allocated on 64-bit systems. It's the fact that ` + 1` on an unsigned int doesn't promote to anything larger, but the compiler wants your size arithmetic to be based on `size_t`.

Using `unsigned short` would fix this, not because it doesn't overflow sizes of buffers, but because `+ 1` on an `unsigned short` promotes to an `int`, which, on any real-world platforms, will be strictly larger than a short, and so adding 1 won't overflow the expression.

bpfoley:
Sigh. Yes, you're right, the root problem C's integer promotion rules that don't promote `value_width+1` beyond int, and it's possible for it to overflow.

I wasn't able to re-open #172, so I just updated this PR to match it.

jrtc27:
Or, alternatively, just make a local fix by changing `char buf[value_width+1];` to `char buf[(size_t)value_width+1];`?

bpfoley:
> Or, alternatively, just make a local fix by changing `char buf[value_width+1];` to `char buf[(size_t)value_width+1];` (or changing the type of the local variable, doesn't really matter)?

Yes. This is definitely the cleaner solution, especially since it avoids the confusion of `WideData` using unsigned ints, yet numDigits using unsigned short.

quark17:
If someone has tested that this resolves the warning with gcc 10, I'm happy with it being merged.

bpfoley:
Yup. I've tested it with gcc 10 (on Ubuntu 20.04)

thotypous:
Sorry for the delay on reviewing this. Thanks a lot for the fix. I have just tested it on gcc 10.1.0 (Arch Linux).

PR #209, opened 2020-07-20
Don't output empty lines if there are no timestamps

bpfoley:


src/comp/TopUtils.hs
```diff
@@ -291,14 +291,15 @@ showCnt cnt s = text $ itos cnt ++ " " ++ s
...
 
-mkGeneratedCComment :: Flags -> IO [String]
-mkGeneratedCComment flags = do
-  suffix <- getClockSuffix flags
+mkGenFileHeader :: Flags -> IO [String]
+mkGenFileHeader flags = do
+  ts <- mkTimestampComment flags
   let compiler = if showVersion flags then version else bsc
-  return $ ["Generated by " ++ compiler, "", suffix, ""]
+      addBlanks = concatMap (\x -> if x == "" then [] else [x, ""])
```
quark17, line 304: Can you add a comment to `addBlanks` to indicate that: it takes a list of single line comments which are to be merged into a single header; deletes any that are empty; and inserts a blank line after each (non-empty) one. (It wasn't immediately obvious to me that it was doing both the delete step and the add-separator step.)

It might have been more readable to just construct the two parts (compiler comment lines and timestamp comment lines) separately and concat them, to be more readable; but I'm ok with abstracting out this function, as long as there's a comment on what it's doing.

src/comp/ASchedule.hs
```diff
@@ -4901,7 +4901,7 @@ dumpConflictMapDOT :: ErrorHandle -> Flags -> String -> AId ->
                       [AId] -> [AId] -> ConflictMap -> IO ()
 dumpConflictMapDOT errh flags prefix modId
                    ifcRuleNames userRuleNames cmap = do
-    suffix <- getClockSuffix flags
+    suffix <- mkTimestampComment flags
```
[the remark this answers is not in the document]
bpfoley: OK, done I think. Minor gotcha that we can't import Flags into Version (because it creates a circular dependency) so we'll have to settle for Bool instead of Flags.

As it turns out, this makes life easier in other places (such as vcdcheck, etc) that don't want to use Flags anyway.

src/comp/TopUtils.hs
bpfoley: Done.

PR #210, opened 2020-07-21
Reduce ghc -O2 FlagsDecode.hs from 135s to 15s(!)

bpfoley:
It turns out that the huge a ++ b ++ c ++ ...  chain in showFlagsRaw
is the issue. It creates a huge set of terms that the simplifier pass
spends a long time trying to reduce.

If instead, we turn it into a list that we intercalate with ",\n", we
a) remove a third of the string terms we concat, and
b) have a list, which the simplifier seems to handle well.

This helps significantly improve the build parallelism on machines
with many cores, and also substantially reduces the single-threaded
compile time.

### DIFF
--- a/src/comp/FlagsDecode.hs
+++ b/src/comp/FlagsDecode.hs
@@ -1790,7 +1790,7 @@ showFlagArg flags f (FRTListString showf) =
 -- user-understandable terms.
 showFlagsRaw :: Flags -> String
 showFlagsRaw flags =
-    let render (k, v) = "\t" ++ k ++ " = " ++ show v
+    let render (k, v) = "\t" ++ k ++ " = " ++ v
         fields =
          [("aggImpConds", show (aggImpConds flags)),
           ("backend", show (backend flags)),
--- a/src/comp/FlagsDecode.hs
+++ b/src/comp/FlagsDecode.hs
@@ -1790,130 +1790,134 @@ showFlagArg flags f (FRTListString showf) =
 -- user-understandable terms.
 showFlagsRaw :: Flags -> String
 showFlagsRaw flags =
-        "Flags {\n" ++
-        "\taggImpConds = " ++ show (aggImpConds flags) ++ ",\n" ++
-        "\tbackend = " ++ show (backend flags) ++ ",\n" ++
-        "\tbdir = " ++ show (bdir flags) ++ ",\n" ++
-        "\tbiasMethodScheduling = " ++ show (biasMethodScheduling flags) ++ ",\n" ++
-        "\tbluespecDir = " ++ show (bluespecDir flags) ++ ",\n" ++
-        "\tcIncPath = " ++ show (cIncPath flags) ++ ",\n" ++
-        "\tcLibPath = " ++ show (cLibPath flags) ++ ",\n" ++
-        "\tcLibs = " ++ show (cLibs flags) ++ ",\n" ++
-        "\tcDebug = " ++ show (cDebug flags) ++ ",\n" ++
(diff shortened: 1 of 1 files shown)


src/comp/FlagsDecode.hs
```diff
@@ -1790,130 +1790,132 @@ showFlagArg flags f (FRTListString showf) =
...
+        "\tshowISyntax = " ++ show (showISyntax flags),
+        "\tshowRangeConflict = " ++ show (showRangeConflict flags),
+        "\tshowModuleUse = " ++ show (showModuleUse flags),
+        "\tshowSchedule = " ++ show (showSchedule flags),
+        "\tshowStats = " ++ show (showStats flags),
+        "\tshowUpds = " ++ show (showUpds flags),
+        "\tstrictMethodSched = " ++ show (strictMethodSched flags),
+        "\tsuppressWarnings = " ++ show (suppressWarnings flags),
+        "\tsynthesize = " ++ show (synthesize flags),
         "\tsystemVerilogTasks = " ++ show (systemVerilogTasks flags) ++ ".\n" ++
```
quark17, line 1896: Did you miss converting this line? (Probably because it mistakenly has a period instead of a comma?)

Out of curiosity, would there be any benefit in further reducing the list to the form:
```
[ ( "aggImpConds", show (aggImpConds flags) ),
  ( "backend", show (backend flags) ),
  ( "bdir", show (bdir flags) ),
```
and then not only insert the newline, but also construct the formatting with the tab and equal sign etc?

bpfoley: Ah. That's what it was: a typo for `,`.

I suspect that turning "\tfoo = " into something where I have to do "\t" ++ foo ++ " = " ++ show bar will be marginally slower (since we now have extra string concatenations), but it would certainly read more nicely.

I suspect it will also be no faster (or perhaps slower) to compile because we've replaced the three expressions (string literal, string concat, show) with three other expressions (string literal, show, pair constructor).

I'll try it out.

src/comp/FlagsDecode.hs
```diff
@@ -1787,133 +1787,136 @@ showFlagArg flags f (FRTListString showf) =
 
 -- This shows the raw flags, rather than showing them in terms of the
 -- externally visible flags.  Use showFlags to display the flags in
--- user-understandable terms.
-showFlagsRaw :: Flags -> String
+-- usshowFlagsRaw :: Flags -> String
```
quark17, line 1790: It looks like there was an accidental delete across these two lines, leaving the beginning of one line and the end of another.

I don't want to step on your toes, so I'll leave it to you to fix.

src/comp/FlagsDecode.hs
```diff
@@ -1787,133 +1787,136 @@ showFlagArg flags f (FRTListString showf) =
...
+          ("vpp", show (vpp flags)),
+          ("vsim", show (vsim flags)),
+          ("verbose", show (verbose flags)),
+          ("verilogDeclareAllFirst", show (verilogDeclareAllFirst flags)),
+          ("warnActionShadowing", show (warnActionShadowing flags)),
+          ("warnMethodUrgency", show (warnMethodUrgency flags)),
+          ("warnUndetPred", show (warnUndetPred flags)),
+          ("vFlags", show (vFlags flags))
+         ]
+        in "Flags {\n" ++
```
quark17, line 1917: This line has a trailing whitespace, flagged by CI

src/comp/FlagsDecode.hs
bpfoley: Thanks. I'm getting tired, so the silly error rate is going up :)

PR #211, opened 2020-07-24
Implement type-level strings

krame505:
Adding type-level strings based on type-level numbers, as discussed in our roadmap for adding generics support.  

This isn't ready to merge yet, more test cases and BSV syntax are still needed.


src/comp/Error.hs
```diff
@@ -1176,6 +1177,12 @@ data ErrMsg
         -- XXX these should contain the type of the constructor
         | EConMismatchNumArgs  String{-String-}      Integer Integer
         | EPartialConMismatchNumArgs  String String{-String-}Integer Integer Integer
+
+        | EStringOf String
+        | EKindStrForNum String
```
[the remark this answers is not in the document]
krame505: Yeah, this is getting a bit unwieldy.  I'm going to leave it for now, but add a comment about this in case we ever add more kinds in the future.

krame505:
Should all be fixed, except for the binary format version strings.  I also added BSV syntax.

src/comp/Error.hs
```diff
@@ -1189,6 +1189,12 @@ data ErrMsg
         | ENoTopTypeSign String
         | WUnusedDef String
         | EConPatArgs String (Maybe String) Int Int
+
+        | EStringOf String
+        | EKindStrForNum String
+        | EKindStrForStar String
+        | EKindStarForStr String
+        | EKindNumForStr String
```
bpfoley: Nit: It seems slightly unfortunate to have the cross product of Kinds here. Would it be better just to have EKindMismatch {- expected -} Kind {- expected -} Kind String?

src/comp/IType.hs
```diff
@@ -94,8 +96,15 @@ cmpT (ITCon _  _ _)      (ITVar _)           = GT
 cmpT (ITCon i1 _ _)      (ITCon i2 _ _)      = compare i1 i2
 cmpT (ITCon _  _ _)      _                   = LT
 
+cmpT (ITNum _)           (ITForAll _  _  _)  = GT
+cmpT (ITNum _)           (ITAp _ _)          = GT
+cmpT (ITNum _)           (ITVar _)           = GT
+cmpT (ITNum _)           (ITCon _ _ _)       = GT
 cmpT (ITNum n1)          (ITNum n2)          = compare n1 n2
-cmpT (ITNum _)           _                   = GT
+cmpT (ITNum _)           _                   = LT
```
bpfoley, line 105: I see the ordering of ITNum vs some types has been changed to LT. What's the thinking here? (Not saying anything is wrong, I just don't have a mental model of what type ordering looks like.)

quark17: This looked odd to me at first, but then I figured it out.  If you look at the whole function, it's written such that: The first constructor is LT than any other constructor; the second constructor is GT than the first but LT anything else; ... the Nth constructor has N-1 explicit GT clauses for the earlier constructors and one catch-all LT clause for anything later ... ; the second to final constructor has explicit GT claus for everything before it, but LT than anything else (which should only match the one remaining constructor); the final constructor is GT than anything else.

So what's happened here is that `ITNum` has gone from being final to being second to final, and `ITStr` has become the final constructor.  So `ITNum` no longer has just one GT clause for everything else; instead, there are now explicit GT clauses for all earlier constructors, and a LT clause for everything else (which should be just `ITStr`).  Now `ITStr` has just the one GT clause for everything else.

src/comp/Error.hs
quark17: I raised this question, but didn't feel strongly enough.  If you feel strongly, we can change it.  However, note two things:

(1) There is already `EUnifyKind` for reporting kind mismatches where the kind can be anything.  This used to be the only message for kind mismatches, but we decided that we could report mismatches between `*` and `#` using special messages that could be friendlier to BSV users (who don't necessarily know what a "kind" is, and aren't exposed to symbols like `*` and `#`).  So we created `EKindNumForStar` and `EKindStarForNum`.  For more complicated mismatches, we resort to `EUnifyKind`.  Although ... mismatches between `*` and `# -> *` can be reported as wrong number of arguments, and mismatches between `* -> *` and `# -> *` can be reported as `EKindNumForStar` on the first argument, etc, so it's unclear to me offhand whether `EUnifyKind` can come up.  But anyway ... with the introduction of '$' kinds, these two `EKindXForY` become six.  I'd be okay with unifying them, but the name would need to reflect the limited cases that it's for, and not simply `EKindMismatch`.  (Maybe `EKindXForY`? or `ESimpleKindMismatch` or something?)

(2) The arguments to the `EKindXForY` constructor couldn't be `Kind`, as you've written, if you're referring to `CType.Kind`.   This is because the `Error` module has to have minimal imports (since it's imported practically everywhere) and can't import `CType`.  I would typically suggest that `EKindXForY` take `String` arguments, since that's what's used nearly everywhere else.  However, we could define `Error.EMsgSimpleKind` (with constructors `Star`, `Num`, `String`) to be used only for this purpose.

src/comp/StdPrel.hs
```diff
@@ -1038,7 +1038,7 @@ tyiInteger = TypeInfo (Just idInteger) KStar [] tiInteger
 
 -- -------------------------
 
--- all preTypes should have identifiers (i.e. be non-numeric) because the usage in MakeSymTab.hs depends on this
+-- all preTypes should have identifiers (i.e. be value) because the usage in MakeSymTab.hs depends on this
```
quark17, line 1041: I'd say "be non-numeric, non-string", since "value" is ambiguous without that context.

Also, this comment is referring to the `ti_qual_id` field of `TypeInfo` in `SymTab.hs`.  The comment there says that numeric kinds have `Nothing` in that field.  What about string types?  I think there a number places where comments say "numeric" and really mean "numeric and string" (or "non-value").

src/comp/TclUtils.hs
```diff
@@ -137,6 +137,9 @@ instance TclObjCvt IType where
     toTclObj (ITNum n) = do
       tn <- toTclObj n
       toTclObj (TLst [TStr "ITNum", TCL tn])
+    toTclObj (ITStr s) = do
```
quark17, line 140: FYI, we'll want to add tests for bluetcl type-analysis commands on string types.

src/comp/TypeAnalysis.hs
```diff
@@ -156,7 +166,7 @@ analyzeType' flags symtab unqual_ty primpair_is_interface = doRight analyze (kin
        case (findType symtab i) of
          Nothing -> Right Unknown
          Just (TypeInfo Nothing k vs ti) ->
-            internalError "analyzeType': non-numeric without Id"
+            internalError "analyzeType': value without Id"
```
quark17, line 169: Again, I'm not sure if "value" is sufficient.  Either "non-numeric and non-string" or maybe "value type".

src/comp/CType.hs
```diff
@@ -319,6 +336,7 @@ cTVarNum name = cTVarKind name KNum
 
 cTCon :: Id -> CType
 cTCon i | all isDigit s = cTNum (read s) (getIdPosition i)
+        | head s == '"' = cTStr (mkFString $ read s) (getIdPosition i)
```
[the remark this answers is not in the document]
krame505: The string here is the quoted and escaped string literal - `read s` extracts the actual string that we want here.

As you mention in a later comment, yes this is used in the BSV parser.  

I will change the `Just KStar` to `Nothing`.

src/comp/StdPrel.hs
krame505: I fixed this and all the other references I found to numeric types that should also be string.

PR #212, opened 2020-07-24
Sprinkle profiling, cabal v3, ghcup docs

bpfoley:


quark17:
Thanks!  Are you okay with my further cleanups?  Since cabal isn't necessary for Debian/Ubuntu, I figure we shouldn't give an install command, since that could easily confuse people.

bpfoley:
Good point. Looks good to me.

PR #214, opened 2020-07-27
Altsyncram init_file_layout value fix

benycze:
Hi all, 

I have found out another bug in the Quartus Verilog template. The current version contained a bad value of INIT_FILE_LAYOUT. More info here:

* https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ug/ug_ram_rom.pdf

The current fix was tested in FPGA and it works for me. It is quite strange that Quartus is fine with a bad parameter value if you are not using the memory init via mif file (during the analysis and fitting).

Best,
Pavel


bpfoley:
Thanks. Sounds good to me, and matches the documentation.

PR #215, opened 2020-07-28
Derive FShow instance for Either.

nanavati:


bpfoley:
LGTM, and doesn't cause any test suite regressions. Thanks!

PR #216, opened 2020-07-28
Add a hie.yaml so haskell-language-server works

bpfoley:
This allows us to get code completion etc in Visual Studio Code and
other editors.

https://mpickering.github.io/ide/posts/2020-07-10-ghc-libdir.html

### DIFF
--- a/src/comp/hie.yaml
+++ b/src/comp/hie.yaml
@@ -1,0 +1,31 @@
+cradle:
+  direct:
+    arguments:
+      - -XForeignFunctionInterface
+      - -XUndecidableInstances
+      - -package base
+      - -package containers
+      - -package array
+      - -package mtl
+      - -package unix
+      - -package regex-compat
+      - -package bytestring
+      - -package directory
+      - -package process
+      - -package filepath
+      - -package time
+      - -package old-time
+      - -package old-locale
+      - -package split
+      - -package syb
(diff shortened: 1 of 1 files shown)


src/comp/hie.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - -package time
+      - -package old-time
+      - -package old-locale
+      - -package split
+      - -package syb
+      - -i../comp:../comp/Libs:../comp/GHC:../comp/GHC/posix:../Parsec
+      - -i../vendor/stp/include_hs
+      - -i../vendor/yices/include_hs
+      - -i../vendor/htcl
+      - -I/usr/include/tcl
```
quark17, line 25: Does it matter that this path is hardcoded?

bpfoley: Oh. Good question. It's like that in the .ghci file too, and I didn't notice it. It doesn't seem to cause any problems.

TBH, I'd much prefer for this file just to contain `- -ghci-script=.ghci`, but for whatever reason that doesn't seem to work with haskell-language-server.

sequencer:
I'm using Sublime with hls, but no luck in opening a bsc project properly.
I tried open bsc root project and comp folder, but neither of them can resolve the `import xxx`
@bpfoley any suggestion to this?

PR #218, opened 2020-07-29
Remove rts_hooks.c in favor of --with-rtsopts

bpfoley:
`rts_hooks` is only used to set the default RTS for bsc, bluetcl and
the extra tools, and all of this can be done more easily with
--with-rtsopts, instead of having extra Makefile machinery.

--with-rtsops has been present since at least GHC 7.0.1.

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -172,7 +172,9 @@ GHCINCLUDES =  \
 	-i$(HTCL_HS)
 GHCTMP = '-tmpdir $(TMPDIR)'
 # Default RTS flags for programs built with a Haskell main
-RTSFLAGS =
+# 256MB heap, 10MB stack, 1 second interval between heap profiles
+# If you modify this, also update rts_opts in ../vendor/htcl/haskell.c
+RTSFLAGS = "-with-rtsopts=-H256m -K10m -i1"
 FVIA ?= -fasm
 GHCFLAGS = \
 	-hide-all-packages \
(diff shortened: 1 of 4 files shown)


src/comp/Makefile
```diff
@@ -172,7 +172,9 @@ GHCINCLUDES =  \
 	-i$(HTCL_HS)
 GHCTMP = '-tmpdir $(TMPDIR)'
 # Default RTS flags for programs built with a Haskell main
-RTSFLAGS =
+# 25GMB heap, 10MB stack, 1 second interval between heap profiles
```
quark17, line 175: Should that 'G' be '6'?

src/comp/Makefile
```diff
@@ -290,78 +287,79 @@ BLUETCL_BUILDLIBS = \
...
 
 .PHONY: showrules
-showrules: $(SOURCES) $(EXTRAOBJS)
+showrules: $(SOURCES)
 	$(PREBUILDCOMMAND)
-	$(BUILDCOMMAND) -main-is Main_$@ $(BUILDFLAGS) $(RTSFLAGS) $(EXTRAOBJ)
+	$(BUILDCOMMAND) -main-is Main_$@ $(BUILDFLAGS) $(RTSFLAGS)
 	$(POSTBUILDCOMMAND)
 
+
```
quark17, line 351: Is this extra blank line intended?

src/comp/Makefile
```diff
@@ -290,78 +287,79 @@ BLUETCL_BUILDLIBS = \
...
+# bluetcl is built in two steps:
+# 1. Build BlueTcl module using -c to emit .hi/.o files, and also the
+# BlueTcl_stub.h reqired by bluetcl_Main.hsc in the next action
```
quark17, line 354: Can you indent the second line three spaces?

src/comp/Makefile
bpfoley: Woops.

src/comp/Makefile
bpfoley: Nope. My bad.

src/comp/Makefile
bpfoley: Done.

PR #223, opened 2020-07-31
Sprinkle Haddock compliant comments

bpfoley:
As an experiment in using haskell-language-server with my editor,
these now show up as tooltips.

Not by any means exhaustive, but a starting point.

See https://haskell-haddock.readthedocs.io/en/latest/markup.html

### DIFF
--- a/src/comp/Error.hs
+++ b/src/comp/Error.hs
@@ -521,8 +521,8 @@ data ErrMsg =
         | EClockedByBadName String [String] -- ^ bad name, available names
         | EResetByBadName String [String] -- ^ bad name, available names
 
+        -- | method s1 generates port s3 conflicting with method s2 at pos
         | EPortNamesClashFromMethod !String !String !String !Position
-          -- | method s1 generates port s3 conflicting with method s2 at pos
         | EPortKeywordClashFromMethod String String
         | EPortNotValidIdentFromMethod String String
 
@@ -702,10 +702,9 @@ data ErrMsg =
         | EUnboundExport String Bool
         | EDuplicateExport String Bool
         | EDuplicateDefInPackageExport String String
-          -- ETypeNotExported:
-          --   type - missing export,
-          --   [(id, pos)] -- definition of id at pos uses type
-        | ETypeNotExported String [(String, Position)]
+        | ETypeNotExported
+          String -- ^ missing export
+          [(String, Position)] -- ^ id defined at pos uses type
         | EMultipleDecl
           String -- ^ var name
           Position -- ^ previous decl position
@@ -791,9 +790,9 @@ data ErrMsg =
         | ELocalRec [String]
         | ENotAnInterface
         | ENotModule String
-        | EPolyField -- ^ reported in old ISyntax, not SV
+        | EPolyField
         | ENotKNum String
-        | EBadGenArg String -- ^ reported in old ISyntax, not SV
+        | EBadGenArg String
         | EBadIfcType String String
         | EBadForeignIfcType String
         | ENoTypeSign String
@@ -821,9 +820,8 @@ data ErrMsg =
             String -- ^ type
             String -- ^ mod
             String -- ^ mod_position
-
         | EForeignModTooManyPorts
-            String  -- ^ method_name
+            String -- ^ method_name
         | EForeignModTooFewPorts
             String -- ^ method_name
         | EForeignModEmptyBody
@@ -1044,8 +1042,10 @@ data ErrMsg =
             [String] -- ^ cycle
             [Doc] -- ^ explanation of edges
             [String] -- ^ other rules in the SCC
-        | EMethodSchedToExec String [String] [Doc]
-              -- method, cycle, explanation of edges
+        | EMethodSchedToExec
+            String -- ^ method
+            [String] -- ^ cycle
+            [Doc] -- ^ explanation of edges
 
         | EInvalidWhen
 
--- a/src/comp/Error.hs
+++ b/src/comp/Error.hs
@@ -673,9 +673,9 @@ data ErrMsg =
         | ESVPRepeatedParamLabel String
         | ESVPUndefinedId String
         | ESVPWrongNumParams
-          String -- ^ name
-          Int -- ^ expected
-          Int -- ^ found
+            String -- ^ name
+            Int -- ^ expected
+            Int -- ^ found
         | ESVPUnmatchedEndIf
         | ESVPNoImportDelimiter
         | ESVPNoClosingParen String
@@ -703,11 +703,11 @@ data ErrMsg =
         | EDuplicateExport String Bool
         | EDuplicateDefInPackageExport String String
         | ETypeNotExported
-          String -- ^ missing export
-          [(String, Position)] -- ^ id defined at pos uses type
+            String -- ^ missing export
+            [(String, Position)] -- ^ id defined at pos uses type
         | EMultipleDecl
-          String -- ^ var name
-          Position -- ^ previous decl position
+            String -- ^ var name
+            Position -- ^ previous decl position
 
         | EForeignNotBit String
         | EPartialTypeApp String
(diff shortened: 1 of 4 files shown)


src/comp/Error.hs
```diff
@@ -2881,19 +2915,20 @@ getErrorText (EConPatArgs c mt expected actual) =
...
--- XXX Errors about consturctor argument mismatches should contain the type of
+-- XXX Errors about constructor argument mismatches should contain the type of
 -- the constructor, but this is overly hard to actually compute.
 getErrorText (EConMismatchNumArgs e {-t-} n1 n2) =
     (Type 143, empty,
      s2par "Too many arguments in the use of the following constructor:" $$
      nest 2 (text e) $$
      s2par ("The constructor expects " ++ show n1 ++
-            " arguments but was used with " ++ show n2 ++ " arguments.") {-$$
+            " arguments but was used with " ++ show n2 ++ " arguments.") {-
```
quark17, line 2925: The `$$` is part of the commented-out code, that would need to be there if the block-comment delimiters were removed.  I think it needs to be kept here (though spacing after the comment delimiter is fine).

src/comp/Error.hs
```diff
@@ -2907,7 +2942,7 @@ getErrorText (EPartialConMismatchNumArgs e t1 {-t2-} n1 n2 n3) =
             then  "with " ++ show n3 ++ " arguments was expected."
             else "was not expected.") $$
      s2par "The expected type is:" $$
-     nest 2 (text t1){- $$
```
quark17, line 2910: Another `$$` that should be kept.

src/comp/Error.hs
```diff
@@ -860,8 +872,11 @@ data ErrMsg
         | EForeignModSynth String
 
         | EForeignFuncStringRes
-        | EForeignFuncBadArgType [(String,Bool)] -- types and which are poly
-        | EForeignFuncBadResType String Bool -- type and whether it is poly
+        | EForeignFuncBadArgType
+            [(String,Bool)] -- ^ types and which are poly
+        | EForeignFuncBadResType
+            String -- ^ type
+            Bool -- is type poly
```
quark17, line 879: Is a `^` missing here?

src/comp/Error.hs
bpfoley: ACK. I removed both of these because `{- $$` was confusing haddock, but I've readded them on another line. Hopefully that will do.

src/comp/Error.hs
```diff
@@ -462,110 +462,108 @@ type MsgContext = Doc
 type EMsg = (Position, ErrMsg)
 type WMsg = EMsg
 
+-- | Lexer/parser error messages
```
quark17, line 465: This is not a comment for the whole type.  I believe it's a comment for the first group (or so) of constructors, indicating that those were errors in lexing/parsing.  The fact that it was before the equal-sign may have been confusing; it should go after the equal-sign and before the first block of constructors.

src/comp/Error.hs
```diff
@@ -622,29 +620,30 @@ data ErrMsg
...
-        | MRestrictions String String  -- please refer to the section on ____ in the Bluespec User Guide for restrictions on using ___
-        | WExperimental String  -- support for ___ is experimental
-        -- duplicate member in (struct, interface, union or typeclass) of
+        | WShadowDecl String Position -- ^ var name, previous declaration pos
+        | WNeverAssigned String -- ^ variable was declared but not assigned
+        | WDeprecated String String String -- ^ what, it's name, optional text
+        | EObsolete String String String -- ^ what, it's name, optional text
+        | MRestrictions String String  -- ^ please refer to the section on ____ in the Bluespec User Guide for restrictions on using ___
+        | WExperimental String  -- ^ support for ___ is experimental
+
```
quark17, line 635: Did you intend to add this blank line?

src/comp/Error.hs
```diff
@@ -1023,23 +1040,39 @@ data ErrMsg
...
+        | EReflexiveUserUrgency
+            String  -- ^ id
+            Position -- ^ pos1
+            Position --  ^ pos2
+
+        | ECombinedSchedCycle
+            [String] -- ^ cycle
+            [Doc] -- ^ explanation of edges
+            [String] -- ^ other rules in the SCC
```
quark17, line 1075: Did you intend to introduce this blank line? (ECombinedSchedCycle and EMethodSchedToExec were next to each, before.)

src/comp/Error.hs
bpfoley: Fixed

src/comp/Error.hs
bpfoley: Removed

src/comp/Error.hs
bpfoley: Removed

src/comp/Error.hs
```diff
@@ -463,109 +463,106 @@ type EMsg = (Position, ErrMsg)
...
 
 data ErrMsg
-
-        -- Lexer/parser error messages
-
-        = EBadCharLit
-        | EBadStringLit
-        | EBadLexChar Char
-        | ECPPDirective -- cpp directive found when not expected
+        = EBadCharLit                   -- ^ Lexer error message
```
quark17, line 466: I think you still misunderstand me?  That comment didn't just apply to the first three constructors ; I think it applies to all the constructors (hundreds?) from here through to `ESVPNoId`.  If you notice, `ESVPNoid` is followed by the comment, "Type checker and static elaboration errors" (with blank lines around it).  If you recall that `ErrMsgTag` has five different tags (like `Parse`, `Type`, etc) -- I believe that the comment "Lexer/parser error messages" was intended (long ago) to label a group of all the `Parse` error constructors.  And the next comment is for the `Type` error constructors.  At one time, the `ErrMsg` constructors might have been grouped by the five different types, and each group might have had a comment like this introducing them.  However, I doubt that the constructors are still in that order; and keeping that in sync may not be worth it.  But if we're not doing an audit of that, maybe we just leave these high level comments as is?

I see that you have kept the comment "Type checker and static elaboration errors" (line 710/708) without any Haddock formatting, and so I think you should do the same with the "Lexer/parser error messages" comment.  If it helps, you can move the equal-sign up to the `data` line, and then keep this comment where it is (with blank lines before and after) and without Haddock formatting, since that's consistent with the later comment for `Type`.

If there's some way to better represent this with the Haddock formatting characters, then that's OK too, but it should be consistently applied to all of the comments of this type.  (I haven't looked to see if all five tags still have a comment like this.)

I also notice that you haven't added formatting for every comment on the constructors here.  For example, there's a comment above `ETypeNotExported` (line 731) that describes the fields of the constructor -- that could be Haddock-ized.  Was that an oversight?

bpfoley: Thanks. I did misunderstand you.

I haven't added haddock formatting to everything simply because I ran out of energy to do them all, and this was just a best effort thing. The more I do in a single PR, the more chances I'll make some mistake :)

src/comp/Error.hs
```diff
@@ -790,11 +791,9 @@ data ErrMsg
         | ELocalRec [String]
         | ENotAnInterface
         | ENotModule String
-          -- reported in old ISyntax, not SV
```
quark17, line 793: Here again is an indented comment.  My guess is that the comment applies to `ENotModule`, not the constructor where you've placed it.

From looking at the pre-github logs, I believe this comment is saying that the `String` argument has been formatted in BSV syntax, not BH/Classic syntax.  Since `EPolyField` does not have an argument, the comment is probably be referring to the `String` argument of `ENotModule`.  I notice, though, that `ENotModule` is not used anywhere!  And `EPolyField` is used in one place with the comment "this should not happen".

Two lines below here, there's another indented comment that was added in the same commit.  It too might apply to the constructor above it (`ENotKNum`) not the one below it (`EBadGenArg`).  I notice that neither of those constructors are used anywhere!

I don't think these comments are relevant any more, so I'll remove them.

PR #229, opened 2020-08-01
Remove compatibility ifdefs for GHC < 7.10

bpfoley:
We were already depending on Applicative being a superclass of Monad
in several places, and prelude exporting more of Control.Applicative,
so compiles already broke with GHC < 7.10.

Furthermore, 7.10.1 is the oldest version we actually test, and it's
the oldest version we document as supported in README.md. It's time
to move on. :)

See https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/7.10

### DIFF
--- a/src/comp/SimExpand.hs
+++ b/src/comp/SimExpand.hs
@@ -43,11 +43,6 @@ import ListUtil (mapFst, mapSnd)
 import SCC (tsort)
 
 import Util (headOrErr, map_insertManyWith, allPairs)
--- traceM was added to Debug.Trace
-#if !defined(__GLASGOW_HASKELL__) || (__GLASGOW_HASKELL__ < 707)
-import Util(traceM)
-#endif
-
 import GraphUtil(extractOneCycle_map, reverseMap)
 
 -- ===============
--- a/src/comp/AAddScheduleDefs.hs
+++ b/src/comp/AAddScheduleDefs.hs
@@ -24,7 +24,7 @@ import Data.List(intersect)
 import Data.Maybe(isJust, fromJust, fromMaybe, maybeToList, mapMaybe)
 
 -- import Trace
--- import Util(traceM)
+-- import Debug.Trace(traceM)
 
 -- -------------------------------------------------------------------
 -- This file contains all of the logic for creating defs describing
--- a/src/comp/ABinUtil.hs
+++ b/src/comp/ABinUtil.hs
@@ -31,7 +31,7 @@ import GenABin(readABinFile)
 
 import qualified Data.Map as M
 
---import Util(traceM)
+--import Debug.Trace(traceM)
 
 -- ===============
 
--- a/src/comp/AConv.hs
+++ b/src/comp/AConv.hs
@@ -43,7 +43,7 @@ import Data.Traversable(forM)
 --import PreStrings(fsUnderscore)
 
 --import Trace
--- import Util(traceM)
+-- import Debug.Trace(traceM)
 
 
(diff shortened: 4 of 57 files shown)


src/comp/SimExpand.hs
```diff
@@ -43,11 +43,6 @@ import ListUtil (mapFst, mapSnd)
 import SCC (tsort)
 
 import Util (headOrErr, map_insertManyWith, allPairs)
--- traceM was added to Debug.Trace
```
quark17, line 46: Can you also, then, remove `traceM` from `Util.hs`?

bpfoley: Done.

src/comp/Makefile
```diff
@@ -130,6 +130,7 @@ endif
 
 PACKAGES = \
 	-package base \
+	-package ghc \
```
quark17, line 133: What is this requirement?  Is it for these changes, or is it something separate?

bpfoley: It's needed for Debug.Trace

quark17: But we were already importing Debug.Trace in several places, so why wasn't it needed before?  (I also couldn't find any documentation that Debug.Trace required this package, but it's also hard to do a general search with the name "ghc"!)

quark17:
I notice that a few places import just "Trace" (mostly commented out, but there are a few uncommented out as well).  I believe this is referring to `bsc/src/comp/GHC/Trace.hs`, which just re-exports `Debug.Trace`.

In the beginning, twenty years ago, the source code could be compiled with several different Haskell compilers.  Any code that was compiler-specific was abstracted in a module that could be defined differently for each compiler.  The subdirectory `GHC` was where GHC's versions were kept.  GHC-specific code is no-longer constrained to that directory (and thus we only support GHC now).  We could eliminate that subdirectory and inline any of its packages.  There are only three: `BinaryIO` (read/write of binary files), `TmpNam`, and `Trace`.

For `Trace`, we could replace any imports of it with an import of `Debug.Trace`.  (It looks like `Deriving.hs` has commented out imports of both.)  Did you want to do that as part of this PR?

src/comp/Makefile
bpfoley: That's funny. I did at one point get a GHC error about it not being able to find `Debug.Trace` because it's a hidden package and you need to add `-package ghc` to make it visible. But I can't reproduce that now. And as you say there are a few modules that already import Debug.Trace. Strange.

bpfoley:
> I notice that a few places import just "Trace" (mostly commented out, but there are a few uncommented out as well). I believe this is referring to `bsc/src/comp/GHC/Trace.hs`, which just re-exports `Debug.Trace`.
> 
> In the beginning, twenty years ago, the source code could be compiled with several different Haskell compilers. Any code that was compiler-specific was abstracted in a module that could be defined differently for each compiler. The subdirectory `GHC` was where GHC's versions were kept. GHC-specific code is no-longer constrained to that directory (and thus we only support GHC now). We could eliminate that subdirectory and inline any of its packages. There are only three: `BinaryIO` (read/write of binary files), `TmpNam`, and `Trace`.
> 
> For `Trace`, we could replace any imports of it with an import of `Debug.Trace`. (It looks like `Deriving.hs` has commented out imports of both.) Did you want to do that as part of this PR?

I saw that too, and I decided I didn't want the scope of this PR to grow too much (that's caused us problems before). I'll do it in a followup PR if that's OK.

PR #231, opened 2020-08-03
Strip some dead functions from Error

bpfoley:


quark17:
This is a library that provides an API.  Just because we're not using some of the functions in the API doesn't necessarily mean that it should be deleted from the API.

The function `bsWarningsAndErrors` takes a list of warnings, a list of errors, and a list of demotable errors.  Most places in the code base just want to report just a warning or report just an error, and so the functions `bsWarning` and `bsError` exist for those situations.  These are just wrappers around `bsWarningsAndErrors` that provide empty lists for the other arguments.  Functions like `bsWarning` are less error-prone and self-documenting.  Currently, no code uses `bsErrorDemotable`, but we should keep it as part of the API.  Further, we should make BSC use it more, in place of `bsError`.  At the moment, very few errors are demotable, and we should change that.

Maybe we don't need `bsErrorDemotableContext`, until there's some code that needs it, but it also doesn't hurt to leave it for now, for completeness of the API.

PR #239, opened 2020-08-05
Remove unused #includes in bluesim kernel

bpfoley:
Noticed a couple by hand, and then cleaned up with iwyu.

While here, separate .o compilation and .d generation into separate
rules, so doesn't rerun $(CXX) -MM every time you want to rebuild an
object.

### DIFF
--- a/src/bluesim/event_queue.cxx
+++ b/src/bluesim/event_queue.cxx
@@ -1,5 +1,4 @@
 #include <vector>
-#include <algorithm>
 #include <cstdio>
 
 #include "event_queue.h"
--- a/src/bluesim/kernel.cxx
+++ b/src/bluesim/kernel.cxx
@@ -1,16 +1,13 @@
 #include <list>
 #include <algorithm>
 #include <cstring>
-#include <ctime>
 #include <cstdio>
 
 #include <pthread.h>
 #include <signal.h>
-#include <errno.h>
 
 #include "mem_alloc.h"
 #include "kernel.h"
-#include "bs_reset.h"
 #include "bs_module.h"
 #include "plusargs.h"
 #include "version.h"
--- a/src/bluesim/mem_file.cxx
+++ b/src/bluesim/mem_file.cxx
@@ -4,7 +4,6 @@
 #include <ctype.h>
 #include <errno.h>
 
-#include "bluesim_kernel_api.h"
 #include "bs_wide_data.h"
 #include "bs_mem_defines.h"
 #include "bs_mem_file.h"
--- a/src/bluesim/portability.cxx
+++ b/src/bluesim/portability.cxx
@@ -4,11 +4,8 @@
 #include <cstdarg>
 #include <cstdlib>
 #include <cstdio>
-#include <cstring>
-#include <sys/types.h>
 #include <errno.h>
 #include <unistd.h>
-#include <fcntl.h>
 
 #include "portability.h"
 
--- a/src/bluesim/wide_data.cxx
+++ b/src/bluesim/wide_data.cxx
@@ -1,6 +1,5 @@
 #include <cstdlib>
 #include <cstring>
-#include <algorithm>
 #include <csignal>
 
 #include "bs_wide_data.h"


quark17:
The PR comment says that it also includes Makefile changes, but I don't see that.

I have no objection to removing unused system libraries, but it looks like this PR is also shuffling around the local headers and I need to understand if that's correct to do.  Remember that this directory is not the only user of the headers -- BSC generates C++ source files (for Bluesim and SystemC) that include headers here.

I see, for example, that `bluesim_kernel_api.h` has been moved out of the imports from `kernel.h`.  I need to understand if that makes sense to do.  (My default assumption is that the structure of imports is intentional.  This is not code that has had lots of authors or evolved organically; it was architected and mostly written by a single person, who was very experienced and intentional.)

bpfoley:
> The PR comment says that it also includes Makefile changes, but I don't see that.

ACK. Again we've been bitten by GitHub only showing the original commit message, and not the newer version if it's been modified. I undid the makefile changes and removed the commit message comment when I realized it wasn't the right thing to tdo.


> I have no objection to removing unused system libraries, but it looks like this PR is also shuffling around the local headers and I need to understand if that's correct to do. Remember that this directory is not the only user of the headers -- BSC generates C++ source files (for Bluesim and SystemC) that include headers here.

Sure. What I did was I used include-what-you-use to remove all the unused system headers, and then any others that it suggested.

> I see, for example, that `bluesim_kernel_api.h` has been moved out of the imports from `kernel.h`. I need to understand if that makes sense to do. (My default assumption is that the structure of imports is intentional. This is not code that has had lots of authors or evolved organically; it was architected and mostly written by a single person, who was very experienced and intentional.)

You could well be right, but in this particular case, I think all the bluespec_*.h headers are ones that specify public APIs, and kernel.h is an implementation header file corresponding to kernel.cpp. As we can see, the only things that import `kernel.h` are `vcd.cxx`, `kernel.cxx`, `module.cxx`, `reset.cxx`, `plusargs.cxx`, the other implementation files.

Furthermore, `kernel.h` isn't installed into `bsc/inst/lib/Bluesim/`, only files prefixed with `bluesim_` and `bs_`.

Looking in the source tree, there don't appear to be that many includes:
```
SimBlocksToC.hs
223:        model_includes = [ cpp_include "bs_model.h"
312:                  [ cpp_system_include "cstdlib"
313:                  , cpp_system_include "time.h"
314:                  , cpp_include "bluesim_kernel_api.h"
315:                  , cpp_include "bs_vcd.h"
316:                  , cpp_include "bs_reset.h"
634:      c_includes  = [ cpp_include "bluesim_primitives.h"
637:                         then [cpp_include "imported_BDPI_functions.h"]
640:      h_includes  = [ cpp_include "bluesim_types.h"
641:                    , cpp_include "bs_module.h"
642:                    , cpp_include "bluesim_primitives.h"
643:                    , cpp_include "bs_vcd.h"

SystemCWrapper.hs
79:        h_includes   = [ cpp_include "bluesim_kernel_api.h"
83:        cxx_includes = [ cpp_include "bluesim_systemc.h"

VPIWrappers.hs
83:         c_segments = [ cpp_system_include "stdlib.h"
84:                      , cpp_system_include "vpi_user.h"
85:                      , cpp_include "bdpi.h"
106:         h_segments = [ cpp_system_include "vpi_user.h"
```
and
```
main.cxx.template
3:#include <getopt.h>
5:#include "bluesim_kernel_api.h"
7:#include "model_BLUESIM_TOP_MODULE.h"
```

quark17:
> `kernel.h` is an implementation header file corresponding to `kernel.cpp`
>  Furthermore, `kernel.h` isn't installed into `bsc/inst/lib/Bluesim/`

Exactly.  But there's the additional question: Is `bluesim_kernel_api.h` a subset of `kernel.h`?  I was thinking that everything declared in `bluesim_kernel_api.h` is implemented in `kernel.cxx`.  That is: Normally, everything that `kernel.cxx` exports would be declared in `kernel.h`.  But the code in this directory is being linked into a library, and we only want to expose a subset of those functions, to the ultimate users of the library.  As you say, we don't release `kernel.h`, and that's because we don't want to expose all of it.  So, we put some of the functions into a file that we do want to export, `bluesim_kernel_api.h`, which is imported by `kernel.h` (to avoid writing those declarations in two places).  That way, local code can import `kernel.h` and external users can import just `bluesim_kernel_api.h`.

One problem is that I do see that some local code does import `bluesim_kernel_api.h`, and not just `kernel.h`.  So that leads me to wonder: Is it true that everything declared in `bluesim_kernel_api.h` is found in `kernel.cxx`?  Or does it include declarations from various files that are linked into `libbskernel.a`?

If we decide that `bluesim_kernel_api.h` contains functions that could be implemented in any of the source files/objects linked into `libbskernel.a` (so not just `kernel.cxx` but also `vcd.cxx` etc), then it makes sense that `kernel.h` shouldn't import it.  I'm OK with that.  But then I would expect that `kernel.h` would have to declare everything available from `kernel.cxx` (and `vcd.h` similarly everything available from `vcd.cxx`) and the file `bluesim_kernel_api.h` would also contain declarations (we'd be duplicating the declarations that we want to expose to the end user).

Right now, `kernel.h` avoids declaring everything, because it can import `bluesim_kernel_api.h`.  Your PR removes this import, but doesn't replace it with explicit declarations.  So you PR seems to be in this middle ground where `kernel.h` is not the full implementation header for `kernel.cpp` -- other source files that need functions declared in `bluesim_kernel_api.h` need to import that, because `kernel.h` isn't declaring them.

It seems like we should go all one way (local files should only import `kernel.h` and not `bluesim_kernel_api.h`) or the other (`bluesim_kernel_api.h` is duplicated declarations from `kernel.h` and possibly other headers), but your PR is a middle ground that doesn't seem right.

(If it helps, rather than duplicating declaraions in `bluesim_kernel_api.h`, each header could put its api declaraions in a separate file.  So `kernel.h` would import `kernel_api.h` and `vcd.h` would import `vcd_api.h` etc; and then `bluesim_kernel_api.h` would contain only imports of `kernel_api.h` and `vcd_api.h` etc.  But my guess is that only `kernel_api.h` would need to exist, and so `bluesim_kernel_api.h` is currently filling that role.)

bpfoley:
Hmm. This is part way between one and the other isn't it?

OK, I don't really feel like pulling on this particular thread at the moment (and I'm not sure it adds much value right now), so I've just reduced the patch to remove unused headers from the `.cxx` files. Sound better?

PR #240, opened 2020-08-05
Get bluesim to handle SIGPIPE the same as Ctrl-C

### LINKED ISSUE #233: Bluespec simulation ignores SIGPIPE, hangs when piped to head
A trivial test such as 

```
package Forever where
{-# verilog forever #-}
forever :: (IsModule m c) => m Empty
forever =
  module
    rules "test": when True ==> $display "time %0t" $time
```

When compiled and run with
```
bsc -sim Forever.bs
bsc -sim -e forever forever.ba
./a.out | head
```

Displays 10 lines and then hangs instead of terminating as we'd expect. Running the same simulation under a verilog simulator terminates on SIGPIPE.

```
time 0
time 10
time 20
time 30
time 40
time 50
time 60
time 70
time 80
time 90
```

bpfoley:
ie gracefully shut down the simulation threads ASAP.

This means ./mysim.exe | /usr/bin/head works as expected.

Fixes #233

### DIFF
--- a/src/bluesim/kernel.cxx
+++ b/src/bluesim/kernel.cxx
@@ -73,30 +73,30 @@ static void wait_for_sim_stop(tSimStateHdl simHdl)
 
 // list of sims that have registered their interest in the signal
 // (use a list, so that iterators are still valid after an erase)
-static std::list<tSimStateHdl> control_c_watchers;
+static std::list<tSimStateHdl> abort_watchers;
 
-static void control_c_handler(int /* unused */)
+static void abort_handler(int /* unused */)
 {
   std::list<tSimStateHdl>::iterator it;
-  for (it = control_c_watchers.begin(); it != control_c_watchers.end(); it++)
+  for (it = abort_watchers.begin(); it != abort_watchers.end(); it++)
     bk_abort_now(*it);
 }
 
-static void add_control_c_watcher(tSimStateHdl simHdl)
+static void add_abort_watcher(tSimStateHdl simHdl)
 {
-  control_c_watchers.push_back(simHdl);
+  abort_watchers.push_back(simHdl);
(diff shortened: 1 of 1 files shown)


src/bluesim/kernel.cxx
```diff
@@ -116,15 +116,19 @@ static void* sim_thread(void* ptr)
...
-  sigaction(SIGINT, &sigint, NULL);
+  /* install signal handlers to shut down simulation */
+  struct sigaction sa;
+  sa.sa_flags = 0;
+  sa.sa_handler = abort_handler;
+  sigemptyset(&sa.sa_mask);
+  /* SIGINT (user types Ctrl-C) */
+  sigaction(SIGINT, &sa, NULL);
+  /* SIGPIPE (usually stdout piped to a program (eg /usr/bin/head) that exits)
+   */
```
quark17, line 127: Is there a reason that the close-comment symbol is on a new line?

This file doesn't seem to be consistent in its use of quoting symbols -- I notice that it uses `//` for one-line comments in other places.  I'd tend to want to avoid block-quotes for one-line comment (just to avoid accidentally not closing the quote), but I'm OK with leaving it as block-quotes here (but I'd keep it as one line).

bpfoley: No particular reason, except the comment itself caused the line to be >79 chars and my editor auto-wrapped it. I've reworded it to bring it back under 80. :)

I agree: I'm not particularly delighted by a mix of formatting styles, and generally I'd use
`//` for single line comments too, but I hadn't really examined the file as a whole and was just staying consistent with the local style. (Besides, patches that just change formatting, or worse, patches that intermix formatting and functionality fixes are often frowned upon.)

PR #241, opened 2020-08-05
Apply substitution to reduced predicates before returning from reduce…

nanavati:
Apply substitution before returning reduced predicates from reducePredsAggressive

This fixes an "assert failed IAp" bug in the attached files (and makes an apSub in ctxreduce redundant).
[FailIAp2.bs.txt](https://github.com/B-Lang-org/bsc/files/5026674/FailIAp2.bs.txt)
[FailIAp.bs.txt](https://github.com/B-Lang-org/bsc/files/5026675/FailIAp.bs.txt)


quark17:
Looks good, thanks!  And doesn't break the testsuite.

PR #244, opened 2020-08-06
Fix missing case for deserializing TyStr

krame505:
This fixes binary deserialization for TyStr, which apparently was missed due to the default case.

### DIFF
--- a/src/comp/BinData.hs
+++ b/src/comp/BinData.hs
@@ -1302,6 +1302,7 @@ instance Bin TyCon where
              0 -> do i <- fromBin; mk <- fromBin; s <- fromBin;
                      return (TyCon i mk s)
              1 -> do i <- fromBin; pos <- fromBin; return (TyNum i pos)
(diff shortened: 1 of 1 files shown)


quark17:
This mistake could have been caught if the testsuite had ever tried to read a `.bo` file with this encoding.  So far, the string tests are contained in single files and do not import other packages (which would involve reading in their `.bo` files).

Having the tests in a single file is fine, however, because there is a feature of the testsuite, for testing that generated `.bo` files are well-formed.  There is a tool called `dumpbo` which will read in a `.bo` files and print its contents, that can be built in `bsc/src/comp/` (with `make dumpbo` for example) but is not by default.  (A number of tools are not built by default and can be bult with `make install-extra` in `bsc/src/comp/`.)  The testsuite will look to see if `dumpbo` exists in the install directory and, if so, will run it on the top-level generated `.bo` files.  In the future, if you edit the encoding, you should build the `dumpbo` tool and test with it.  (Also, poking @bpfoley , to take note for when we add testsuite to the CI.)

PR #245, opened 2020-08-07
Add a user hook to Makefile for adding GHC flags.

jinwoo:
This will make it easy to adapt the build to the user's build environment without modifying the Makefile.

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -186,6 +186,7 @@ GHCFLAGS = \
 	$(PACKAGES) \
 	$(GHCINCLUDES) \
 	$(GHCTMP) \
+	$(GHCUSERFLAGS) \
 
 # GHC will recompile a module if certain flags (such as -I) have changed.
 # To avoid recompiling the shared modules each time we compile a tool,


jinwoo:
The CI failure on macos seems unrelated to this PR. PR 244 also has the same failure.

quark17:
There should definitely be a way for the user to add flags to the GHC command line.

What's the best way to do that?  My first thought would have been to change the assignment for `GHCFLAGS` in the Makefile to use `+=` so that it adds to whatever flags the user may have already set.  Is using `GHCUSERFLAGS` better?  I'd guess that it's less likely to accidentally pull in flags that weren't intended.

Note that, while not a clean way to do it, the user can currently add flags by defining `GHC` in the environment, because of this in the makefile:
```
GHC ?= ghc
```
One could insert flags at the start of the GHC command that way.  Though it'd of course be better to use `GHCFLAGS` or `GHCUSERFLAGS` for that.

jinwoo:
> What's the best way to do that? My first thought would have been to change the assignment for `GHCFLAGS` in the Makefile to use `+=` so that it adds to whatever flags the user may have already set. Is using `GHCUSERFLAGS` better? I'd guess that it's less likely to accidentally pull in flags that weren't intended.

Using `+=` sounds like a good alternative. What I'm worried about though is that some GHC flags are order-dependent. Especially when you need to "reset" flags that are previously set. So I think ideally the user flags hook should come at the end.

After sending this PR out, I also realized that there are a few more places where I need to do something similar. So we may have to devise a general pattern and use it in a consistent manner wherever needed.

> 
> Note that, while not a clean way to do it, the user can currently add flags by defining `GHC` in the environment, because of this in the makefile:
> 
> ```
> GHC ?= ghc
> ```
> 
> One could insert flags at the start of the GHC command that way. Though it'd of course be better to use `GHCFLAGS` or `GHCUSERFLAGS` for that.

It seems to me that this is quite "hacky" and also has the same problem regarding the flag order.

I also have to mention that these user hooks may not be sufficient for my purpose and I may have to just make local modifications to the Makefile anyway. But providing user hooks is probably beneficial in general?

jinwoo:
Inserting flags to `GHC` worked in our case. Because this PR doesn't particularly look better than using `GHC`, I'm going to close this.

PR #248, opened 2020-08-13
Use -show-timestamps and -show-version in model codegen

bpfoley:
Make the C++ get_creation_time return 0 for -no-timestamps.
Make the bluesim wrapper have a 0 creation_time for -no-timestamps.
Make the C++ get_version return 0,0,NULL,"" for -no-show-version.

The C++ changes produce more reproducable builds which leads to
better caching with CCache et al.

### DIFF
--- a/src/comp/SimBlocksToC.hs
+++ b/src/comp/SimBlocksToC.hs
@@ -484,7 +484,9 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
                                ]
         parts = wordsBy (=='.') versionname
         (year, month, annotation) =
-          if ((length parts == 2) || (length parts == 3)) &&
+          if not (showVersion flags) then
+                  (mkUInt32 0, mkUInt32 0, mkNULL)
+          else if ((length parts == 2) || (length parts == 3)) &&
              (all isDigit (parts!!0)) &&
              (all isDigit (parts!!1))
           then let y = mkUInt32 (read (parts!!0))
@@ -495,7 +497,8 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
                    m = mkUInt32 0
                    a = if (versionname == "") then mkNULL else mkStr versionname
                in  (y, m, a)
-        build = mkStr buildVersion
+        bv = if showVersion flags then buildVersion else ""
+        build = if bv == "" then mkNULL else mkStr bv
         gv_def =
           define get_version
                  (block [ stmt (cDeref (var "year"))       `assign` year
@@ -507,7 +510,9 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
         get_creation_time = function (userType "time_t")
                                      (mkScopedVar "get_creation_time")
                                      []
-        (TimeInfo _ clock_time@(TOD t _)) = creation_time
+        (TimeInfo _ clock_time@(TOD t _)) = if (timeStamps flags)
+                                            then creation_time
+                                            else TimeInfo 0 (TOD 0 0)
         time_str = calendarTimeToString (toUTCTime clock_time)
         gct_def = define get_creation_time
                          (comment time_str (ret (Just (mkUInt64 t))))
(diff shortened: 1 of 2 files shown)


bpfoley:
Some interesting stats from this btw. On my 36 core Xeon, doing a `make -j36 checkparallel` without this patch causes a considerable number of CCache cache misses because of the timestamps in the C++ source. (I'm not trying to use -O0, or lld or anything like this.)

As a reminder `bsc_link_objects` generates C++ source from modules and then compiles it into object files using the C++ compiler.

On a completely empty cache, it takes approximately 1700s of a 5120s build:
```
  WALLCLOCK sec        CPU sec      SYSTEM sec   #calls  CATEGORY
  1695.0 ( 33%)   1446.2 ( 35%)    244.7 ( 25%)    1174  bsc_link_objects
  5120.2 (100%)   4080.5 (100%)    992.9 (100%)   14967  TOTAL
```

On a second run, `bsc_link_objects` reduces to ~870s, but because of the C++ timestamps, the CCache hit rate is <60%.

```
  WALLCLOCK sec        CPU sec      SYSTEM sec   #calls  CATEGORY
   870.8 ( 20%)    691.2 ( 20%)    174.4 ( 18%)    1174  bsc_link_objects
  4449.5 (100%)   3444.1 (100%)    958.4 (100%)   14967  TOTAL

cache hit (direct)                   423
cache hit (preprocessed)            1248
cache miss                          1162
cache hit rate                     58.98 %
```

With this patch, the second run of the test suite has an almost perfect hit rate (it would be 100% but some of the tests generate inputs using random numbers), shaving another ~550s off the build time.

```
  WALLCLOCK sec        CPU sec      SYSTEM sec   #calls  CATEGORY
   329.2 (  9%)    226.3 (  8%)     97.1 ( 11%)    1174  bsc_link_objects
  3827.5 (100%)   2906.9 (100%)    871.8 (100%)   14967  TOTAL

cache hit (direct)                  1215
cache hit (preprocessed)            1660
cache miss                             3
cache hit rate                      99.9 %
```

src/comp/SimBlocksToC.hs
```diff
@@ -495,7 +497,7 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
                    m = mkUInt32 0
                    a = if (versionname == "") then mkNULL else mkStr versionname
                in  (y, m, a)
-        build = mkStr buildVersion
+        build = mkStr (if showVersion flags then buildVersion else "")
```
quark17, line 500: I notice that for `annotation`, we specifically return `mkNull` and not an empty string.  Do you know offhand if the same should be done for `build`?   Should this be defined as:
```
if not (showVersion flags) then mkNull
else if (buildVersion == "") then mkNull
else mkStr buildVersion
```
I think that might be preferable, but I guess it would depend on whether any users are expecting the pointer to be non-null.  But I would think it should be OK.  It's this generated code that we're talking about (for a Bluesim model rooted at `mkMod`:
```
void MODEL_mkMod::get_version(unsigned int *year,
                              unsigned int *month,
                              char const **annotation,
                              char const **build)
{
  *year = 0u;
  *month = 0u;
  *annotation = NULL;
  *build = "d142ff8";
}
```
If the build string is empty, we might as well assign to null.

bpfoley: It's probably OK. The only place that we actually use it ourselves is in the bluetcl `bk_version`, and that uses `fromCString` for both `annotation` and `build`.

src/comp/SimBlocksToC.hs
```diff
@@ -495,7 +497,7 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
                    m = mkUInt32 0
                    a = if (versionname == "") then mkNULL else mkStr versionname
                in  (y, m, a)
-        build = mkStr buildVersion
+        build = if showVersion flags then mkStr buildVersion else mkNULL
```
quark17, line 500: My earlier comment proposed also checking when `buildVersion` is empty and returning `mkNULL` in that case too.  This is new functionality and I'm happy to add it as a separate commit; but I was wondering if there was a reason that you didn't include it in your commit?

bpfoley: No reason. I just didn't read your comment carefully enough. Done.

PR #249, opened 2020-08-13
Improve provisos in FixedPoint package.

darius-bluespec:
FixedPoint does not support integer bits less than 1.  Add
provisos to enforce this.


quark17:
Thank you, Darius.  Sorry that it took so long to make a decision on this PR.  I'm going to accept these changes.  Part of my fear was not knowing whether this would create new proviso difficulties for anyone using the FixedPoint library.  For anyone reading: if you do experience a problem, please let us know!

PR #250, opened 2020-08-14
Capture if a type is a true enum in the TIdata alternative of TIsort

bpfoley:
This is patch 4/14 in the patchset in #161. Do this to make #161
easier to review.

Signed-off-by: Brian Foley <bpfoley@google.com>

### DIFF
--- a/src/comp/TclUtils.hs
+++ b/src/comp/TclUtils.hs
@@ -190,7 +190,7 @@ instance TclObjCvt TISort where
                             tfs <- toTclObj t
                             toTclObj $ TLst [TStr "alias", TCL tfs]
 
-    toTclObj (TIdata is)  = do
+    toTclObj (TIdata is _)  = do
                             tfs <- mapM toTclObj is
                             toTclObj $ TLst [TStr "enum", TLst (map TCL tfs)]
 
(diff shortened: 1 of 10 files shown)


bpfoley:
I've checked locally that this doesn't cause any testsuite regressions.

src/comp/TclUtils.hs
```diff
@@ -190,7 +190,7 @@ instance TclObjCvt TISort where
                             tfs <- toTclObj t
                             toTclObj $ TLst [TStr "alias", TCL tfs]
 
-    toTclObj (TIdata is)  = do
+    toTclObj (TIdata is _)  = do
                             tfs <- mapM toTclObj is
                             toTclObj $ TLst [TStr "enum", TLst (map TCL tfs)]
```
quark17, line 195: Maybe this is beyond the scope of this PR, but I'm concerned that this function has been reporting `enum` in all cases.  Now that we know when it's a union or not, we might want to report `union` (or `tagged union`?) in some cases.  It's possible that nothing is ever dumping the `TISort`, so it hasn't mattered -- I don't see `enum` in any of the expected output files in `bsc.bluetcl` in the testsuite.

Issue #205, opened 2020-07-15
Poor Code Generation For Non-Consecutive And/Or Non-Zero-Based Enums


jrtc27:
Consider the following code (the methods are `always_enabled` purely to remove the RDY/EN signals that aren't relevant here):

```bsv
`define CONCAT(a, b) a``b

typedef Bit #(4) Type0;
typedef enum { A, B, C, D, E, F, G, H, I } Type1 deriving (Bits, Eq);
typedef enum { A = 1, B, C, D, E, F, G, H, I } Type2 deriving (Bits, Eq);
typedef enum { A, B = 2, C, D, E = 6, F = 8, G, H = 12, I = 15 } Type3 deriving (Bits, Eq);

typedef `CONCAT(Type, `TEST) Type;

interface Test_IFC;
   (* always_enabled *)
   method Action put (Type v);
   (* always_enabled *)
   method Type get;
endinterface

(* synthesize *)
module `CONCAT(mkTest, `TEST) (Test_IFC);
   Reg #(Type) rg_v <- mkRegU;

   method Action put (Type v);
      rg_v <= v;
   endmethod

   method Type get;
      return rg_v;
   endmethod
endmodule
```

I would expect it to generate the same code across all four tests, as they're equivalent, and the values in the Enum aren't relevant unless you actually want to perform comparisons on values. For Test0 and Test1, everything is as expected, and bsc generates the following code for get and put:

```verilog
  // value method get
  assign get = rg_v ;

  // register rg_v
  assign rg_v$D_IN = put_v ;
  assign rg_v$EN = 1'd1 ;

  // handling of inlined registers

  always@(posedge CLK)
  begin
    if (rg_v$EN) rg_v <= `BSV_ASSIGNMENT_DELAY rg_v$D_IN;
  end
```

i.e., the enum is just a way to provide a type-safe set of unique `Bit #(4)` constants.

However, if the enum label values are altered in any way such that they are no longer 0 to n-1 in that order (by permuting that sequence, shifting, or introducing holes), then suddenly bsc decides it wants to enumerate all possible values when variables of that type are involved. For Test2 it gives the following:

```verilog
  // value method get
  always@(rg_v)
  begin
    case (rg_v)
      4'd1, 4'd2, 4'd3, 4'd4, 4'd5, 4'd6, 4'd7, 4'd8: get = rg_v;
      default: get = 4'd9;
    endcase
  end

  // register rg_v
  always@(put_v)
  begin
    case (put_v)
[...]

PR #251, opened 2020-08-14
Add a carefully-constructed fallthrough case to derived pack

bpfoley:
This helps (pack . unpack) to optimize to an identity.

Guarded by -derive-pack-last-clause and False by default.

This is a slightly tweaked and squashed version of patches 13 & 14/14
from the patchset in #161. Do this to make #161 easier to review

### DIFF
--- a/src/comp/Deriving.hs
+++ b/src/comp/Deriving.hs
@@ -4,7 +4,7 @@ module Deriving(derive) where
 import Data.List(nub, intercalate)
 import Util(log2, checkEither, unions, toMaybe, headOrErr, lastOrErr)
 import Error(internalError, EMsg, ErrMsg(..), ErrorHandle, bsError)
-import Flags(Flags)
+import Flags(Flags, derivePackLast)
 import Position
 import Id
 import PreIds(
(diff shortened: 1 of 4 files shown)


bpfoley:
I've checked locally that this doesn't cause any testsuite regressions with `-derive-pack-last-clause` set and unset.

src/comp/Deriving.hs
```diff
@@ -561,11 +561,26 @@ doDBits dpos type_name type_vars original_tags tags =
                 [CClause [CPCon1 type_name
                           (getCISName (headOrErr "doDBits" tags)) (CPVar id_x)] []
                  (cVApply idPack [vx])]
-            | otherwise = zipWith mkPk tags field_bit_sizes
+            | otherwise = zipWith mkPk tags field_bit_sizes ++ [packLast | derivePackLast flags]
```
quark17, line 564: The function `doDBits` is defined in two parts, one definition for when the type is an enum (no data attached to any constructors) and one definition otherwise.  Here, we're only changing the non-enum part;  this same change (adding an explicit default to the derived `pack` function) could be done for enums, on line 460.

Although, this whole PR may prove to be unnecessary.  The enum part of `doDBits` is itself defined in two parts, depending on whether the constructor-to-bit mapping is contiguous or not (and using `PrimChr`/`PrimOrd` to represent the mapping when it is contiguous).  On PR #205 , the idea came up to merge those two into one, using a single primitive for the mapping, whether contiguous or not (and perhaps even redefining `PrimChr`/`PrimOrd` to be that primtive).  Perhaps we could go further and merge both definitions of `doDBits`, into a single definition? For example, the `pack` function could be constructed as `PrimOrd` (for the tag) optionally concatenated with a case-expression (for the payload).

quark17:
This (as with #283 ) has been replaced by a different approach using PrimOrd, that will be filed as a new PR.

PR #252, opened 2020-08-14
Don't lose (if p1 c1 else if p2 c2 else ...) == c optimization

bpfoley:
Patch 5/14 from patchset in #161. Pulled out to make #161 easier to
review.


bpfoley:
Checked locally this doesn't cause any testsuite regressions.

quark17:
This has been replaced by #297 .  (This commit has an extra clause for `ICUndet` with an expression in `imVal`, but Ravi says that undets aren't chosen by BSC that way anymore, which he later discovered -- and he suggests that `imVal` be removed, so that others aren't led astray by it.)

PR #253, opened 2020-08-14
Add support for bit concatenations with matching boundaries to improveIf

bpfoley:
Patch 8/14 from patchset in #161. Pulled out into a separate PR to
make #161 easier to review.


bpfoley:
Checked locally that it doesn't cause any testsuite regressions.

quark17:
This has been replaced by #291

PR #258, opened 2020-08-20
Allow mutually recursive type class instances (fix #257)

### LINKED ISSUE #257: "Recursive dict binds" internal error resulting from mutual instance dependency through a default instance
I encountered this error while working on generics.  The following is a minimal test case that illustrates the error:
```haskell
data Foo = Foo

data C a = C a

class G a r | a -> r where
  f :: a -> r

instance G Foo (C Foo) where
  f = C

class S a where
  s :: a -> Integer

instance (G a r, S' r) => S a where
  s x = s' $ f x

class S' a where
  s' :: a -> Integer

instance (S a) => S' (C a) where
  s' _ = 42

res :: Integer
res = s Foo
```
This results in the error message
```
Internal Bluespec Compiler Error:
Please report this failure to your Bluespec technical contact.
If you do not know your contact, you may email support@bluespec.com.
The following internal error message should be included in your
correspondence along with any other relevant details:
Bluespec Compiler (build 9c9ce37)
recursive dict binds: ([(_tcdict1018,
   RecDict.G RecDict.Foo (RecDict.C RecDict.Foo),
   RecDict.RecDict.G~RecDict.Foo~RecDict.C~RecDict.Foo),
  (_tcdict1011,
   RecDict.S' (RecDict.C RecDict.Foo),
   RecDict.RecDict.S'~RecDict.C~a �_tctyvar1015 �RecDict.Foo _tcdict1017 _tcdict1018),
  (_tcdict1012,
   RecDict.G RecDict.Foo (RecDict.C RecDict.Foo),
   RecDict.RecDict.G~RecDict.Foo~RecDict.C~RecDict.Foo),
  (_tcdict1007,
   RecDict.S RecDict.Foo,
   RecDict.RecDict.S~a �_tctyvar1009 �RecDict.Foo _tcdict1011 _tcdict1012)],
 [],
 [],
 Subst [(_tctyvar1002, Prelude.Integer),
	(_tctyvar1003, RecDict.Foo),
	(_tctyvar1006, RecDict.Foo),
	(_tctyvar1008, ())])
```
Note that the error is avoided by removing the `S a` constraint on the `S'` instance, replacing `a` with `Foo` in the `S` instance, changing the `S` instance to `S' (C a) => S a`, or removing the call to `s Foo` in `res`.  The same code is accepted by Haskell, with the extensions UndecidableInstances, FlexibleInstances, MultiParamTypeClasses, and FunctionalDependencies enabled.  

This error is raised by `checkBinds` in `TCMisc.hs`.  It seems that this check is being overly cautious, because removing it (and also preserving `bs` in the return of `sat` for the "sat recursive" case) allows this test case to compile as expected.

krame505:
This is needed for generics, see #257.  Note that there is a test case in the test suite (`./bsc.typechecker/instances/MutuallyRecursiveInstances.bsv`) that was previously expecting an internal compiler error that now compiles successfully.  This change is further tested by generics (yet to be merged.)

### DIFF
--- a/src/comp/TCMisc.hs
+++ b/src/comp/TCMisc.hs
@@ -296,9 +296,6 @@ sat dvs ps p =
     case lookfor_result of
       Just (bs, (s, [])) ->
          -- tie the recursive knot!
-         -- Unfortunately this causes local mutually recursive
-         -- bindings of dictionaries (EUnsupportedMutualRecursion in IConv),
-         -- so for the simply recursive case, it is cleaned up with cleanUpBinds
          -- traces ("recursive knot: " ++ (ppString p) ++ " = " ++ (ppString lookfor_result))$
          case p of
            (VPred vp (PredWithPositions (IsIn cl tys) poss))
--- a/src/comp/TCMisc.hs
+++ b/src/comp/TCMisc.hs
@@ -20,7 +20,6 @@ import qualified Data.Set as S
 import System.IO.Unsafe(unsafePerformIO)
 
 import ListMap(lookupWithDefault)
-import ListUtil(mapSnd)
 import Util
 import PFPrint
 import Position
@@ -41,7 +40,6 @@ import PreIds
 import StdPrel(isPreClass)
 import CSyntax(CDefl(..), CExpr(..), CPat(..), CQual(..),
                CLiteral(..), cTApply, cVApply, anyTExpr)
-import CFreeVars
 import Literal
 import IntLit
 import SymTab
@@ -136,7 +134,6 @@ satisfy' dvs es ps = do
        -- satTraceM ("satisfy (join)' = " ++ ppString (ps0, bs0, s0))
        (vp, binds, s) <- sMany dvs es [] bs0 s0 ps0
        -- satTraceM ("satisfy (result)' = " ++ ppString (vp,binds,s))
-       checkBinds es vp binds
        return (vp, binds, s)
   where
         -- sMany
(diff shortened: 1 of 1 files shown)


quark17:
Huh.  That was easy enough!  Looking at the commit log (from prior to open BSC), it seems like this code might have originally been making up for limitations elsewhere in BSC, which have since been fixed.

I notice that the comment at line 330/301 still mentions the removed `cleanUpBinds` function.  I would guess that whole comment is obsolete now and should be removed -- the `EUnsupportedMutualRecursion` error that it mentions in `IConv` no longer exists there.  The case-expression there has two arms for cases that are handled specially: recursive Bits instance and recursive solving of numeric types.  I would guess offhand that both of these are still good/necessary, but I'd like to think a little more on it (and experimenting with removing them -- I notice that the recursive Bits error is tested by deriving Bits on a recursive struct in `bsc.typechecker/deriving/RecDeriveStructManyFields.bs`).

This doesn't break anything in the testsuite, except for that one test which now compiles.

I did wonder what the behavior would look like if you accidentally created an infinite loop with recursive instances: I modified `MutuallyRecursiveInstances.bsv` to not have the base case for `x==0` and then called `is_odd` in a synthesized module, and found that BSC gives warnings about unfolding steps and then a stack overflow.  That seems good; so I'll add that as a test case, when this is committed.

krame505:
Is there anything still blocking this PR or can it be merged now?

quark17:
There are two recursive cases that are still specially handled (Bits and numeric typeclasses) and I had wanted to explore those more, in case something could be eased or simplified there.  In any case, that can be done later and need not hold up this PR.

The arm for numeric typeclasses (`isPreClass`) seems reasonable, although I don't know if it can actually be reached.  The commit log says that it was added to avoid an internal error, but I couldn't tell if there was a particular test in the suite for it.   It was added as part of a major upgrade to numeric type handling (using a new `NumEq` class) and a number of tests in `bsc.bugs/b578/` and `bsc.bugs/b675/` were part of that.  (It could be interesting to remove that case arm and run the testsuite, and see if any example breaks.)

For Bits, the compiler reports error T0115, which can be easily searched for in the suite.  My guess is that this case arm is an attempt to catch a user's error earlier, to save the user the headache of discovering the problem later and to perhaps better report the issue.  In `bsc.compile/letrec/` there are examples `DeriveBits{1,2,3,3b}.bsv` and in `bsc.typechecker/deriving/` there is `RecDeriveStructManyFields.bsv`.

PR #260, opened 2020-08-25
Enable building on FreeBSD.

davidchisnall:
Some cleanup to pull generic things (e.g. finding the TCL installation)
to the top level.  Add missing cases.

The existing build system conflates of 'has BSD tools' vs 'has GNU tools',
'is ELF' vs 'is Mach-O' and 'is Linux' vs 'is Darwin'.  I have not tried
to clean up that conflation, only to add the missing case.

### DIFF
--- a/platform.mk
+++ b/platform.mk
@@ -29,7 +29,7 @@ endif
 OSTYPE = $(shell $(TOP)/platform.sh ostype)
 export OSTYPE
 
-ifneq ($(OSTYPE), $(findstring $(OSTYPE), Linux Darwin))
+ifneq ($(OSTYPE), $(findstring $(OSTYPE), Linux Darwin Freebsd))
 $(error OSTYPE environment not recognized: $(OSTYPE))
 endif
 
@@ -56,3 +56,7 @@ export CFLAGS
 export CXXFLAGS
 
 ## --------------------
+## Set up the TCL shell and include paths
+TCLSH = $(shell $(TOP)/platform.sh tclsh)
+TCL_INC_FLAGS = $(shell $(TOP)/platform.sh tclinc)
+TCL_LIB_FLAGS = $(shell $(TOP)/platform.sh tcllibs)
--- a/platform.sh
+++ b/platform.sh
@@ -8,6 +8,9 @@ usage()
     echo "Valid options are:"
     echo "   ostype   "
     echo "   machtype "
+    echo "   tclsh "
+    echo "   tclinc "
+    echo "   tcllibs "
     echo "   c++_shared_flags"
 }
 
@@ -44,6 +47,12 @@ fi
 ## "x86_64", "i386", etc
 MACHTYPE=$(echo ${MACHTYPE} | cut -d'-' -f1)
 
+# If we see a BSD-style amd64 instead of x86_64, rewrite it to the more generic
+# version.
+if [ ${MACHTYPE} = "amd64" ] ; then
+    MACHTYPE=x86_64
+fi
+
 if [ "$1" = "machtype" ] ; then
     echo ${MACHTYPE}
     exit 0
@@ -63,6 +72,88 @@ if [ "$1" = "c++_shared_flags" ] ; then
 fi
 
 ## =========================
+## Find the TCL shell command
+TCL_SUFFIX=
+if [ ${OSTYPE} = "Darwin" ] ; then
+    # Have Makefile avoid Homebrew's install of tcl on Mac
+    TCLSH=/usr/bin/tclsh
+elif [ ${OSTYPE} = Freebsd ] ; then
+    # The FreeBSD tcl packages are versioned.  Find the right one.
+    if [ -n "`which tclsh8.7`" ] ; then
+        TCL_SUFFIX=8.7
+        TCL_ALT_SUFFIX=87
+    elif [ -n "`which tclsh8.6`" ] ; then
+        TCL_SUFFIX=8.6
+        TCL_ALT_SUFFIX=86
+    elif [ -n "`which tclsh8.5`" ] ; then
+        TCL_SUFFIX=8.5
+        TCL_ALT_SUFFIX=85
+    fi
+    TCLSH=`which tclsh${TCL_SUFFIX}`
+else
+    TCLSH=`which tclsh`
+fi
+
+if [ "$1" = "tclsh" ] ; then
+    echo ${TCLSH}
+    exit 0
+fi
+
+if [ "$1" = "tclinc" ] ; then
+    # Try pkg-config
+    TCL_INC_FLAGS=`pkg-config --silence-errors --cflags-only-I tcl${TCL_SUFFIX}`
+    # If pkg-config didn't work with the first prefix, try the alternative version.
+    # For example, on FreeBSD, the tcl87 package installs tclsh8.7, but tcl87.pc
+    if [ -z "${TCL_INC_FLAGS}" ] ; then
+        TCL_INC_FLAGS=`pkg-config --silence-errors --cflags-only-I tcl${TCL_ALT_SUFFIX}`
+    fi
+    # If pkg-config doesn't work, try some well-known locations
+    if [ -z "${TCL_INC_FLAGS}" ] ; then
+        if [ -f "/usr/local/include/tcl${TCL_SUFFIX}/tcl.h" ] ; then
+            TCL_INC_FLAGS="/usr/local/include/tcl${TCL_SUFFIX}"
+        elif [ -f "/usr/include/tcl${TCL_SUFFIX}/tcl.h" ] ; then
+            TCL_INC_FLAGS="/usr/include/tcl${TCL_SUFFIX}"
+        else
+            exit 1
+        fi
+    fi
+    echo ${TCL_INC_FLAGS}
+    exit 0
+fi
+
+if [ ${OSTYPE} == "Darwin" ] ; then
+    LIB_SUFFIX=dylib
+else
+    LIB_SUFFIX=so
+fi
+
+if [ "$1" = "tcllibs" ] ; then
+    # Try pkg-config
+    TCL_LIB_FLAGS=`pkg-config --silence-errors --libs tcl${TCL_SUFFIX}`
+    # If pkg-config didn't work with the first prefix, try the alternative version.
+    # For example, on FreeBSD, the tcl87 package installs tclsh8.7, but tcl87.pc
+    if [ -z "${TCL_LIB_FLAGS}" ] ; then
+        TCL_LIB_FLAGS=`pkg-config --silence-errors --libs tcl${TCL_ALT_SUFFIX}`
+    fi
+
+    if [ -n "${TCL_LIB_FLAGS}" ] ; then
+        echo ${TCL_LIB_FLAGS}
+        exit 0
+    fi
+
+    # If pkg-config doesn't work, try some well-known locations
+    TCL_VER=`echo 'catch { puts [info tclversion]; exit 0}; exit 1' | ${TCLSH}`
+    for L in /usr/lib /usr/local/lib ; do
+        for V in ${TCL_VER} ${TCL_SUFFIX} ${TCL_ALT_SUFFIX} ; do
+            if [ -f "${L}/libtcl${V}.${LIB_SUFFIX}" ] ; then
+                echo -L${L} -ltcl${V}
+                exit 0
+            fi
+        done
+    done
+    exit 1
+fi
+
 
 usage
 exit 1
--- a/src/Makefile
+++ b/src/Makefile
@@ -8,13 +8,11 @@ PREFIX?=$(TOP)/inst
 ifndef NO_DEPS_CHECKS
 CC_TOOLS=$(CC) $(CXX) $(LD)
 BSC_TOOLS=ghc ghc-pkg
-BLUETCL_TOOLS=tclsh
 YICES_TOOLS=gperf autoconf
 STP_TOOLS=flex bison
 
 TOOLS=$(CC_TOOLS) \
       $(BSC_TOOLS) \
-      $(BLUETCL_TOOLS) \
       $(YICES_TOOLS) \
       $(STP_TOOLS) \
 
@@ -47,7 +45,7 @@ all: install
 install clean full_clean:
 	$(MAKE)  -C vendor/stp   PREFIX=$(PREFIX)  $@
 	$(MAKE)  -C vendor/yices PREFIX=$(PREFIX)  $@
-	$(MAKE)  -C vendor/htcl  PREFIX=$(PREFIX)  $@
+	$(MAKE)  -C vendor/htcl  PREFIX=$(PREFIX) TCL_INC_FLAGS=$(TCL_INC_FLAGS) $@
 	# we need to build targets from here sequentially, as they operate in the same workspace
 	$(MAKE)  -C comp -j1   PREFIX=$(PREFIX)  $@
 	$(MAKE)  -C Libraries  PREFIX=$(PREFIX)  $@
--- a/src/bluesim/Makefile
+++ b/src/bluesim/Makefile
@@ -38,6 +38,7 @@ HEADERS = bluesim_kernel_api.h \
 
 # These ld export maps get copies to the inst/lib/Bluesim area
 LINKFILES = bs_linux_export_map.txt \
+            bs_freebsd_export_map.txt \
             bs_darwin_export_map.txt
 
 # -------------------------
--- a/src/bluetcl/Makefile
+++ b/src/bluetcl/Makefile
@@ -4,6 +4,8 @@ TOP:=$(PWD)/../..
 PREFIX?=$(TOP)/inst
 INSTALLDIR = $(PREFIX)/lib/tcllib/bluespec
 
+TCLSH = $(shell $(TOP)/platform.sh tclsh)
+
 FILES = \
 	tclIndex \
 	pkgIndex.tcl \
@@ -29,7 +31,7 @@ all: tclIndex
 # pkg_mkIndex.tcl is produced here
 .PHONY: tclIndex
 tclIndex: *.tcl
-	./tclIndex.sh "$(TCLFILES)" "$(PACKAGES)"
+	TCLSH=${TCLSH} ./tclIndex.sh "$(TCLFILES)" "$(PACKAGES)"
 
 .PHONY: install
 install: tclIndex
--- a/src/bluetcl/tclIndex.sh
+++ b/src/bluetcl/tclIndex.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 #\
-exec tclsh "$0" "$@"
+exec ${TCLSH} "$0" "$@"
 
 set tclFiles [lindex $argv 0]
 set packFiles [lindex $argv 1]
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -61,22 +61,10 @@ HTCL_INC_FLAGS = -L$(HTCL_HS)
 HTCL_LIB_FLAGS = -lhtcl
 
 # TCL version
-ifeq ($(OSTYPE), Darwin)
-# Have Makefile avoid Homebrew's install of tcl on Mac
-TCLSH=/usr/bin/tclsh
-else
-TCLSH=tclsh
-endif
 TCL_VER = $(shell echo 'catch { puts [info tclversion]; exit 0}; exit 1' | $(TCLSH) || echo fail)
 ifeq ($(TCL_VER),fail)
 $(error TCL version check failed)
 endif
-
-# TCL include flags
-TCL_INC_FLAGS = $(shell pkg-config --silence-errors --cflags-only-I tcl || echo -I/usr/include/tcl)
-
-# TCL library flags
-TCL_LIB_FLAGS = -ltcl$(TCL_VER)
 
 # -----
 # GHC
--- a/src/comp/bsc.hs
+++ b/src/comp/bsc.hs
@@ -1787,6 +1787,9 @@ cxxLink errh flags toplevel names creation_time = do
             Darwin -> ["-dynamiclib", "-fPIC"] ++ libdirflags ++
                       ["-exported_symbols_list", exportmap] ++ stripflags ++
                       ["-o", soFile]
+            Freebsd -> ["-shared", "-fPIC", "-Wl,-Bsymbolic"] ++ libdirflags ++
+                       ["-Wl,--version-script=" ++ exportmap] ++ stripflags ++
+                       ["-o", soFile]
         -- show is used for quoting
         opts = map show $ linkFlags flags
         files = map show compile_names ++ ["-lm"] ++ userlibs
@@ -1823,8 +1826,9 @@ cxxLink errh flags toplevel names creation_time = do
 cleanseSharedLib :: ErrorHandle -> Flags -> String -> IO ()
 cleanseSharedLib errh flags soFile = do
     let switches = case getOSType of
-                      Linux  -> ["-x"]
-                      Darwin -> ["-u", "-x"]
+                      Linux   -> ["-x"]
+                      Freebsd -> ["-x"]
+                      Darwin  -> ["-u", "-x"]
         cmd = unwords $ ["strip"] ++ switches ++ [soFile]
     when (verbose flags) $ putStrLnF ("exec: " ++ cmd)
     rc <- system cmd
--- a/src/comp/update-build-system.sh
+++ b/src/comp/update-build-system.sh
@@ -9,12 +9,13 @@ OSTYPE=`${PLATFORM_SH} ostype`
 echo "module BuildSystem (OSType(..), osToString, getOSType) where" > BuildSystem.hs.new;
 echo "" >> BuildSystem.hs.new;
 
-echo "data OSType = Linux | Darwin" >> BuildSystem.hs.new;
+echo "data OSType = Linux | Darwin | Freebsd" >> BuildSystem.hs.new;
 echo "" >> BuildSystem.hs.new;
 
 echo "osToString :: OSType -> String" >> BuildSystem.hs.new;
 echo "osToString Linux  = \"Linux\"" >> BuildSystem.hs.new;
 echo "osToString Darwin = \"Darwin\"" >> BuildSystem.hs.new;
+echo "osToString Freebsd = \"Freebsd\"" >> BuildSystem.hs.new;
 echo "" >> BuildSystem.hs.new;
 
 echo "getOSType :: OSType" >> BuildSystem.hs.new;
--- a/src/vendor/htcl/Makefile
+++ b/src/vendor/htcl/Makefile
@@ -1,4 +1,4 @@
-GHCFLAGS += -Wall $(shell pkg-config --silence-errors --cflags-only-I tcl || echo -I/usr/include/tcl)
+GHCFLAGS += -Wall ${TCL_INC_FLAGS}
 GHC ?= ghc
 
 # We use GHC to compile this, so it has the proper RTS includes
--- a/src/vendor/stp/src/Makefile
+++ b/src/vendor/stp/src/Makefile
@@ -8,14 +8,18 @@ INCLUDE_DIR=$(STPDIR)/include
 HEADERS=c_interface/*.h
 
 # BSD cp does not have the -d flag
+ifneq ($(OSTYPE), Linux)
+CP = cp -Rf
+else
+CP = cp -df
+endif
+
 ifeq ($(OSTYPE), Darwin)
 LIBRARIES=lib/libstp.dylib
 SNAME = libstp.dylib
-CP = cp -Rf
 else
 LIBRARIES=lib/libstp.so.1 lib/libstp.so
 SNAME = libstp.so.1
-CP = cp -df
 endif
 
 .PHONY: all
--- a/src/vendor/yices/v2.6/stub/Makefile
+++ b/src/vendor/yices/v2.6/stub/Makefile
@@ -40,7 +40,7 @@ AR	= ar -r
 RM	= rm -rf
 
 # BSD cp does not have the -d flag
-ifeq ($(OSTYPE), Darwin)
+ifneq ($(OSTYPE), Linux)
 CP	= cp -Rf
 else
 CP	= cp -df


davidchisnall:
The first Ubuntu build failed to check out the code, the second and third were cancelled for some reason.  Not sure why...

davidchisnall:
The Ubuntu CI failure is very odd.  bsc is complaining that a directory does not exist, but the line immediately before this is creating that directory.  This looks like a problem unrelated to this PR.  I have locally validated that I can build on Ubuntu 20.04.

bpfoley:
> The Ubuntu CI failure is very odd. bsc is complaining that a directory does not exist, but the line immediately before this is creating that directory. This looks like a problem unrelated to this PR.

I agree. I've seen intermittent CI failures like this in the past and I've never been able to track down what was going on. It feels like the sort of problem you'd get with a networked filesystem, but I don't think the CI setups are using that for the build machines.

Thanks for the PR, by the way. I've been busy and on vacation, so I haven't had a chance to examine it yet, but I'm looking forward to reviewing it soon unless Julie gets there first.

quark17 (commented):
This is great!  I'm glad to see it compiling on FreeBSD!

I also like how the code for deciding the Tcl flags is consolidated into the platform script -- both to avoid cluttering the makefiles (with the growing code) and to avoid duplicating the code across the multiple makefiles that use it.

I've given some thoughts below, but I'd feel better if @bpfoley can review this.

quark17:
The CI error looks like a parallel make problem.  In `src/Libraries/Base1/Makefile`, there's a target to make the directory, and the only attempt to make it happen first is by listing it first in the `build` targets -- it's not specified as a dependency of the targets that actually depend on it.
```
build: $(BUILDDIR) $(OBJS)

$(BUILDDIR):
        mkdir -p $@
```
I'd guess that the executions of `mkdir` and `bsc` are happening at the same time, even though `make` has reported the commands in an order.  Note, though, that these commands are intermingled in the lines of of the error message from BSC!  The error from BSC starts with `Error: Command line: (S0092)` (on line 9438) after which we see the two commands and then we see the rest of the error message.  This seems to confirm that the printout of the commands does not reflect the execution time of the commands.

platform.mk
```diff
@@ -56,3 +56,19 @@ export CFLAGS
 export CXXFLAGS
 
 ## --------------------
+## Set up the TCL shell and include paths
+TCLSH = $(shell $(TOP)/platform.sh tclsh)
+ifeq ($(TCLSH), )
+$(error Unable to find tclsh)
```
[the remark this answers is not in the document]
davidchisnall: I can make this conditional by making things that want this info set a `WANT_TCL` variable or similar before including `platform.mk`.  That would probably also speed up builds slightly: getting the TCL variables takes a second or so on my machine, so skipping this for things that don't need TCL makes sense.

Ideally, I'd replace all of this goo with CMake and just use the FindTCL package to get all of these things in a completely portable way, but completely rewriting the build system is probably not something I have time for soon (though if it were likely to be accepted upstream, I might get to it eventually).

src/Makefile
```diff
@@ -8,13 +8,11 @@ PREFIX?=$(TOP)/inst
 ifndef NO_DEPS_CHECKS
 CC_TOOLS=$(CC) $(CXX) $(LD)
 BSC_TOOLS=ghc ghc-pkg
-BLUETCL_TOOLS=tclsh
```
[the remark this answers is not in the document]
davidchisnall: Yes, exactly.

src/Makefile
```diff
@@ -47,7 +45,7 @@ all: install
 install clean full_clean:
 	$(MAKE)  -C vendor/stp   PREFIX=$(PREFIX)  $@
 	$(MAKE)  -C vendor/yices PREFIX=$(PREFIX)  $@
-	$(MAKE)  -C vendor/htcl  PREFIX=$(PREFIX)  $@
+	$(MAKE)  -C vendor/htcl  PREFIX=$(PREFIX) TCL_INC_FLAGS=$(TCL_INC_FLAGS) $@
```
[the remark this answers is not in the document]
davidchisnall: I didn't want to add an include in places where I didn't udnerstand why they didn't include it already, but if it's not a problem to do so then I agree that would be cleaner.

src/bluesim/Makefile
```diff
@@ -38,6 +38,7 @@ HEADERS = bluesim_kernel_api.h \
 
 # These ld export maps get copies to the inst/lib/Bluesim area
 LINKFILES = bs_linux_export_map.txt \
+            bs_freebsd_export_map.txt \
```
davidchisnall, line 41: I don't really like this.  I think the right thing to do here is to have a `bs_elf_export_map.txt` and use that for Linux and FreeBSD.

src/bluetcl/Makefile
```diff
@@ -4,6 +4,8 @@ TOP:=$(PWD)/../..
 PREFIX?=$(TOP)/inst
 INSTALLDIR = $(PREFIX)/lib/tcllib/bluespec
 
+TCLSH = $(shell $(TOP)/platform.sh tclsh)
```
davidchisnall, line 7: Is there a reason that this file doesn't include `platform.mk` and get this variable?

src/bluetcl/Makefile
```diff
@@ -29,7 +31,7 @@ all: tclIndex
 # pkg_mkIndex.tcl is produced here
 .PHONY: tclIndex
 tclIndex: *.tcl
-	./tclIndex.sh "$(TCLFILES)" "$(PACKAGES)"
+	TCLSH=${TCLSH} ./tclIndex.sh "$(TCLFILES)" "$(PACKAGES)"
```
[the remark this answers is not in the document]
davidchisnall: I don't understand why this script is not already invoked with TCLSH, but I'm happy to make the change.

src/bluetcl/tclIndex.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 #\
-exec tclsh "$0" "$@"
+exec ${TCLSH} "$0" "$@"
```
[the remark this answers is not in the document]
davidchisnall: Removing the first four lines of this file and renaming it to `tclIndex.tcl` probably makes the most sense.

src/comp/bsc.hs
```diff
@@ -1787,6 +1787,9 @@ cxxLink errh flags toplevel names creation_time = do
             Darwin -> ["-dynamiclib", "-fPIC"] ++ libdirflags ++
                       ["-exported_symbols_list", exportmap] ++ stripflags ++
                       ["-o", soFile]
+            Freebsd -> ["-shared", "-fPIC", "-Wl,-Bsymbolic"] ++ libdirflags ++
```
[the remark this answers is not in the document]
davidchisnall: This is one of the cases of doing things that are binary format dependent but triggering them based on the OS name.  Properly cleaning this up will involve doing a map from OS to binary format and then keying this on the binary format.  I'll have a go at this, though the last time I wrote any Haskell was around 15 years ago, so we'll see how it goes.

davidchisnall (commented):
Thanks for the feedback!  I've left a few questions, but I'll aim to address the flagged issues in the next few days.

davidchisnall:
The latest version addresses reviewer feedback and passes CI.  Cirrus CI offers free CI on FreeBSD, if you want to stop it breaking in the future.

bpfoley:
Hi David, this looks excellent thanks. I tried to reproduce this build on a vanilla FreeBSD 12.1 x86_64 virtual machine and I ran into a couple of minor issues that I had to work around, but I'd like to get this PR merged, so I'll deal with those in some followup pull requests.

Julie, I'm happy to merge this as is if you are.

In particular, I installed GHC (8.8) from FreeBSD ports, and Yices's autoconf had problems finding libgmp in /usr/local, and I ran across some problems with the Yices stub library that I'll have to fix up. Do you perhaps have local patches against the yices submodule your local checkout?

I'll also add some FreeBSD notes to the README detailing the packages and setup required to build.

platform.mk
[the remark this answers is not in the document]
quark17: Sorry for not responding to these before you did more work!

Yes, there is interest in replacing all this with something better.  I don't know if we've decided what that is, though (whether it's CMake or something else).

As for the `WANT_TCL` mechanism: I'm happy to know that a simple fix can be implemented if we need it; for now, I would tend to keep things simple and leave it out.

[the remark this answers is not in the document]
quark17: The `platform.mk` calls the shell script multiple times, to get each value.  For the Tcl variables, there's three calls (and each time, the script recomputes `TCLSH`).  I suspect that the additional run time is because of this.  If `platform.mk` could run the shell once, grabbing assignments for everything and then then just parsing out what it needs, that might be a better solution to run time than guarding with `WANT_TCL`?

src/bluetcl/Makefile
quark17: You added this line -- are you asking if it's OK to include `platform.mk` instead?  If only one variable is needed, it's cheaper to just call the script (since the makefile will not only do that, but also call the script to populate other variables that aren't needed).  For OSTYPE and MACHTYPE, `platform.mk` does check that the script returns an expected value, so it's best not to call the script directly for those, I guess. (The makefile also sets CFLAGS and CXXFLAGS itself, not as part of the script. I think that was only there for historical reasons, to support cross-compiling, and can be removed.)  It probably makes sense that makefiles should always import `platform.mk` instead of calling the script directly, but if the efficiency is a concern (as we add more variables) then I have no objection to calling the script directly in a situation like this, where a single variable is needed and it's not a variable that `platform.mk` does further operating on.

src/bluetcl/tclIndex.sh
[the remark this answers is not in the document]
quark17: Yes, we should probably change this to a tcl script and remove the first few lines.

src/bluetcl/Makefile
```diff
@@ -29,7 +31,7 @@ all: tclIndex
 # pkg_mkIndex.tcl is produced here
 .PHONY: tclIndex
 tclIndex: *.tcl
-	./tclIndex.sh "$(TCLFILES)" "$(PACKAGES)"
+	${TCLSH} ./tclIndex.sh "$(TCLFILES)" "$(PACKAGES)"
```
quark17, line 34: The file should be renamed from `.sh` to `.tcl`

src/comp/Makefile
```diff
@@ -1,5 +1,6 @@
 PWD:=$(shell pwd)
 TOP:=$(PWD)/../..
+WANT_TCL=yes
```
quark17, line 3: I would group this with the `platform.mk` include (that is, move it down a line).

quark17 (commented):
Just some tiny fixes, but it looks good.  I also think that `platform.mk` could be improved (as mentioned in earlier comments), but that could be done separately.

quark17:
I've also installed the Cirrus CI app on the bsc* repos and will look into setting up a FreeBSD CI here.

quark17:
Actually, @davidchisnall , if you have any experience writing `.cirrus.yml` files and want to include one in this PR, that would be welcome.  It may take me some time to figure this out (and I don't see GHC/Haskell among its language examples, which hopefully isn't a bad sign).

src/bluetcl/Makefile
davidchisnall: I didn't make the change because this script collects the names of everything with a .tcl extension in the directory and I didn't want it to collect itself.

davidchisnall:
> In particular, I installed GHC (8.8) from FreeBSD ports, and Yices's autoconf had problems finding libgmp in /usr/local, and I ran across some problems with the Yices stub library that I'll have to fix up. Do you perhaps have local patches against the yices submodule your local checkout?

Yes, sorry, I have this:

```diff
diff --git a/Makefile.build b/Makefile.build
index 0a5b000e..87cac7b4 100644
--- a/Makefile.build
+++ b/Makefile.build
@@ -483,7 +483,7 @@ install-linux install-unix: install-default
 # be added to the hints file.' In other words, ldconfig on FreeBSD doesn't create the symbolic link,
 # as on Linux.
 install-freebsd: install-default
-       $(LDCONFIG) -m $(DESTDIR)$(libdir) && (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(MAJOR).$(MINOR) libyices.so)
+       (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(MAJOR) libyices.so)

 install-netbsd: install-default
        (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(YICES_VERSION) libyices.so)
```

What is your preferred way of doing a submodule PR?

src/comp/Makefile
```diff
@@ -1,5 +1,6 @@
...
 
 include $(TOP)/platform.mk
```
davidchisnall, line 5:
```suggestion

WANT_TCL=yes
include $(TOP)/platform.mk
```

I put this above because I thought it was the same category of thing as `$(TOP)`, which is also used by `platform.mk`.

bpfoley:
Could we rename the top level Makefile to GNUmakefile, and replace the Makefile with
```
default:
        @echo "BSC requires GNU Make, run gmake instead"
        @false
```

BSD make is really not happy with all the GNU-isms we rely on heavily, and fails in a way that's probably not helpful to newbies. Unlike BSD make, GNU Make will look for a `GNUmakefile` before a `Makefile`, thus giving us a way to select between the two.

bpfoley:
Once I get it to build without problems, I'll update the README.md with instructions for FreeBSD. Just a reminder for myself:

```
Requires 

pkg install \
    gmake \
    ghc \
    gperf autoconf \
    flex bison \
    hs-cabal-install \
cabal v1-install regex-compat syb old-time split

pkg install \
    tcl87 \
    pkg-conf
```

bpfoley:
Also, could you apply this diff please? Otherwise building with gmake YICES_STUB=1 fails
```
--- a/src/vendor/yices/v2.6/Makefile
+++ b/src/vendor/yices/v2.6/Makefile
@@ -37,7 +37,7 @@ ifeq ($(YICES_STUB),)
                $(MAKE) install \
                )
 else
-       (cd stub ; make install)
+       $(MAKE) -C stub install
 endif
        ln -fsn $(YICES_INST)/include
        ln -fsn $(YICES_INST)/lib
```

bpfoley:
> > In particular, I installed GHC (8.8) from FreeBSD ports, and Yices's autoconf had problems finding libgmp in /usr/local, and I ran across some problems with the Yices stub library that I'll have to fix up. Do you perhaps have local patches against the yices submodule your local checkout?
> 
> Yes, sorry, I have this:
> 
> ```diff
> diff --git a/Makefile.build b/Makefile.build
> index 0a5b000e..87cac7b4 100644
> --- a/Makefile.build
> +++ b/Makefile.build
> @@ -483,7 +483,7 @@ install-linux install-unix: install-default
>  # be added to the hints file.' In other words, ldconfig on FreeBSD doesn't create the symbolic link,
>  # as on Linux.
>  install-freebsd: install-default
> -       $(LDCONFIG) -m $(DESTDIR)$(libdir) && (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(MAJOR).$(MINOR) libyices.so)
> +       (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(MAJOR) libyices.so)
> 
>  install-netbsd: install-default
>         (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(YICES_VERSION) libyices.so)
> ```
> 
> What is your preferred way of doing a submodule PR?

Could you submit a PR to https://github.com/SRI-CSL/yices2 if they don't already have this fix, and once it's submitted, we can update which hash we're importing from the yices2 repo.

bpfoley:
> > In particular, I installed GHC (8.8) from FreeBSD ports, and Yices's autoconf had problems finding libgmp in /usr/local, and I ran across some problems with the Yices stub library that I'll have to fix up. Do you perhaps have local patches against the yices submodule your local checkout?


I'm also seeing a bunch of liker errors like
```
checking for libgmp.a in /usr/local/lib... found
checking whether /usr/local/lib/libgmp.a is usable... no
checking for libgmp.a in /usr/lib... no
checking for libgmp.a in /lib... no
configure: WARNING: *** No usable libgmp.a library was found ***
```
[...]
```
ranlib ../build/x86_64-unknown-freebsd12.1-release/lib/libyices.a
cc -m64 -I/usr/local/include -fvisibility=hidden  -Wall -Wredundant-decls -O3 -fomit-frame-pointer -fno-stack-protector -L/usr/local/lib  \
   -o ../build/x86_64-unknown-freebsd12.1-release/bin/yices ../build/x86_64-unknown-freebsd12.1-release/obj/frontend/yices.o ../build/x86_64-unknown-freebsd12.1-release/lib/libyices.a -lm
ld: error: undefined symbol: __gmp_version
>>> referenced by yices_reval.c
>>>               yices_reval.o:(yices_main) in archive ../build/x86_64-unknown-freebsd12.1-release/lib/libyices.a
```

Did you have to jump through any hoops to get your build to find GMP in /usr/local?

davidchisnall:
> Did you have to jump through any hoops to get your build to find GMP in /usr/local?

No, I'm not sure why that worked for me.  The `-L/usr/local/lib` flag is telling it to search in the right place, but it seems to be missing a `-lgmp`?

jinwoo:
> Did you have to jump through any hoops to get your build to find GMP in /usr/local?

FYI. In our local build, I have a patch like below:

```
@@ -32,7 +32,9 @@ install:
 ifeq ($(YICES_STUB),)
 	(cd $(YICES_SRC) ; \
 		autoconf ; \
-		./configure --prefix=$(YICES_INST) ; \
+		./configure --prefix=$(YICES_INST) \
+		--with-pic-gmp=$(GMP_LIB) \
+		--with-pic-gmp-include-dir=$(GMP_INCLUDE_DIR) ; \
 		$(MAKE) $(BUILD_ARGS); \
 		$(MAKE) install \
 		)
```

I.e., I pass `--with-pic-gmp` and `--with-pic-gmp-include-dir` explicitly. It may help this case too.

bpfoley:
> > Did you have to jump through any hoops to get your build to find GMP in /usr/local?
> 
> FYI. In our local build, I have a patch like below:
> 
> ```
> @@ -32,7 +32,9 @@ install:
>  ifeq ($(YICES_STUB),)
>  	(cd $(YICES_SRC) ; \
>  		autoconf ; \
> -		./configure --prefix=$(YICES_INST) ; \
> +		./configure --prefix=$(YICES_INST) \
> +		--with-pic-gmp=$(GMP_LIB) \
> +		--with-pic-gmp-include-dir=$(GMP_INCLUDE_DIR) ; \
>  		$(MAKE) $(BUILD_ARGS); \
>  		$(MAKE) install \
>  		)
> ```
> 
> I.e., I pass `--with-pic-gmp` and `--with-pic-gmp-include-dir` explicitly. It may help this case too.

Unfortunately it didn't, but I was able to get it working with the following. @davidchisnall could you include this please?

--- a/src/vendor/yices/v2.6/Makefile
+++ b/src/vendor/yices/v2.6/Makefile
@@ -8,6 +8,9 @@ REPONAME=yices2
 YICES_SRC=$(REPONAME)
 YICES_INST=$(abspath $(PWD)/$(REPONAME)-inst)

+GMP_LIB_DIR ?= $(shell pkg-config --variable libdir gmp)
+GMP_INCLUDE_DIR ?= $(shell pkg-config --variable includedir gmp)
+
 ifeq ($(OSTYPE), Darwin)
 # On macOS we need to work around two yices build issues:
 # 1. yices explicitly sets the install_name for libyices.2.dylib to be
@@ -20,8 +23,16 @@ ifeq ($(OSTYPE), Darwin)
 #    depend on Homebrew. Instead, statically link libgmp into libyices.
 #    Note that we must statically link by using the archive file as Apple's ld
 #    doesn't understand -Bstatic.
-LIBGMPA=$(shell pkg-config --variable=libdir gmp)/libgmp.a
-BUILD_ARGS=libyices_install_name=libyices.2.dylib LIBS=$(LIBGMPA)
+BUILD_ARGS=libyices_install_name=libyices.2.dylib LIBS=$(GMP_LIB_DIR)/libgmp.a
+endif
+
+ifeq ($(OSTYPE), Freebsd)
+# On FreeBSD we need to work around the static vs dynamic libgmp detection
+# which gets confused by libgmp.h not being on the standard include path.
+# Using ./configure --with-pic-gmp --with-pic-gmp-include-dir fails in other
+# ways, so set CFLAGS by hand.
+LDFLAGS+=-L$(GMP_LIB_DIR)
+CFLAGS+=-I$(GMP_INCLUDE_DIR)
 endif

davidchisnall:
Sorry for the delay.  I had to take last week off to handle some personal issues, I hope to get back to this soon.

platform.sh
```diff
@@ -63,6 +72,103 @@ if [ "$1" = "c++_shared_flags" ] ; then
 fi
 
 ## =========================
+## Find the TCL shell command
+TCL_SUFFIX=
+if [ ${OSTYPE} = "Darwin" ] ; then
+    # Have Makefile avoid Homebrew's install of tcl on Mac
+    TCLSH=/usr/bin/tclsh
+elif [ ${OSTYPE} = Freebsd ] ; then
```
jrtc27, line 80: The use of quotes is inconsistent

platform.sh
```diff
@@ -63,6 +72,103 @@ if [ "$1" = "c++_shared_flags" ] ; then
...
+    # The FreeBSD tcl packages are versioned.  Find the right one.
```
jrtc27, line 81: This sounds like a deficiency in ports to me. Should it not be like perl where they're versioned but the PERL5_DEFAULT version isn't (well, with a symlink so both work)? Having to maintain this list here is ugly.

platform.sh
```diff
@@ -63,6 +72,103 @@ if [ "$1" = "c++_shared_flags" ] ; then
...
+fi
+
+
+if [ "$1" = "tclinc" ] ; then
+    # Try pkg-config
+    TCL_INC_FLAGS=`${PKG_CONFIG} --silence-errors --cflags-only-I tcl${TCL_SUFFIX}`
+    # If pkg-config didn't work with the first prefix, try the alternative version.
+    # For example, on FreeBSD, the tcl87 package installs tclsh8.7, but tcl87.pc
+    if [ -z "${TCL_INC_FLAGS}" ] ; then
+        TCL_INC_FLAGS=`${PKG_CONFIG} --silence-errors --cflags-only-I tcl${TCL_ALT_SUFFIX}`
```
jrtc27, line 114: TCL_ALT_SUFFIX isn't always set by this point (it's a shame there's no `set -u` to scream in such cases.. similarly no `set -e` makes me sad).

src/comp/bsc.hs
```diff
@@ -1822,9 +1822,9 @@ cxxLink errh flags toplevel names creation_time = do
 -- strip unwanted symbols from a .so file
 cleanseSharedLib :: ErrorHandle -> Flags -> String -> IO ()
 cleanseSharedLib errh flags soFile = do
-    let switches = case getOSType of
-                      Linux  -> ["-x"]
-                      Darwin -> ["-u", "-x"]
+    let switches = case getBinFmtType of
+                      ELF   -> ["-x"]
+                      MachO  -> ["-u", "-x"]
```
jrtc27, line 1827: Spacing is off

src/comp/bsc.hs
```diff
@@ -1774,17 +1774,17 @@ cxxLink errh flags toplevel names creation_time = do
...
-          case getOSType of
-            Linux -> ["-shared", "-fPIC", "-Wl,-Bsymbolic"] ++ libdirflags ++
+          case getBinFmtType of
+            ELF -> ["-shared", "-fPIC", "-Wl,-Bsymbolic"] ++ libdirflags ++
                      ["-Wl,--version-script=" ++ exportmap] ++ stripflags ++
                      ["-o", soFile]
-            Darwin -> ["-dynamiclib", "-fPIC"] ++ libdirflags ++
+            MachO -> ["-dynamiclib", "-fPIC"] ++ libdirflags ++
                       ["-exported_symbols_list", exportmap] ++ stripflags ++
                       ["-o", soFile]
```
jrtc27, line 1789: Spacing is off

platform.sh
hliebel: I installed `tcl-wrapper` with the `pkg` installer tool. It hides the tcl version info and provides a `tclsh` to use..

davidchisnall:
I have a bit of time to work on this again.  It looks as if FreeBSD no longer maintains GHC packages in the ports system and instead recommends using cabal or stack to build something with local copies of all dependencies.

davidchisnall:
I've pushed some changes, but moving the build for the Haskell bits to using Cabal rather than relying on system-wide package installation is not something I know how to do.

src/bluetcl/Makefile
quark17: This PR has removed the initial lines from `tclIndex.sh`, so it is no longer a shell script, and we should probably change the name to not have `.sh` as an extension.  I'm OK renaming it to `build_tclIndex` with no extension, or to give an extension like `.script` or `.tcl_script`.

We could even give it the extension `.tcl` ... the reason we can't do that is because its dependency is `*.tcl`, but that's pulling in unrelated files and could be changed to specify the correct dependencies: `TCLFILES PACKAGES`.

src/comp/Makefile
```diff
@@ -61,23 +61,11 @@ HTCL_INC_FLAGS = -L$(HTCL_HS)
 HTCL_LIB_FLAGS = -lhtcl
 
 # TCL version
-ifeq ($(OSTYPE), Darwin)
-# Have Makefile avoid Homebrew's install of tcl on Mac
-TCLSH=/usr/bin/tclsh
-else
-TCLSH=tclsh
-endif
 TCL_VER = $(shell echo 'catch { puts [info tclversion]; exit 0}; exit 1' | $(TCLSH) || echo fail)
```
quark17, line 64: Remove this block of code for`TCL_VER`, too, since its only use was removed

src/comp/bsc.hs
```diff
@@ -1780,19 +1780,19 @@ cxxLink errh flags toplevel names creation_time = do
...
+        exportmap = let kernel = map toLower (binFmtToString getBinFmtType)
                     in  show $ (bluespecDir flags) ++ "/Bluesim/" ++
                                "bs_" ++ kernel ++ "_export_map.txt"
         -- this flag doesn't seem to work, so we use a separate call to "strip"
         stripflags = [] -- if (cDebug flags) then [] else ["-Wl,-x"]
         switches =
-          case getOSType of
-            Linux -> ["-shared", "-fPIC", "-Wl,-Bsymbolic"] ++ libdirflags ++
+          case getBinFmtType of
+            ELF -> ["-shared", "-fPIC", "-Wl,-Bsymbolic"] ++ libdirflags ++
```
quark17, line 1790: If you put two spaces between `ELF` and `->`, then it will align with the later lines

src/comp/bsc.hs
```diff
@@ -1780,19 +1780,19 @@ cxxLink errh flags toplevel names creation_time = do
         -- show is used for quoting
         libdirflags = (map (("-L"++) . show) (cLibPath flags))
         userlibs = map (("-l"++) . show) (cLibs flags)
-        exportmap = let kernel = map toLower (osToString getOSType)
+        exportmap = let kernel = map toLower (binFmtToString getBinFmtType)
```
quark17, line 1783: The name `kernel` is no longer an accurate description for this variable, since this isn't about the OS.  Change the name to something like `binfmt`

platform.sh
```diff
@@ -63,6 +68,92 @@ if [ "$1" = "c++_shared_flags" ] ; then
 fi
 
 ## =========================
+## Find the TCL shell command
+if [ ${OSTYPE} = "Darwin" ] ; then
```
quark17, line 72: A tiny nit, only if you happen to be editing this file for other reasons: I'd put an empty line after the `##` header comment.

quark17 (changes requested):
Thank you again for this work!  I have a few last things to fix -- but hopefully we're close.  (And let me know if you need me to take any of this over the finish line.)

This updates the yices submodule to its head as of 24 Aug 2021 (I believe?), which is great!  I see that Yices has release tags, and we could decide to stay at a specific tag.  But updating to the head is fine by me, and it passes the testsuite.

quark17:
I wasn't sure what you were getting at by these comments (and I'm not familiar with newer cabal, and not at all with stack), but let me see if I understand:

> I have a bit of time to work on this again. It looks as if FreeBSD no longer maintains GHC packages in the ports system and instead recommends using cabal or stack to build something with local copies of all dependencies.
> ...
> I've pushed some changes, but moving the build for the Haskell bits to using Cabal rather than relying on system-wide package installation is not something I know how to do.

As the `INSTALL.md` says, you can run `cabal` to install the libraries (when not available from the package manager):
```
$ cabal update
$ cabal v1-install regex-compat syb old-time split
```
This is using the older cabal `v1-install` (as opposed to the regular `cabal install` in newer versions).  I believe it should be installing the libraries locally, or at least can be configured to.  The newer `install` (aka `v2-install`) can also be used -- however one change would be needed in `src/comp/Makefile`, to give a flag to GHC to tell it where to find the v2-installed libraries (and I believe that flag is `-user-package-db`, although I could be misremembering).  (Also, you can give flags like `--lib` and `--package-env` to `cabal install` to tell it where to install the libraries, if you want them in the current directory, say.)

My understanding is that the recommended way of using `cabal` now is that you don't need to explicitly run `cabal install`, but that it happens automatically for you by `cabal build` for the project.  That is, we'd write a cabal file that describes how to build BSC and it lists the dependencies, that cabal would install into the local build.  But until we've written a cabal file, while we're still manually running GHC to compile BSC, we can also manually install the libraries locally (as cabal would do), and that should work.

If using `v2-install` is helpful, we could add a variable to the `src/comp/` makefile to control inserting the necessary GHC flag to tell it the path to the locally installed libraries.

davidchisnall:
If I understand correctly, the v2 version creates a self-contained install.  That would be ideal as it would then make it easy to create a package for bsc that didn't conflict with anything.

quark17:
> The newer `install` (aka `v2-install`) can also be used -- however one change would be needed in `src/comp/Makefile`, to give a flag to GHC to tell it where to find the v2-installed libraries

FYI, the command that I had used to be able to build with `v2-install` libraries was this:
```
GHC="ghc -package-env default" make
```

PR #265, opened 2020-09-30
SimExpand: Replace union with stableOrdNub in mergeUses to fix a scal…

nanavati:
…ing issue.

### DIFF
--- a/src/comp/SimExpand.hs
+++ b/src/comp/SimExpand.hs
@@ -42,7 +42,7 @@ import ListUtil (mapFst, mapSnd)
 
 import SCC (tsort)
 
-import Util (headOrErr, map_insertManyWith, allPairs)
+import Util (headOrErr, map_insertManyWith, allPairs, stableOrdNub)
 import GraphUtil(extractOneCycle_map, reverseMap)
 
 -- ===============
@@ -1952,7 +1952,7 @@ rUses m (UseInfo _ pred_reads body_reads body_writes) =
         (pred_read_uses, body_action_uses `union` body_read_uses)
 
 mergeUses :: [[(AId, AId)]] -> [(AId, AId)]
-mergeUses = foldr union []
+mergeUses = stableOrdNub . concat
 
 -- Returns the method uses in an action.
 -- A use is a method Id on a particular instance (instId, methId)


nanavati:
This happened to fix my particular performance issue, but I wonder if we should be going whole-hog and replacing [(AId,AId)] uses list with a Set everywhere.

bpfoley:
I'd say this point fix looks good enough to start with.

What is the performance issue it fixes? Do you have a small reproducer (or something we could turn into a reproducer) that we could add to the test suite?

nanavati:
It's an O(n^2) scaling issue that comes up while building a rule to used method map for Bluesim code generation. From what I can tell, a rule with hundreds of action method calls  (which may also need to be combined with hundreds of foreign function calls and/or value method calls) should trigger the problem. I haven't tried to make a standalone test case though.

quark17:
I'm attaching an example that takes 36 seconds in SimExpand on my mac with existing BSC and that goes to fractions of a second with this PR.  It is parameterized on a vector size `N` and these numbers are for 2048.  At 1024, SimExpand only takes 3-ish seconds.

[PR265.bsv](https://github.com/B-Lang-org/bsc/files/5374703/PR265.bsv.txt)

In Bluespec's bug database, there was one example (bug 1825) of SimExpand taking a long time, but it was an old example (and had never been boiled down).  It required some edits to get it compiling with newer BSC, and then it didn't take much time in SimExpand.  It's possible that my edits changed the compile-time, but I'd guess that the bug1825 example was just hitting an inefficiency that has since been fixed (possibly by you in 2017).  I doubt that the1825 example is creating large use sets, so I wouldn't expect it to exhibit the issue fixed by this PR.

The issue here is that SimExpand creates a map from rule names in the parent to a list of submodule methods that the rule calls, and it does this by walking over parts of the rule to generate lists for each part, which then are merged with `List.union`.  The spot fix is to merge those lists with a more efficient function that uses a `Set` to remove duplicates, but it is still concat'ing and walking `List` structures.  I agree that it might be a better fix for this process to just create `Set` values from the start, and to merge them with `Set.union`.  But are there any designs that start to strain the efficiency of this spot fix?  I would guess that, before then, other stages would start to be the bottleneck -- like `simMakeCBlocks` or `writeC`.  There are also other places in `SimExpand` that are working with lists, and it might be better to use sets for those, too.  Until we have examples that stress the current efficiency, I'm OK with just the spot fix.

It'd be interesting if the testsuite could be used to report stats about run-time in specific stages.  For example, to report the average time spent in `SimExpand` before and after this commit, and to report the largest changes in time for individual tests.  This would be useful for knowing whether a commit makes the situation worse for other designs (as well as possibly identifying other examples that are improved).

Issue #187, opened 2020-06-29
A successful build using stack.


drhodes:
Hi all, not sure if this is the right place for this message.  Anyways, I wanted to build bsc without installing a system wide ghc.  Six line changes later, bsc is building with stack, albeit with the checktool disabled.

For anyone interested, here's the diff: 
https://github.com/B-Lang-org/bsc/compare/master...drhodes:stack-build

-----
```bash
steps to build with the above changes 

/bsc $ stack --version
Version 2.1.3,

/bsc $ stack init 
/bsc $ stack install regex-compat syb old-time split

/bsc $ make all
success!
```
-----

This was easier than it should have been, no yaks were shaved, no deities invoked, etc, etc.  
(more information on [stack](https://docs.haskellstack.org/en/stable/README/))

quark17:
Cool.  Maybe someone can adjust the makefiles to support conditionally using stack.  FYI, the tool checks can be disabled by defining `NO_DEPS_CHECKS` when calling `make`.

PR #267, opened 2020-10-16
Makefiles now support haskell stack

drhodes:
Hi, this pull request enables building the Bluespec toolchain with haskell stack, as was discussed in this issue:

https://github.com/B-Lang-org/bsc/issues/187

I tested the build on two different debian machines, with and without a system-wide GHC installed using both `make install` and the new rule `make stack-install`. `make check` is passing.

The README has also been updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,11 +17,25 @@ rem_inst:
 rem_build:
 	rm -fr $(BUILDDIR)
 
+.PHONY: rem_stack
+rem_stack:
+	rm -f stack.yaml stack.yaml.lock
+	rm -fr $(TOP)/.stack-work
+	rm -fr $(TOP)/src/.stack-work 
+
 # -------------------------
 
 .PHONY: install
 install:
 	$(MAKE)  -C src  PREFIX=$(PREFIX)  install
+
+# -------------------------
+
+.PHONY: stack-install
+stack-install:
+	stack init --force
+	stack install regex-compat syb old-time split
+	-$(MAKE)  -C src  PREFIX=$(PREFIX) NO_DEPS_CHECKS=yes USING_STACK=yes install
 
 # In the future, this will be much more expansive, and run the actual test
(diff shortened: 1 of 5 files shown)


drhodes:
Oops, I wasn't sure if that merge branch commit message was an actual commit.  I can close this pull request, squash the two commits and open another pull request.

quark17:
Thank you for this.  Sorry that there hasn't been any activity on it.

Since this was submitted, the build instructions have been updated to explain how one can build using cabal to install dependencies:
```
$ cabal v2-install --package-env=default regex-compat syb old-time split
$ make GHC="ghc -package-env default"
```
I think that we could certainly add a similar mention of how one can build using stack currently:
```
$ stack install regex-compat syb old-time split
$ make GHC="stack ghc --" NO_DEPS_CHECKS=yes
```

It would be nice if both of these alternative build methods could be abstracted into the makefile, but I don't think I would do it the way you have proposed.  I would need to think on it some more.  (Offhand, I might prefer an environment variable like `GHC_TYPE=` with a fixed set of supported values, like `cabal-v2`, `stack`, etc.  This could be added to any target, instead of having to create an alternative for each target.)

It would also be nice if the dependency checks did not have to be disabled for stack.  The checks are currently done using `ghc-pkg`, but perhaps the check could be done with a different command when using a different GHC installation type.

However, it may not be worth implementing anything in the makefiles just yet.  There is some home that the current makefiles can be replaced by something better (like `cmake` or something else).  Maybe that other system would have a better way to choose between GHC types.

For now, I would be happy with just documenting the make command for building using stack.

drhodes:
Hi, thanks for taking the time to review my pull request.  I have a couple looming deadlines at the moment, but I put this on my todo list. About a make alternative .. I haven't used it, but nix has been getting a lot of good press these days, especially from haskellers.

PR #269, opened 2020-10-20
Handle empty structs properly in IConv

nanavati:
A minor cleanup to make a struct with no fields properly in IConv (rather than making an undefined value that's never used).

This means we can get rid of a workaround for PrimMakeUndefined in Deriving and a different one for undefined interfaces in IExpand and, most importantly, unblocks implementing PrimMakeUndefined with generics.

### DIFF
--- a/src/comp/IConv.hs
+++ b/src/comp/IConv.hs
@@ -20,6 +20,7 @@ import Error(internalError, ErrorHandle)
 import Flags(Flags)
 import Position
 import CSyntax
+import CSyntaxTypes
 import CFreeVars(getFVDl)
 import Id
 import PreStrings(fsTilde)
@@ -437,8 +438,14 @@ iConvE errh flags r env pvs (CConT ti c es) =
         let (t, (m, n)) = lookupConType flags ti c r
         in  iAps (ICon c (ICCon t m n)) [] (map (iConvE errh flags r env pvs) es)
 -- Ccase
-iConvE errh flags r env pvs (CStructT ct []) =
-  buildUndef flags r env (getPosition ct) UDontCare (iConvT flags r ct)
+iConvE errh flags r env pvs (CStructT ct []) = con
+  where con = iAPs ict itvs
+        ict = ICon ti (ICTuple tupty [])
+        tupty = foldr (\tv t -> ITForAll (tv_name tv) (iConvK $ tv_kind tv) t) it tvs
+        it = iConvT flags r ct
+        tvs = tv ct
+        itvs = map (iConvT flags r. TVar) tvs
+        ti = fst $ splitITApCon it
 iConvE errh flags r env pvs eee@(CStructT ct fs@((f,_):_)) =
  --trace (ppReadable (eee, map fst fs)) $
  --trace (ppReadable (map (\ (f,_) -> lookupSelType flags ti f r) fs)) $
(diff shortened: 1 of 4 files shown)


src/comp/bsc.hs
```diff
@@ -467,9 +467,9 @@ compilePackage
     --------------------------------------------
     start flags DFinternal
     imod <- iConvPackage errh flags symt mod'
-    iPCheck flags symt imod "internal"
     t <- dump errh flags t DFinternal dumpnames imod
     when (showISyntax flags) (putStrLnF (show imod))
+    iPCheck flags symt imod "internal"
```
quark17, line 472: Can you explain this edit?  Was it intended to be included in this PR?

The effect of this change is (1) to remove the time taken in `iPCheck` from the time reported for the stage (by `dump`) and (2) to allow dumping `imod` even if `iPCheck` fails.  Both of these are reasonable things to want -- I assume that this edit was to help in debugging.  However, this edit is only changing one place -- while every other call to `iPCheck`, `iMCheck`, and `aCheck` is still being done before the associated call to `dump`.  That suggests to me that this might have been an edit during debugging that wasn't intended to be checked in?

src/comp/IConv.hs
```diff
@@ -437,8 +438,14 @@ iConvE errh flags r env pvs (CConT ti c es) =
         let (t, (m, n)) = lookupConType flags ti c r
         in  iAps (ICon c (ICCon t m n)) [] (map (iConvE errh flags r env pvs) es)
 -- Ccase
-iConvE errh flags r env pvs (CStructT ct []) =
-  buildUndef flags r env (getPosition ct) UDontCare (iConvT flags r ct)
+iConvE errh flags r env pvs (CStructT ct []) = con
```
quark17, line 441: I wonder if this arm of `iConvE` could be unified with the next arm (for `CStructT` with non-empty `fs`).  It seems like they both produce the same `con` and then the non-empty arm just does the further step of wrapping it with `iAps con [] fs'`.  A single arm for all `CStruct` could avoid duplicating the code for `con`?

quark17 (changes requested):
Any idea why IExpand was using a don't-care?  Was there some reason for it in the past that is no longer a reason now?

This passes the testsuite, so I see no current reason to not do it. (Some `bsc.typechecker` tests that compare the ISyntax dump fail and will need to be updated to account for this change in the representation, but that's fine.)

If the two places I've commented can be resolved, I'd be happy to merge this.  Thanks!

src/comp/bsc.hs
nanavati: I did initially change it for debugging, but the new version was so much easier to understand that I thought it was worth keeping. I think that's partially because it's IConv (where we previously didn't have any ISyntax to look at if things go wrong). I'm not opposed to making the change more generally, but I think this particular case is a bigger deal than the others.

src/comp/IConv.hs
nanavati: While they're conceptually doing something quite similar, I think the details are very different.

Once there are fields, you need to look in the symbol table to understand the field type (and then reconcile it with the concrete types you have). That's a whole host of complications that don't come up in the field-free case (which was still tricky to figure out in its own way).

nanavati:
> Any idea why IExpand was using a don't-care? Was there some reason for it in the past that is no longer a reason now?

I think the reason was purely that making the right ICTuple for the field-free case is tricky. While is cleaner, it took Lucas and I several tries to figure it out. We wouldn't have bothered except for the impact on generically implementing PrimMakeUndefined (which, beyond the cleanup, helps us understand the limits of generics).

PR #275, opened 2020-11-03
Stop "optimizing" constructor applications against undefined values

nanavati:
That transformation can interfere with simplifying constructor tag tests for derived Bits instances.


quark17:
Oops, I meant to squash merge this.  Oh well.  Anyway, thanks for this.  It was interesting that the only testsuite difference was the re-appearance of warnings in one Bluesim simulation (bsc.verilog/sysMips).  I assume that the out-of-bounds regfile access is ultimately unused, and that resolving it would be better done with optimization or better code-gen in the Bluesim backend (rather than here in ISyntax).

PR #278, opened 2020-11-13
tweak to b37ad3c; install libstp_stub.so when STP_STUB=1

jcumming:
Hello, 

I'm updating the NixOS metadata to use bsc 103357f32cf63f2ca2b16ebc8e2c675ec5562464. 

To get everything running, I needed to make a small patch to `src/vendor/stp/Makefile` to copy `libstp_stub.so` to `inst/lib/SAT`. 

It would be great to be able to upstream this patch so we don't need to carry it anymore in the nixpkgs repo. The rest of the nix metadata changes are in jcumming/nixpkgs@442d1aed02bc112852e7542d0406bd78e4f72a62.

There are still 13 failures in `bsc-testsuite` (without SystemC), but those mostly seem to be NixOS-isms (like /bin/echo not existing).

### DIFF
--- a/src/vendor/stp/Makefile
+++ b/src/vendor/stp/Makefile
@@ -9,12 +9,13 @@ ifeq ($(STP_STUB),)
 SRC = src
 else
 SRC = src_stub
+SNAME += lib/libstp_stub.so
 endif
 
 ifeq ($(OSTYPE), Darwin)
-SNAME=libstp.dylib
+SNAME = lib/libstp.dylib
 else
-SNAME=libstp.so.1
+SNAME += lib/libstp.so.1
 endif
 
 all: install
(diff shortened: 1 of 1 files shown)


quark17:
This is a good catch, thanks.  I've been meaning to test building with `STP_STUB=1` on macOS, to see whether a similar change is needed in that case, too.

My feeling is that we might want to do one of two things: (A) Remove `SNAME` and install `lib/*`, so that the makefile installs whatever files were created, or (B) rename `SNAME` to `INSTALLNAMES` and make sure that we populate it properly with the names for all combinations of OSes and stub or no stub.

Also, if you've discovered that there are tools that the testsuite requires (such as `echo`) that are not mentioned in the README but are worth documenting, feel free to contribute that.

quark17:
PR #973 has an alternate fix.  Any objection if we go with that, and close this PR?  (My hope is that eventually we can eliminate the need for a stub object, by just statically compiling BSC without reference to STP, when requested.)

PR #283, opened 2020-12-12
Pack last clause

nanavati:
Add a carefully-constructed fallthrough testcase to derived pack to help optimizing "pack . unpack".

This looks like an unambiguous improvement and passes the testsuite so I did not add a flag for controlling the behavior.

### DIFF
--- a/src/comp/Deriving.hs
+++ b/src/comp/Deriving.hs
@@ -561,11 +561,26 @@ doDBits dpos type_name type_vars original_tags tags =
                 [CClause [CPCon1 type_name
                           (getCISName (headOrErr "doDBits" tags)) (CPVar id_x)] []
                  (cVApply idPack [vx])]
-            | otherwise = zipWith mkPk tags field_bit_sizes
+            | otherwise = zipWith mkPk tags field_bit_sizes ++ [packLast]
         mkPk tag field_sz =
             CClause [CPCon1 type_name (getCISName tag) (CPVar id_x)] []
                         (cVApply idPrimConcat
                          [litSz (cis_tag_encoding tag), pkBody field_sz])
+
+        -- Construct a custom fallthrough case to help (pack . unpack)
+        -- optimize to the identity. Explicitly using anyExpr instead of relying
+        -- on the UNoMatch case inserted by IConv means that this undetermined value
+        -- will not optimize away during static elaboration, preserving structure
+        -- that ITransform can rebuild into an identity.
+        -- The explicit tag ++ payload concatenation is not strictly necessary,
+        -- but should help make the reconstruction more robust now that
+        -- improveIf in IExpand can merge matching bit concatenations.
+        packLast = CClause [CPAny decl_position] []
+                           (cVApply idPrimConcat
+                                    [ hasSz (anyExprAt decl_position) num_tag_bits_ctype
+                                    , anyExprAt decl_position
+                                    ])
+
         pkBody sz = cVApply idPrimConcat [anyExprAt decl_position,
                                           hasSz (cVApply idPack [vx]) sz ]
         litSz k = hasSz (CLit $ num_to_cliteral_at decl_position k)
--- a/src/comp/Deriving.hs
+++ b/src/comp/Deriving.hs
@@ -561,6 +561,11 @@ doDBits dpos type_name type_vars original_tags tags =
                 [CClause [CPCon1 type_name
                           (getCISName (headOrErr "doDBits" tags)) (CPVar id_x)] []
                  (cVApply idPack [vx])]
+            -- Because our Boolean optimizations are good, the special last clause
+            -- adds overhead without any benefit for the two-constructor case.
+            -- This lines up with our special treatment if "boolean-like" constructors
+            -- in improveIf in IExpand.
+            | num_tags == 2 = zipWith mkPk tags field_bit_sizes
             | otherwise = zipWith mkPk tags field_bit_sizes ++ [packLast]
         mkPk tag field_sz =
             CClause [CPCon1 type_name (getCISName tag) (CPVar id_x)] []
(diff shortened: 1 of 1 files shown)


src/comp/Deriving.hs
```diff
@@ -561,11 +561,31 @@ doDBits dpos type_name type_vars original_tags tags =
                 [CClause [CPCon1 type_name
                           (getCISName (headOrErr "doDBits" tags)) (CPVar id_x)] []
                  (cVApply idPack [vx])]
-            | otherwise = zipWith mkPk tags field_bit_sizes
+            -- Because our Boolean optimizations are good, the special last clause
+            -- adds overhead without any benefit for the two-constructor case.
+            -- This lines up with our special treatment if "boolean-like" constructors
```
quark17, line 566: Change `if` to `of`

If it's OK to flatten your commits into one, then I can make this change when I accept the PR.  Or do you want to keep the separate commit history?

quark17 (commented):
I'm willing to accept this PR.  Just two questions:

(1) The `doDBits` function has one arm that applies for enums (union types where no constructor has data) and one arm that applies otherwise (union types where some have data).  You've added an explicit default for the second case.  Is there a reason to not do the same for enums?  The tests in `bsc.verilog/derived_pack/` include tests of an enum with power-of-two number of tags and an enum that doesn't fill the power-of-two range; however, in both cases the tags are contiguous.  (I don't recall if BH/Classic has convenient syntax for specifying gaps, which is what the test is written in.)  Is it worth adding non-contiguous examples to the testsuite?

(2) I have a sense that this is just a workaround -- BSC isn't optimizing the kind of don't-care that it creates for case-statement fall-through, so we create an explicit don't-care (that is a different kind that's handled differently).  We're fixing the derived `pack`; but users may also want to be aware of this when writing case-statements (to make sure to include explicit don't-care defaults), yes?  And if we do eventually fix the don't-care handling, could this explicit default become unnecessary? (I guess it won't hurt to have it?)

src/comp/Deriving.hs
nanavati: > I'm willing to accept this PR. Just two questions:
> 
> (1) The `doDBits` function has one arm that applies for enums (union types where no constructor has data) and one arm that applies otherwise (union types where some have data). You've added an explicit default for the second case. Is there a reason to not do the same for enums? The tests in `bsc.verilog/derived_pack/` include tests of an enum with power-of-two number of tags and an enum that doesn't fill the power-of-two range; however, in both cases the tags are contiguous. (I don't recall if BH/Classic has convenient syntax for specifying gaps, which is what the test is written in.) Is it worth adding non-contiguous examples to the testsuite?

BH/Classic doesn't have a convenient way to specify non-contiguous enums so I haven't been interested in that case. For ordinary contiguous enums no change is necessary - PrimOrd and PrimChr already cancel and there aren't the multiple moving parts (tag, padding and payload for generate ADTs) to deal with.

It's possible non-contiguous enums work fine because of the if-chain optimization mentioned below, but I haven't checked.

> (2) I have a sense that this is just a workaround -- BSC isn't optimizing the kind of don't-care that it creates for case-statement fall-through, so we create an explicit don't-care (that is a different kind that's handled differently). We're fixing the derived `pack`; but users may also want to be aware of this when writing case-statements (to make sure to include explicit don't-care defaults), yes? And if we do eventually fix the don't-care handling, could this explicit default become unnecessary? (I guess it won't hurt to have it?)

The special clause has two purposes:

1.  To add back in the (x == n) predicate for the last clause so that the:
    `if (x == 0) then 0 else if (x == 1) then 1 else if ... if (x == n ) n else _ => x`  optimization fires more reliably for the tag.

2. To break the don't care at the tag/payload boundary so that the new `improveIf` clause for `PrimConcat` properly separates the tag and payload.

I don't think either of those is particularly relevant to user code. I wouldn't expect a user to write that sort of if-chain outside of writing their own `Bits` instance (where they'll have to be careful for plenty of other reasons too).

And the `PrimConcat` improvement is implemented generally in the evaluator (in one of my follow-on changes).  The only special thing here is specifically trying to take advantage of it.

nanavati: > Change `if` to `of`
> 
> If it's OK to flatten your commits into one, then I can make this change when I accept the PR. Or do you want to keep the separate commit history?

If you want to flatten the commit history, you should delete the n == 2 special case. Further testing revealed that it actually doesn't help, I just don't have a targeted test case that shows that yet.

quark17: What do you mean by "it actually doesn't help"?  The comment that's on the special n==2 case suggests that it wasn't added because it helps; it seems to have been added because the n==2 case was already working and that an extra clause just adds extra overhead.  If you feel that the extra overhead is not significant, or not worth avoiding, then I can delete that clause.

quark17: > BH/Classic doesn't have a convenient way to specify non-contiguous enums so I haven't been interested in that case.

You may not be interested in improving that situation, but I think that we also need to be concerned with not regressing those situations either.  For example, when you add reduction rules for if-expr, I will want to know what impact it will have on pack/unpack for those.  Would you be opposed to adding a BSV example to the test directory?

nanavati: Rather than worrying about regressing the custom-tag case, I just fixed it: https://github.com/B-Lang-org/bsc/pull/298

quark17: You didn't answer the question about `num_tags==2`.  I'm ready to merge this, but I didn't understand whether that arm should be left in or taken out.

nanavati:
Has been superseded with better code. See https://github.com/nanavati/bsc/tree/pack-ord-tag (which hasn't been carved into a separate pull request for now until we've sorted out some of the existing tangles).

PR #291, opened 2021-01-14
Optimizations of concatenations with undefined values

nanavati:
This isn't quite ready-to-merge because I don't have all of the targeted test cases I want, but I wanted to share the code now to get feedback in our meeting (since once part is a little odd).

### DIFF
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -4372,6 +4372,16 @@ improveIf f t cnd (IAps (ICon i1 c1@(ICTuple {})) ts1 es1)
   -- unambiguous improvement since the ICTuple has propagated out
   return ((IAps (ICon i1 c1) ts1 es'), True)
 
+-- push if improvement inside bit concatenations with matching boundaries
+-- this is a post-pack version of the struct/tuple case above
+improveIf f t cnd thn@(IAps concat@(ICon _ (ICPrim _ PrimConcat)) ts1@[ITNum sx, ITNum sy, _] [thn_x, thn_y])
+                  els@(IAps        (ICon _ (ICPrim _ PrimConcat)) ts2                         [els_x, els_y])
+  | ts1 == ts2 = do
+  when doTraceIf $ traceM ("improveIf PrimConcat triggered " ++ ppReadable (cnd,thn,els))
+  (x', _) <- improveIf f (itBitN sx) cnd thn_x els_x
+  (y', _) <- improveIf f (itBitN sy) cnd thn_y els_y
+  return (IAps concat ts1 [x', y'], True)
+
 improveIf f t cnd thn@(IAps chr@(ICon _ (ICPrim _ PrimChr)) ts1 [chr_thn])
                   els@(IAps     (ICon _ (ICPrim _ PrimChr)) ts2 [chr_els]) = do
   when doTraceIf $ traceM ("improveIf PrimChr triggered " ++ show (cnd,thn,els))
--- a/src/comp/ITransform.hs
+++ b/src/comp/ITransform.hs
@@ -988,13 +988,13 @@ iTrAp ctx p@(ICon _ (ICPrim _ PrimConcat)) ts@[s1@(ITNum i1), s2@(ITNum i2), s3@
 
     -- (if c thn _) ++ e --> thn ++ e
     [IAps (ICon _ (ICPrim _ PrimIf)) _ [_, thn, els], e]
-      | isUndet thn -> iTrAp2 ctx p ts [els, e]
-      | isUndet els -> iTrAp2 ctx p ts [thn, e]
+      | isUndet thn && noRefs els -> iTrAp2 ctx p ts [els, e]
+      | isUndet els && noRefs thn -> iTrAp2 ctx p ts [thn, e]
 
     -- e ++ (if c thn _) --> e ++ thn
     [e, IAps (ICon _ (ICPrim _ PrimIf)) _ [_, thn, els]]
-      | isUndet thn -> iTrAp2 ctx p ts [e, els]
-      | isUndet els -> iTrAp2 ctx p ts [e, thn]
+      | isUndet thn && noRefs els -> iTrAp2 ctx p ts [e, els]
+      | isUndet els && noRefs thn -> iTrAp2 ctx p ts [e, thn]
 
     _ -> iTrApTail ctx p ts as
 
@@ -1344,6 +1344,12 @@ mkZero t = iMkLit (aitBit t) 0
 isUndet :: IExpr a -> Bool
 isUndet (ICon _ (ICUndet { imVal = Nothing })) = True
 isUndet _ = False
+
+-- Guard optimizations that are not valid in the presence of implicit conditions.
+noRefs :: IExpr a -> Bool
+noRefs (IRefT {})    = False
+noRefs (IAps f _ es) = all noRefs (f:es)
+noRefs _             = True
 
 isIfElseOfIConInt :: IExpr a -> Bool
 isIfElseOfIConInt (IAps (ICon _ (ICPrim _ PrimIf)) [t] [cnd, thn, els]) =
--- a/src/comp/ITransform.hs
+++ b/src/comp/ITransform.hs
@@ -985,6 +985,17 @@ iTrAp ctx p@(ICon _ (ICPrim _ PrimConcat)) ts@[s1@(ITNum i1), s2@(ITNum i2), s3@
         iTrAp2 ctx pif [itBitN i3] [cnd0, t', e']
       where t' = iTrApExp ctx p ts [t0, t1]
             e' = iTrApExp ctx p ts [e0, e1]
+
+    -- (if c thn _) ++ e --> thn ++ e
+    [IAps (ICon _ (ICPrim _ PrimIf)) _ [_, thn, els], e]
+      | isUndet thn -> iTrAp2 ctx p ts [els, e]
+      | isUndet els -> iTrAp2 ctx p ts [thn, e]
+
+    -- e ++ (if c thn _) --> e ++ thn
+    [e, IAps (ICon _ (ICPrim _ PrimIf)) _ [_, thn, els]]
+      | isUndet thn -> iTrAp2 ctx p ts [e, els]
+      | isUndet els -> iTrAp2 ctx p ts [e, thn]
+
     _ -> iTrApTail ctx p ts as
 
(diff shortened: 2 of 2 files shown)


quark17:
To continue the discussion from PR #283 : if this is all just about a particular case-expression on constructors ... why not explore, instead, writing the Bits instance using PrimChr/PrimOrd (as first suggested for enums in issue #205 )?  That is, something like:
```
pack x = primOrd (leftConOf x) ++ case ...
```
Where the case-expression for the data is the same, but the tags have been written directly as PrimOrd.  It seems like the whole point of the default arm and the reduction rules is to get to this point, so why not write it directly that way?  Then, we don't have to worry about whether the new reduction rules are bad for other situations.

Because that's what I want to understand: Pushing `if` into the sides of a concat (when the sizes are the same) makes sense for these concats (from the derived `pack`), but are there other situations where that is working against progress?  I think I'd feel safer if that reduction rule was conditional on whether the new form exposes further optimization (rather than always applying).  Or, if you can provide some evidence or reasoning that this optimization is always ok.

nanavati:
> To continue the discussion from PR #283 : if this is all just about a particular case-expression on constructors ... why not explore, instead, writing the Bits instance using PrimChr/PrimOrd (as first suggested for enums in issue #205 )? That is, something like:
> 
> ```
> pack x = primOrd (leftConOf x) ++ case ...
> ```
> 
> Where the case-expression for the data is the same, but the tags have been written directly as PrimOrd. It seems like the whole point of the default arm and the reduction rules is to get to this point, so why not write it directly that way? Then, we don't have to worry about whether the new reduction rules are bad for other situations.
> 
> Because that's what I want to understand: Pushing `if` into the sides of a concat (when the sizes are the same) makes sense for these concats (from the derived `pack`), but are there other situations where that is working against progress? I think I'd feel safer if that reduction rule was conditional on whether the new form exposes further optimization (rather than always applying). Or, if you can provide some evidence or reasoning that this optimization is always ok.

I've implemented the suggestion with PrimOrd and it looks pretty nice! See: https://github.com/nanavati/bsc/tree/pack-ord-tag
(should I carve that out as another pull request or is the stack already too large?)

I definitely prefer that to creating the custom-fragmented undefined value. And, as a bonus, when combined with my changes to optimize customized enumerations, we're _this_ close to supporting custom tags in derived Bits instances for all ADTs (not just enumerations). I think the only major thing missing is syntax.

That being said, these concat optimizations are still necessary to cover all of my derived Bits instance testcases. My main argument for why that's a good idea is that pushing ifs inside concats is the post-packing version of something we already do: pushing ifs inside ICTuple. More generally, it's the same mux whether the ifs are inside or outside the concat. The only difference is by pushing the ifs inside, we open the door to optimizations that only work on particular fragments. Does that make sense?

src/comp/IExpand.hs
```diff
@@ -4422,9 +4422,9 @@ improveIf f t cnd thn@(IAps ssp@(ICon _ (ICPrim _ PrimSetSelPosition)) ts1 [pos_
 -- mildly duplicates some work in doIf (isUndet thn)
 -- but the overlapping work for els has been moved here
 improveIf f t cnd thn els | ICon _ (ICUndet { iuKind = u }) <- thn,
-                            u == UNoMatch || u == UNotUsed || isSimpleType t = return (els, True)
+                            u == UNotUsed || isSimpleType t = return (els, True)
```
quark17, line 4451: Can you expand on the one sentence in the commit message, to say more about what's going on here?

src/comp/IExpand.hs
```diff
@@ -4372,6 +4372,32 @@ improveIf f t cnd (IAps (ICon i1 c1@(ICTuple {})) ts1 es1)
   -- unambiguous improvement since the ICTuple has propagated out
   return ((IAps (ICon i1 c1) ts1 es'), True)
 
+improveIf f t cnd thn@(IAps concat@(ICon _ (ICPrim _ PrimConcat)) ts1@[ITNum sx, ITNum sy, _] [thn_x, thn_y])
```
quark17, line 4375: Do you want to add comments on these, to explain and/or illustrate what each is doing?  (I also think it's worth including your explanation that this intuitively is the bitified equivalent of the preceding code to "push if improvement inside structs/tuples".)

src/comp/ITransform.hs
```diff
@@ -985,6 +985,17 @@ iTrAp ctx p@(ICon _ (ICPrim _ PrimConcat)) ts@[s1@(ITNum i1), s2@(ITNum i2), s3@
         iTrAp2 ctx pif [itBitN i3] [cnd0, t', e']
       where t' = iTrApExp ctx p ts [t0, t1]
             e' = iTrApExp ctx p ts [e0, e1]
+
+    -- (if c thn _) ++ e --> thn ++ e
```
quark17, line 989: What is meant by "restore" in the commit message?  Were these opts removed at some point?  Or maybe this commit is cherry-picked from a branch where you added the opts, but then removed them, but then restored them -- and so this commit was restoring the opts on that branch, but here it is adding the opts for the first time -- or?

quark17 (commented):
As mentioned out of band: I'm excited that this approach is working!  Thank you!  I know that you are still adjusting this PR, but here are some comments/questions I had on the current commits.

quark17:
I ran BSC with these commits on the testsuite (with testPR22), and had a quick look at the differences.  I'm sure that you've done the same, but in case my analysis is helpful, I'm sharing.

For the most part, it looks like improvements! But there's two tests that show changes (missing conditions) that may be regressions (but they need a closer look, because maybe it's an improvement).  Specifically:
```
Running ./bsc.bugs/b1894/b1894.exp ...
FAIL: did not match {...} in `bsc.bugs/b1894/mkTop.cxx'
```
Here, we're grepping for an expression that contains several terms, and one of the terms is a fall-through don't-care, which has been eliminated -- that's good -- but there's a second term that is also missing, after your changes, and I don't know if that's good.

And:
```
Running ./bsc.scheduler/relax-schedule/relax-schedule.exp ...
FAIL: `GetUrgency.bsv.bsc-sched-out.filtered' differs from `GetUrgency.bsv.bsc-sched-out.expected.filtered'
```
Here, part of the condition for the method `get` is missing!

And I guess a third place requires a closer look, too:
```
Running ./bsc.lib/rwire/rwire.exp ...
FAIL: `BypassReg.bsv.bsc-sched-out.filtered' differs from `BypassReg.bsv.bsc-sched-out.expected.filtered'
```
Here, in the dump of resources shows that the call to `the_r_the_r.write` has a different signal name as its argument.  This could just be a change in name, but it's worth seeing if it's not a bigger change.

The other things I noticed are:
```
Running ./bsc.evaluator/undefined/undefined.exp ...
FAIL: `sysListReg3D.v.filtered' differs from `sysListReg3D.v.expected.filtered'
```
The order of terms in the Verilog changed.  (In this case, possibly for the better, since for loops over i, j, and k now result in terms in that order, not reverse order.)

```
Running ./bsc.names/signal_names/signal_names.exp ...
FAIL: `Method.bsv.bsc-vcomp-out.filtered' differs from `Method.bsv.bsc-vcomp-out.expected.filtered'
```
Here, the expression `whas ++ (whas ? wget : _)` has been improved to just `whas ++ wget` and signal names created based on the RHS have improved (because the RHS has improved).

```
Running ./bsc.scheduler/relax-schedule/relax-schedule.exp ...
FAIL: `IBuffer2.bsv.bsc-sched-out.filtered' differs from `IBuffer2.bsv.bsc-sched-out.expected.filtered'
```
Again, the compiler-picked names have improved because the RHS has less concats in it (presumably because the output no longer things of the form `x[12:4] ++ x[3:0]` and now just has `x` or `x[12:0]`).  There is also a reordering of terms, but they seem to be the same terms (based on skimming, but I could be wrong).

```
Running ./bsc.verilog/undet/undet.exp ...
FAIL: did not match {b \? data\[31\:0\] \: incoming\[31\:0\]} in `bsc.verilog/undet/sysMaybeMux.v'
```
Here, the expressions like `data[31:0]` have just become `data`, which is a great.

```
Running ./bsc.verilog/verilog.exp ...
FAIL: `sysMips.c.out' differs from `sysMips.out.expected'
```
The iverilog output has different warnings -- these are warnings about array accesses that are ultimately unused, and I would guess that the don't-care value that's being presented to the array indexing has changed, so we see a different set of warnings.  I think that's probably OK.

src/comp/Deriving.hs
```diff
@@ -838,7 +839,7 @@ doDDeepSeqCond i vs ocs cs = Cinstance instance_cqt $
                     -- there are no contexts (see def of CInternalSummand)
                     -- so we don't need to check for them
                     else idPrimSeqCond
-        seq_clauses = map seqSummand cs
+        seq_clauses = map seqSummand cs ++ [CClause [CPVar id_x] [] (CVar $ idIdentityAt noPosition)]
```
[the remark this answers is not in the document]
nanavati: So, for a sum type, PrimDeepSeqCond needs to figure out what alternative it has and then recurse into the contents to surface implicit conditions within them. That corresponds to the seqSummand clauses.

But, for Bluespec, we also need to ask what happens when you fall off the edge of a case statement (as you might if the number of alternatives isn't a power of 2). In that case PrimDeepSeqCond makes a ''no-match" don't care. In the old world, where those don't cares are always thrown away and this doesn't matter.

In the new world, "no-match" don't cares *aren't* thrown away so the condition that takes you over the edge survives. If that structure survives then impCondOf will crawl it when surfacing implicit conditions - making logically equivalent but unnecessarily complicated expressions.

The insight here is that we don't need to expose a no-match don't care at the end of PrimDeepSeqCond. If you've tested for and recursed into all of the alternatives the only thing left is to not evaluate anything when you fall of the edge. Doing that means PrimDeepSeqCond ends on an identity function instead of a don't care - simplifying the implicit conditions exposed by a couple of tests.

Does that make sense?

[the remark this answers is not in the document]
quark17: Yes, I buy that: If we fall off the end of the clauses in PrimDeepSeqCond, we shouldn't return a don't-care but instead return `id` (indicating no further evaluation of the first argument `x` but still returning the second argument `y`).

src/comp/ITransform.hs
nanavati: I could have sworn I was restoring an optimization from a previous version of ITransform, but my code archaeology can't find it so I've changed the commit message.

src/comp/IExpand.hs
nanavati: I've added a comment to the main PrimConcat optimization. Further testing revealed that the other two alternatives (involving concats and undefined values) are no longer necessary (probably because of the changes in #300).

PR #295, opened 2021-01-16
prims: delete some old verilog files

### LINKED ISSUE #139: Verilog primitives should probably be renamed
Some of the Verilog primitives that are used by `bsc`-emitted code have rather unfortunate names. For example, taking the name `TriState` or `Counter` seems rather rude in the scope of a global project. Unfortunately, Verilog has no mechanism for controlling visibility of symbols, or doing namespaced/hygenic imports, so we're left with few options other than just renaming them to be as unique as possible. This is basically the same problem encountered in C++ headers which is why `__they_use_lots__of_underscored_names_you_wont_choose`.

While I don't think we need to go so far, perhaps just renaming all the modules to have the prefix `__BLUESPEC_` would be enough? So `FIFO0` becomes `__BLUESPEC_FIFO0`.

I also think that to be "future compatible", we should encourage tools interfacing with `bsc` to:

  1) Scan the `Verilog` directory, and track that every `.v` filename will map to a single Verilog module name (e.g. `__BLUESPEC_FIFO1.v` contains `module __BLUESPEC_FIFO1(...)`.) This could be made part of the `bsc` UI as well or something -- perhaps `bsc -prim-to-file __BLUESPEC_FIFO1` could print out the filename for use in automation, or you could just use Tcl.
  2) Assume any `__BLUESPEC_` entities that appear as "blackboxes" in a design can have the blackbox "filled" with one of these modules. This is what tools like [yosys-bluespec](https://github.com/thoughtpolice/yosys-bsv) do: they assume any missing entities in the module hierarchy that match the files in the `Verilog/` directory are to be "filled" by those modules.
  3) Users can also just add all the files to their project and let the synthesizer figure it out, like they currently do.

This should mean that users, provided they automate their tooling correctly, should be mostly oblivious to adding/changing/renaming Verilog primitives.

---

On a side note, `lib/Libraries/main.v`, which provides a tool-neutral harness for "one-click" simulation compiles, should probably be renamed and moved out of this directory, since it's not meant to be used in simulation, and

thoughtpolice:
These aren't used anywhere. Discovered in preparation to try and propose a fix #139.


thoughtpolice:
Re: point 3 &mdash; there is an odd discrepancy I discovered in the `GatedClock` primitive where it behaves completely differently on Vivado. I don't know what should be done about this, but it seems these are in conflict?

```diff
--- Verilog/GatedClock.v        2021-01-16 02:42:30.358357579 -0600
+++ Verilog.Vivado/GatedClock.v 2021-01-15 23:41:08.262363786 -0600
@@ -17,7 +17,7 @@
 // To avoid glitches, CLK_GATE_OUT only changes  when CLK_IN is low.
 // CLK_GATE_OUT follows CLK_GATE_IN in the same cycle, but COND is first
 // registered, thus delaying the gate condition by one cycle.
-// In this model, the oscillator CLK_OUT stop when the CLK_GATE_IN or
+// In this model, the oscillator CLK_OUT does *not* stop when the CLK_GATE_IN or
 // COND are deasserted.
 module GatedClock(
                  // ports for the internal register
@@ -54,7 +54,7 @@

    assign COND_OUT = COND_reg;

-   assign CLK_OUT = CLK_IN & new_gate ;
+   assign CLK_OUT = CLK_IN; // & new_gate ;
    assign CLK_GATE_OUT = new_gate ;

    // Use latch to avoid glitches
@@ -90,4 +90,3 @@
 `endif // BSV_NO_INITIAL_BLOCKS

 endmodule // GatedClock
-
```

quark17:
Thanks for contributing cleanups!  I haven't yet had a close look, but I think it's probably good to use ifdefs when the only difference is an attribute.  I've asked @czeck to have a look, too, to have a second (and better) set of eyes on the Verilog files.  But I did want to reply to your comment about `GatedClock`: The difference in these files is not a problem.

For gated clocks in a BSC design, BSC expects an oscillator wire and a gate wire, but BSC is agnostic about whether the oscillator still oscillates when the gate is de-asserted.  BSC does not use the wires in a way that it matters.  This allows the user to decide what is better for their synthesis (or simulation) target.  So it's OK for the generic and Vivado versions to differ -- although we can certainly revisit the decision.  (Maybe @czeck can say something about the original choices that went into it.)

czeck:
Hi Austin,
Thanks for the PR.  They look good.
I saw that you merged verilog.Vivado and Verilog.quartus.  How did you
handle the differences in BRAM2.v  I did not follow those differences.

WRT to Verilog/GatedClock.v, the bsc model disables actions when gate is
low, regardless of clock edge.   IIRC, Vivado (circa 2010) did not like
having a clock oscillator gated.  In their architecture, there are clock
regions each with dedicated clock routing, so throwing in a gate was not
allowed.  This led to the difference in the GatedClock module.    I do not
have a means to test this.

Which tools have you tried this verilog?
Best,
Ed.

quark17:
Thank you again for these cleanups.  As you can see, I have merged the first two commits, leaving the third.  (You may want to rebase this PR on the head, with just that commit?)

When a version of a module differs just by an attribute, having them in one file with an `ifdef` around the attribute makes sense just to avoid duplication that can get out of sync.  But I worry that it may be inevitable that different tools may need drastically different modules.  (For example, the Quartus versions of BRAM modules look quite different -- but they were created with old versions of Quartus and the situation may be improved now.)

I want to understand if the idea here is that all versions of a primitive would be defined in the same file.  That is, if the Quartus version is wildly different, then there may just be a second `module..endmodule` block in the file, and `ifdef` directives would pick the appropriate one?  The more I mull it over, the more I think I could get behind that idea.

PR #297, opened 2021-01-22
Extend the if-else "constant tree" optimization to include undefined values.

nanavati:
Specifically allow the LHS constants to be undefined in:
(if b1 then c1 else if b2 then c2 ... else if cn) == c -->
if b1 then (c1 == c) else if b2 then (c2 == c) ... else (cn == c)

Note that this will not create or destroy undefined values by itself
because any resulting equality will still be undefined.

### DIFF
--- a/src/comp/ITransform.hs
+++ b/src/comp/ITransform.hs
@@ -877,7 +877,8 @@ iTrAp ctx e0@(ICon _ (ICPrim _ PrimEQ)) [ITNum 1] [e, ICon _ (ICInt { iVal = Int
         | i == 1    = (e, True)
         | otherwise = internalError ("conApN ==")
 
--- if b1 then c1 else if b2 then c2 else ... cn == c
+-- (if b1 then c1 else if b2 then c2 else ... cn) == c -->
+-- if b1 then (c1 == c) else if b2 then (c2 == c) ... (cn == c)
 -- (other direction is handled by the flip rule below)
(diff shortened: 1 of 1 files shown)


quark17:
Nice catch!  I need to look a little closer, but I would guess that this is obviously a good improvement to merge.

Are there any other optimizations that look for constants, where an undefined should be accepted as well?  I would guess not -- most of them are places where, if the op is applying to an undefined, then it'd be best to just reduce that (to undefined or whatever) than to worry about reordering the expression.  (Although PrimExtract is different.  What happens to PrimExtract with an undefined index?)

nanavati:
I don't think so. The other places I see isIConInt in ITransform are places where the actual constants matter such as: `e + c1 == c2 --> e == c2 - c1` (or when using associativity or commutativity to permute expressions for more constant-folding).

As for PrimExtract with an undefined index, I made a quick example and it doesn't get out of typechecking (bit-vector of unknown size) which, in retrospect, isn't surprising.

The one gap I saw is that we have rules for `x == _ --> _`, but not for the other relational ops. I guess they don't come up nearly as much. Do you think those are worth adding?

quark17:
It would be interesting to see if rules for other relational ops make a difference.  The one I'd be most concerned with is not-equal, but I gather that there's no primitive for that (it just becomes PrimNot and PrimEq).

PR #299, opened 2021-01-26
Let people spell prefixes correctly in Classic.

nanavati:


quark17:
The `IfcPragma` that is generated is `PIPrefixStr`.  I wonder if this isn't a misspelling of the plural, but an abbreviation of "string"?

I also notice that in `Pragma.hs`, the attribute names between PPrint and PVPrint differ (both on "prefix" and "ports"):
```
instance PPrint IfcPragma where
    pPrint d _ (PIArgNames ids)       = text "arg_names ="   <+>
                                        brackets ( (sepList (map (ppVarId d) ids) comma) )
    pPrint d _ (PIPrefixStr flds)     = text "prefixs ="       <+> doubleQuotes (text flds)
    ...

instance PVPrint  IfcPragma where
    pvPrint d _ (PIArgNames ids)       = text "ports ="   <+>
                                         brackets ( (sepList (map (doubleQuotes . (ppVarId d)) ids) comma) )
    pvPrint d _ (PIPrefixStr flds)     = text "prefix ="       <+> doubleQuotes (text flds)
    ...
```

Is there a reason that the BSV names might not be available in the BH/Classic?  @czeck added this code, but that was in 2005, so it might be too wishful to ask if he remembers why.

Anyway, I would prefer adding support for `prefix` rather than `prefixes`.  And also, update the PPrint instance to match the preferred name.  (And I'd consider whether also the preferred name for `arg_names` should become `ports`.)

nanavati:
> The `IfcPragma` that is generated is `PIPrefixStr`. I wonder if this isn't a misspelling of the plural, but an abbreviation of "string"?
> 
> I also notice that in `Pragma.hs`, the attribute names between PPrint and PVPrint differ (both on "prefix" and "ports"):
> 
> ```
> instance PPrint IfcPragma where
>     pPrint d _ (PIArgNames ids)       = text "arg_names ="   <+>
>                                         brackets ( (sepList (map (ppVarId d) ids) comma) )
>     pPrint d _ (PIPrefixStr flds)     = text "prefixs ="       <+> doubleQuotes (text flds)
>     ...
> 
> instance PVPrint  IfcPragma where
>     pvPrint d _ (PIArgNames ids)       = text "ports ="   <+>
>                                          brackets ( (sepList (map (doubleQuotes . (ppVarId d)) ids) comma) )
>     pvPrint d _ (PIPrefixStr flds)     = text "prefix ="       <+> doubleQuotes (text flds)
>     ...
> ```
> 
> Is there a reason that the BSV names might not be available in the BH/Classic? @czeck added this code, but that was in 2005, so it might be too wishful to ask if he remembers why.
> 
> Anyway, I would prefer adding support for `prefix` rather than `prefixes`. And also, update the PPrint instance to match the preferred name. (And I'd consider whether also the preferred name for `arg_names` should become `ports`.)

My recollection is that everyone agreed that 'prefixs' was wrong, but that no one was going to see it anyway so it wasn't worth changing. That being said, prefix is a perfectly reasonable name and I'm happy to change that. The only reason for a plural would be if that attribute accumulated a list of strings from stacked subinterfaces, but it doesn't.

As for arg_names, I'm not wild about changing it to ports. When Lucas transitions to improved wrapper generation, one of the things I'm hoping to support is methods with multiple output ports (so that you can render a struct output naturally at a synthesis boundary but not pay the O(n^2) penalty in scheduling by having a method for each field).

quark17:
Keeping `arg_names` seems right (or, at least, it would be renamed `arg_ports`).  In the BSV syntax, the attribute is placed right on the argument identifier, and it's called `port`, which is fine because it's naming the port for the thing that it's attached to.  The BSV parser collects these up and puts them into one pragma under the covers.  The BH/Classic attribute syntax is different -- not attached to the individual arguments -- so it would need a name that clarifies that it's a list for the arguments.  That's fine.

For `prefixs`:  The BSV syntax proposal was always `prefix` from the start (in Bluespec bug 928 and then expanded on in an email discussion).  My guess is that, during code development, Ed thought that the Classic might have to do the same thing that was done for `port` on arguments: while BSV can have separate `prefix` attributes for each method, the Classic might have needed a single `prefixes` attribute on the interface, listing all the method prefixes at the same time (as method names paired with their prefix string).  That turned out not to be the case, and I'm guessing that Ed didn't update the attribute name or the comment, when he checked it in.  And then no one ever looked at it, since no one used it in Classic.  (There's only one commit for this feature.  It's not like both BSV and Classic used "prefixs" but we later updated BSV to just "prefix" and didn't also update Classic; and it's not that the attribute did take a list but was changed to not take it later.  There's also no mention of "prefixs" or "prefixes" in bug 928 or the email discussion.)  Anyway, the current resolution looks good, and I'll merge it.  (I might include a comment saying that "prefixs" is accepted for backwards compatibility.)

quark17:
Oh, except this fails to compile :)
```
Error: "GetPut.bs", line 116, column 27: (P0005)
  Unexpected "prefix"; expected "arg_names", "prefixs", "prefix", "ready",
  "enable", "result", "always_ready", "always_enabled", or "#-}"
```
One thing I notice is that the `expected` list is autogenerated and includes the `prefixs` that we're only keeping for backwards compatibility.  Do I recall correctly that there's a parser combinator that allows something to match but hides it from the `expected` list (that we could use to hide `prefixs`)?

And this may be the answer to my earlier question: "Is there a reason that the BSV names might not be available in the BH/Classic?"  Specifically, could `prefix` have been a keyword or something in Classic, that was the reason that a different name had to be picked?

quark17:
It does appear that `L_prefix` is a lexical token in `Lex.hs`.  It's not used by anything.  And I can't recall what meaning it could have had.

We could remove `L_prefix` -- what would be the down side of that?  Otherwise, we could have the parser accept `L_prefix` as an attribute name (which seems unclean, but I can't see why we can't do it); or we could pick a different name.

nanavati:
This is what I get for saying: "This is simple enough. No need to interrupt my testsuite run".

I did a little bit of archaeology and it looks like Lennart added it when doing some stuff with fixity. So far as I can tell, the compiler doesn't use it so I'd say to junk the keyword. Want to take care of that?

PR #300, opened 2021-01-26
Use PrimOrd to improve pack . unpack optimization by building the tag and payload separately

nanavati:


quark17:
I assume that this is a first step at replacing #291 (which contains a number of other commits, before getting to the PRs actual commits).  There are 5 real commits in that PR and this is the first two.  I assume that you're trying to break the problem down into smaller units that are easier to reason about and get merged -- or maybe to merge this part while you still experimenting with the later optimizations that are needed to get the output fully optimized.  Is this PR ready for review, or are you still working?

I had comments on these commits in #291, which still apply here.  You've answered one (on the change to PrimDeepSeqCond).  I'm still curious about the removal of UNoMatch from the improveIf rule -- I think maybe you mentioned it in your answer about PrimDeepSeqCond, but I could use some more clarification.

In #291, I also listed the testsuite changes that resulted, and I pointed out two tests that has suspicious changes (possibly regressions, that needed to be investigated).  I've run the testsuite for this PR and fortunately those two changes do not appear.  So I think it's probably safe to merge in this PR.  (And it causes 12 tests in `derived_bits` to start passing!)

However, there is one test where the change should be investigated, to make sure that it really is just a signal name change and not a behavior change.  That is this test:
```
Running ./bsc.lib/rwire/rwire.exp ...
FAIL: `BypassReg.bsv.bsc-sched-out.filtered' differs from `BypassReg.bsv.bsc-sched-out.expected.filtered'
```

quark17:
You're using `PrimOrd` for the `pack` function.  Would it help to use `PrimChr` in the `unpack` function?

nanavati:
> You're using `PrimOrd` for the `pack` function. Would it help to use `PrimChr` in the `unpack` function?

This may be a failure of imagination on my part, but I don't see how.

You can use PrimOrd on any ICCon for pack by extracting the tag and ignoring the contents. For unpacking you have to deal with them. I could imagine a new primitive that went from the tag value to the constructor function, but that would end having a different type at different tags. There may be ways to deal with that, but I think they would end up effectively equivalent to the constructor applications we're currently using in unpack.

Does that make sense or did you have a different idea?

quark17:
As discussed offline, that makes sense -- each constructor function could have a different type, so it wouldn't be possible to represent.

I also confirmed that the BypassReg difference is just a signal renaming.  The old behavior had this (when dumping prior to scheduling):
```
the_r_the_rw_wget____d2 :: Bit 16;
the_r_the_rw_wget____d2  = the_r_the_rw.wget;
```
while the new behavior has this:
```
x__h131 :: Bit 16;
x__h131  = the_r_the_rw.wget;
```
This does seem like a minor regression in the name, though.  I wonder if it's because we've used `CPVar id_x` in the construction of the pack function?  I wonder if either tagging the identifier with a property (bad? internal?) or using a different identifier (starting with underscores?) would cause BSC to recognize that it shouldn't preserve the name.

quark17:
I take that back.  The `x` here comes from `Prelude.bs`:
```
mkReg v = liftModule $
      ...
      interface
        ...
        _write x = fromPrimAction (_r.write (pack x))
```
In the `BypassReg` example, the value read from the RWire is written to a Reg.

That behavior is unrelated to the current PR, so I'll go ahead and merge this.  (Also, Deriving code has been using `id_x` since before this PR, so it's not something new introduced here.)

PR #301, opened 2021-01-27
Add LaTeX source for BSC documents

quark17:
Bluespec Inc had a couple iterations of the User Guide.  The latest version presented BSC through the lens of the BDW GUI, which is now in its own repo (as a separate tool built on Bluetcl).  For that repo, I had taken the User Guide and extracted just the BDW parts, to make a BDW document.  This time, I removed the BDW parts, leaving a User Guide document just for BSC (and Bluetcl).  I also removed mentions of licensing and tools or libraries that are not part of open BSC, and tried to clean up the language in places.  The resulting document could use a lot of work: the intro part, now that it doesn't have BDW, seems a bit redundant; a first few sections now just say "TBD", since the text there was too obsolete to keep.  And the contents might need to be updated in places.  Somewhere I have a script that checks that the flag documentation matches BSC's help message, which I will try to dig up, so that we can automate some checking of the documentation.  These commits add the building of the docs to the CI, but I don't think that it detects warnings -- such as a dangling cross-reference -- and it would be good to catch that.

This also contains a Libraries Reference Guide, which I constructed by extracting the libraries part of the BSV Language Manual, and again adjusting the language in places and removing libraries that are not in this repo.  Again, the document could use work.

The building of the docs occurs in the same CI as the build and test of the tool.  The installation of LaTeX packages adds a couple minutes to each OS environment -- it may be that we can install just the pieces that are needed, rather than whole packages, if that helps reduce the time.  Also, maybe it can be done in parallel, or perhaps in separate CIs (one for the tools, one for the docs), so that people can get results back sooner just for the tools?


quark17:
The README will need to be updated to mention the required tools for building the docs.  I haven't worked out what the precise requirements are, so if anyone felt compelled to do that, that'd be welcomed.

quark17:
I pushed a new version that makes both of the changes that I was considering:  The docs are not built by default with `make` from the root directory, but can be built with `make install-doc`; the CI builds the docs as separate jobs.  (The jobs are labelled "Build:" and "Build doc:", which might look too similar.  I wonder if just "Doc:" would be quickly visually distinguishable?)

Issue #276, opened 2020-11-08
Suggest missing 'valueOf' in error message


quark17:
If `valueOf` is left out of this code:
```
typedef 32 DataWidth;

function Bit#(DataWidth) fn (Bit#(DataWidth) x);
  //if (valueOf(DataWidth) == 64)
  if (DataWidth == 64)
    return 8;
  else
    return 4;
endfunction
```
then BSC, noticing that the name `DataWidth` is capital, looks for a constructor by that name and doesn't find one, so it reports an error about an unbound constructor name:
```
Error: "Test.bsv", line 5, column 7: (T0003)
  Unbound constructor `DataWidth'
```
It should be easy for BSC, in this case, to look for the name in the type domain and, if it's found, add a message that this name is a type and maybe a pseudo-function like `valueOf` is missing.

PR #315, opened 2021-02-04
Suggest missing 'valueOf' in error message

brooksbp:
I thought I would attempt #276 since it is labeled "good first issue". Please let me know how this change could be better. Thank you!

### DIFF
--- a/src/comp/Error.hs
+++ b/src/comp/Error.hs
@@ -691,7 +691,7 @@ data ErrMsg =
         | EBadInstanceOverlap String String Position
 
         | EUndefinedTask String
-        | EUnboundCon String
+        | EUnboundCon String (Maybe String)
         | EUnboundVar String
         | EUnboundField String
         | EUnboundClCon String
@@ -1880,8 +1880,13 @@ getErrorText (EDeriveCtx s) =
 
 getErrorText (EUndefinedTask c) =
     (Type 2, empty, s2par ("Unsupported system task " ++ ishow c))
-getErrorText (EUnboundCon c) =
-    (Type 3, empty, s2par ("Unbound constructor " ++ ishow c))
+getErrorText (EUnboundCon c maybeSuggest) =
+    (Type 3, empty,
+     let intro_msg = "Unbound constructor " ++ ishow c
+         msg = case maybeSuggest of
+           Nothing -> intro_msg
+           Just fname -> intro_msg ++ ".  Perhaps " ++ quote fname ++ " is missing?"
+     in s2par msg)
 getErrorText (EUnboundVar c) =
     (Type 4, empty, s2par ("Unbound variable " ++ ishow c))
 getErrorText (EUnboundField c) =
--- a/src/comp/TIMonad.hs
+++ b/src/comp/TIMonad.hs
@@ -468,9 +468,16 @@ findCons ct i = do
          Nothing -> errorAtId (EConstrAmb (pfpString ct')) i
          Just di -> case [ a | ConInfo {ci_id = i', ci_assump = a} <- cs, qualEq di i'] of
                    [a] -> return (updAssumpPos i a, di)
-                   []  -> errorAtId EUnboundCon i
+                   []  -> errSuggest r i
                    _   -> internalError "findCons ambig"
-     Nothing -> errorAtId EUnboundCon i
+     Nothing -> errSuggest r i
+  where
+    errSuggest :: SymTab -> Id -> TI (Assump, Id)
+    errSuggest r i =
+      case findType r i of
+        Just (TypeInfo _ KNum _ _) -> err (getIdPosition i, EUnboundCon (pfpString i) (Just "valueOf"))
+        Just (TypeInfo _ KStr _ _) -> err (getIdPosition i, EUnboundCon (pfpString i) (Just "stringOf"))
+        _ -> err (getIdPosition i, EUnboundCon (pfpString i) Nothing)
 
 findTyCon :: Id -> TI TyCon
 findTyCon i = do


quark17:
Awesome, thank you!

It might be better to reuse `EUnboundCon` and add an additional argument, rather than creating a new tag.  I would suggest a `String` argument, and if it's empty, then you get the original message, and if it contains a string then the message includes "maybe you meant" plus the string.

I suggest this because BSC now has a `KString` kind (in addition to `KNum`) and there is a `stringOf` pseudo-function to convert from the type to a String.  So we'll want a case arm for that, as well.

Also, `valueOf(c)` is in BSV syntax, but the user might have been writing in BH/Classic syntax.  Messages should display in the syntax that the file uses.  You could avoid the issue by just mentiong `valueOf` without applying it in an expression.  Otherwise, you can use the `isClassic` function to determine the syntax to use.  But maybe the most robust thing to do is to construct the CSyntax expression and then use `pfpString` to print it -- which consults `isClassic` to decide whether to use `ppString` or `pvString`.  (FYI, the CSyntax expression to create isn't straightforward either.  The most robust thing to do would be to create the same expression made by the parsers -- which `CVPrint` looks for an prints specially, although I notice that `PPrint` doesn't have special handling for it, which it probably should.  But I might just avoid printing an expression in the message, to avoid the issue.)

Also, I notice there are two places that generated an `EUnboundCon` error.  The other is a few lines above.  That case occurs if there happens to be constructors by that name, but none has the type that we're expecting.  In that case, would we also want to check for a numeric (or string) type by the same name?

That does point out: we have an expected type `ct`; should we check that it matches the return type of `valueOf` (or `stringOf`) before suggesting it?

Also: `findCons` is used in a few places where it may not be possible for `valueOf` to occur.  For instance, it's used to look up struct constructors, where I believe the syntax makes it clear that it's a struct.  We may want to pass an argument to `findCons` to indicate when we know that it must be a constructor, and not to do the additional check?

quark17:
That said, we don't have to be perfect to be helpful.  It's probably OK not to make any changes based on my last three paragraphs. (If we add the suggestion when and only when there is no constructor by the name but there is a type by the name, that should cover most situations and likely won't suggest it in wrong situations.)  But I would do the first three paragraphs (use the same tag, include `KString`, don't hardcode just BSV syntax).

brooksbp:
@quark17 please lmk whether latest push address first three paragraphs!

> That does point out: we have an expected type `ct`; should we check that it matches the return type of `valueOf` (or `stringOf`) before suggesting it?

How would this be done?  Could the `Kind` be compared against the `Type`, e.g. if the `findType` lookup returns a `TypeInfo _ KNum _ _`, check that the `ct :: Type` is a ... `TCon TyNum _ _`?

src/comp/Error.hs
```diff
@@ -1880,8 +1880,10 @@ getErrorText (EDeriveCtx s) =
 
 getErrorText (EUndefinedTask c) =
     (Type 2, empty, s2par ("Unsupported system task " ++ ishow c))
-getErrorText (EUnboundCon c) =
+getErrorText (EUnboundCon Nothing c) =
```
quark17, line 1883: I would keep these as a single case-arm, just to guarantee that the shared parts don't get out of sync.  For example:
```
getErrorText (EUnboundCon c maybeSuggest) =
   (Type 3, empty, s2par ("Unbound constructor" ++ missingFuncSuggestion maybeSuggest))
  where missingFuncSuggestion (Nothing) = ""
        missingFuncSuggestion (Just fname) = ".  Perhaps " ++ quote fname ++ " is missing?"
```
This is following the format of errors like `ENonTerminalActionValue`, that use a common function for the additional message.  Another possible model is `ECtxErrPrimParam`, which just has a case-expression inline (which might be more concise and readable, here).

src/comp/TIMonad.hs
```diff
@@ -468,9 +468,16 @@ findCons ct i = do
...
-                   []  -> errorAtId EUnboundCon i
+                   []  -> errSuggest r i
                    _   -> internalError "findCons ambig"
-     Nothing -> errorAtId EUnboundCon i
+     Nothing -> errSuggest r i
+  where
+    errSuggest :: SymTab -> Id -> TI (Assump, Id)
+    errSuggest r i =
+      case findType r i of
+        Just (TypeInfo _ KNum _ _) -> errorAtId (EUnboundCon (Just "missing valueOf")) i
```
quark17, line 478: I would just provide "valueOf" (and "stringOf") as the strings here, and leave the format of the message (and words like "missing") up to `getErrorText`.

quark17:
> > That does point out: we have an expected type `ct`; should we check that it matches the return type of `valueOf` (or `stringOf`) before suggesting it?
> 
> How would this be done? Could the `Kind` be compared against the `Type`, e.g. if the `findType` lookup returns a `TypeInfo _ KNum _ _`, check that the `ct :: Type` is a ... `TCon TyNum _ _`?

The typechecking of an expression like `fn(arg)` works by looking at the type of `fn` and seeing that it expects an argument of type `T` and then typechecking `arg` and checking that it is an expression of type `T`.  If not, then there's a type mismatch between the function and the argument.  What we are addressing here is the situation where the user wrote `fn(DataWidth)` but they should have written `fn(valueOf(DataWidth))`.  The return type of `valueOf(DataWidth)` is an `Integer` value.  If the function `fn` expects an `Integer` argument, then that might give us further confidence that the user should have included `valueOf`.

However, `fn` might work over multiple types.  In that case, the expected type might be a variable (say, `v`).  Typechecking will reconcile a return type of `Integer` with a variable `v` by learning that `v=Integer` (and replacing all uses of `v` with `Integer`).

If the `fn` expected type `Bit#(n)`, then simply adding `valueOf` around `DataWidth` would not cause the code to start compiler.  Something else would still be missing (possibly `fromInteger`, to convert from `Integer` to `Bit`).

My comment was that we could check whether the context (the `fn` in this case) is expecting either `Integer` or a variable (like `v`), and only suggest `valueOf` in those cases.  However, I think we should suggest `valueOf` regardless -- if the expected type is `Bit`, that doesn't imply that `valueOf` isn't needed, it could be that there is more that needs to be fixed.  I think it's a reasonable heuristic to always suggest `valueOf`, and let the user figure it out.  (And making sure to phrase the suggestion questioningly, to indicate that it's a heuristic, that may be wrong.)

To answer your question though: If we did want to check (which I now think we shouldn't), we'd match on `ct` using something like this:
```
case (ct) of
  // 'Integer' type
  TCon (TyCon i _ _) | (i == idInteger) -> True
  // variable
  TVar _ -> True
  // otherwise
  _ -> False
```

quark17:
FYI, this checking of `ct` is what is being done in the `Just cs` arm of `findConVis`.  That arm handles the situation where the identifier that the user wrote (`DataWidth`) happens to be the name of a constructor, and in fact the name of multiple constructors.  For example it might be a constructor of type `T1` and `T2`.  In that case, BSC looks at the type expected by the context (`ct`) to see if it matches `T1` or `T2`.  If it does, then `findCons` concludes that the identifier is that constructor; if `ct` is a variable, then BSC doesn't have enough info to disambiguate which constructor is being used, so BSC reports an `EConstrAmb` errors; if `ct` is an unrelated type (say it's `T3`), then BSC concludes that the user is referring to an unknown constructor and reports `EUnboundCon`.

We're adding "perhaps valueOf is missing?" to the `EUnboundCon` message, but note that we could also add it to the `EConstrAmb` error (when the identifer also has the same name as a numeric type)!  I don't imagine that this happens often.  But maybe it doesn't hurt to include it?  (To create an example that triggers this, take the original code and add two type declarations with constructors by the name `DataWidth`.)

src/comp/Error.hs
```diff
@@ -691,7 +691,7 @@ data ErrMsg =
         | EBadInstanceOverlap String String Position
 
         | EUndefinedTask String
-        | EUnboundCon String
+        | EUnboundCon (Maybe String) String
```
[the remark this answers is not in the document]
brooksbp: Fixed in latest push!

src/comp/Error.hs
brooksbp: Fixed in latest push, case-expression style!

src/comp/TIMonad.hs
brooksbp: Fixed in latest push!

brooksbp:
> handles the situation where the identifier that the user wrote (DataWidth) happens to be the name of a constructor, and in fact the name of multiple constructors

> (To create an example that triggers this, take the original code and add two type declarations with constructors by the name DataWidth.)

```bsv
typedef 32 DataWidth;
typedef Bit#(32) DataWidth;

function Bit#(DataWidth) fn(Bit#(DataWidth) x);
   if (DataWidth == 64)
      return 8;
   else
      return 4;
endfunction
```

```
Error: "/home/brian/test.bsv", line 2, column 18: (T0011)
  Declaration of `DataWidth' conflicts with previous declaration at
  "/home/brian/test.bsv", line 1, column 12
```

Hmm... how is it possible to have multiple constructors of the same name?

quark17:
> ```
> typedef Bit#(32) DataWidth;
> ```

This is defining a new type, not a new constructor.  To add new constructors, you would need something like this:
```
typedef union tagged {
  Bit#(16) DataWidth;
} T1;

typedef union tagged {
  Bit#(8) DataWidth;
} T2;
```

I notice, though, that if you add both of these to your existing example, you get a message like this:
```
Error: "Test.bsv", line 12, column 8: (T0019)
  Constructor `DataWidth' is not disambiguated by type `_tctyvar1009'
```
and if you add just one, then you get this:
```
Error: "Test.bsv", line 8, column 18: (T0031)
  The provisos for this expression could not be resolved because there are no
  instances of the form:
    Eq#(function Test::T1 f(Bit#(16) x1))
```
Neither of these are trigger the `EUnboundCon` case arm that you added.  They are testing the earlier case arms:
```
// When there is just one, BSC uses it
Just [ConInfo { ci_id = ti, ci_assump = a }] -> return (updAssumpPos i a, ti)

// When there are multiple, BSC looks at the type expected by the context;
// but if the context doesn't narrow it down, then BSC gives up and calls it ambiguous
Nothing -> errorAtId (EConstrAmb (pfpString ct')) i
```
The `EUnboundCon` error is reported when there is no constructor by that name, or when there are multiple by that name, but the context is expecting a specific type that matches none of them.  To trigger this second situation, the example needs to have a context that is expecting a specific type.  For example:
```
function Integer getDataWidth();
   return DataWidth;
endfunction
```
Or, like this:
```
function Bit#(DataWidth) fn(Bit#(DataWidth) x);
   Integer dw = DataWidth;
   if (dw == 64)
      return 8;
   else
      return 4;
endfunction
```
The problem with the original example is that the context is `== 64`, which can work for many different types -- `Integer`, `Bit#(8)`, `UInt#(32)`, etc.

These are interesting test cases, because maybe we want to update `findCons` to treat these situations different when there's also a type alias by the same name as the constructor(s).  However, you don't need to handle that in this PR.  And presumably it will be uncommon for the type alias to have the same name as a constructor -- really the most benefit is probably gained just from suggesting `valueOf` (and `stringOf`) in the case where there are no constructors by that name.

src/comp/TIMonad.hs
```diff
@@ -468,9 +468,16 @@ findCons ct i = do
...
          Just di -> case [ a | ConInfo {ci_id = i', ci_assump = a} <- cs, qualEq di i'] of
                    [a] -> return (updAssumpPos i a, di)
-                   []  -> errorAtId EUnboundCon i
+                   []  -> errSuggest r i
                    _   -> internalError "findCons ambig"
-     Nothing -> errorAtId EUnboundCon i
+     Nothing -> errSuggest r i
+  where
+    errSuggest :: SymTab -> Id -> TI (Assump, Id)
+    errSuggest r i =
```
quark17, line 476: I might have written it like this, to minimize duplication of code (that can get out of sync):
```
errSuggest r i =
  let mSuggest = case findType r i of
                   Just (TypeInfo _ KNum _ _) -> Just "valueOf"
                   Just (TypeInfo _ KStr _ _) -> Just "stringOf"
                   _ -> Nothing
  in  err (getIdPosition i, EUnboundCon (pfpString i) mSuggest)
```

brooksbp:
Somehow this PR was closed when I attempted to rebase to the latest `main` branch.  Here's the new PR https://github.com/B-Lang-org/bsc/pull/340

src/comp/TIMonad.hs
brooksbp: Fixed in latest push!

Issue #319, opened 2021-02-11
"Too much" elaboration of lists?


gameboo:
Some of our CHERI Flute/Toooba CI jobs recently started failing. After some investigation, I decided that it was probably appropriate to open an issue here regarding what I observe. The TLDR:

```
$ bsc -sim -g top SmallTest.bsv && bsc -sim -e top
Error: "List.bs", line 334, column 17: (S0015)
  Bluespec evaluation-time error: tail
  During elaboration of the body of rule `smallTest' at "SmallTest.bsv", line
  62, column 8.
  During elaboration of `top' at "SmallTest.bsv", line 56, column 8.
```

where `SmallTest.bsv` is:

```bsv
import List :: *;
//import Printf::*;

// ROT_BY_A FIRST_HOT_A : tail of empty list error
// ROT_BY_A FIRST_HOT_B : build ok
// ROT_BY_B FIRST_HOT_A : build ok
// ROT_BY_B FIRST_HOT_B : build ok

`define ROT_BY_A
`define FIRST_HOT_A

`ifdef ROT_BY_A
function List#(a) rotateBy(Integer n, List#(a) xs) =
  append(drop(n, xs), take(n, xs));
`else // ROT_BY_B
function List#(a) rotateBy(Integer n, List#(a) xs);
  for (Integer i = 0; i < n; i = i + 1) xs = rotate(xs);
  return xs;
endfunction
`endif

`ifdef FIRST_HOT_A
function Maybe#(List#(Bool)) firstHotToOneHot(List#(Bool) xs);
  List#(Bool) outList = Nil;
  Bool found = False;
  Integer n = length(xs);
  for (Integer i = 0; i < n; i = i + 1) begin
    Bool elem = head(xs);
    xs = tail(xs);
    outList = cons((!found) ? elem : False, outList);
    if (elem) found = True;
  end
  //return message(sprintf("%d", n), (found) ? Valid(reverse(outList)) : Invalid);
  return (found) ? Valid(reverse(outList)) : Invalid;
endfunction
`else // FIRST_HOT_B
function Maybe#(List#(Bool)) firstHotToOneHot(List#(Bool) xs);
  List#(Bool) outList = Nil;
  Bool found = False;
  Integer n = length(xs);
  for (Integer i = 0; i < n; i = i + 1) begin
    outList = cons((!found) ? xs[i] : False, outList);
    if (xs[i]) found = True;
  end
[...]

PR #322, opened 2021-02-18
Candidate fix for #319

nanavati:
Candidate fix for issue #319

Separate out and tweak logic for determining when to remove undetermined values in improveIf.

Add -trace-eval-if support for these cases.

### DIFF
--- a/src/comp/Deriving.hs
+++ b/src/comp/Deriving.hs
@@ -40,7 +40,6 @@ import PreIds(
               idPrimOrd, idPrimChr,
               idPrimSplit, idPrimConcat, idPrimTrunc,
               idFormat,
-              idIdentityAt,
               )
 import CSyntax
 import CSyntaxUtil
@@ -842,7 +841,7 @@ doDDeepSeqCond i vs ocs cs = Cinstance instance_cqt $
                     -- there are no contexts (see def of CInternalSummand)
                     -- so we don't need to check for them
                     else idPrimSeqCond
-        seq_clauses = map seqSummand cs ++ [CClause [CPVar id_x] [] (CVar $ idIdentityAt noPosition)]
+        seq_clauses = map seqSummand cs
         seq_def = CDef idPrimDeepSeqCondNQ fn_cqt seq_clauses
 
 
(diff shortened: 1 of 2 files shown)


quark17:
I confirm that this reverts all the behavioral changes reported in #319.  There are 5 failures in the testsuite:
```
Running ./bsc.bugs/b1894/b1894.exp ...
FAIL: did not match {2047u \& \(\(\(\(\(tUInt32\)\(\(tUInt8\)0u\)\) << 3u\) \| \(\(\(tUInt32\)\(DEF_cond__h[0-9]+\)\) << 2u\)\) \| \(tUInt32\)\(DEF_AVMeth_s_m\)\);} in `bsc.bugs/b1894/mkTop.cxx'

Running ./bsc.evaluator/undefined/undefined.exp ...
FAIL: `sysListReg3D.v.filtered' differs from `sysListReg3D.v.expected.filtered'

Running ./bsc.names/signal_names/signal_names.exp ...
FAIL: `Method.bsv.bsc-vcomp-out.filtered' differs from `Method.bsv.bsc-vcomp-out.expected.filtered'

Running ./bsc.scheduler/relax-schedule/relax-schedule.exp ...
FAIL: `GetUrgency.bsv.bsc-sched-out.filtered' differs from `GetUrgency.bsv.bsc-sched-out.expected.filtered'
FAIL: `IBuffer2.bsv.bsc-sched-out.filtered' differs from `IBuffer2.bsv.bsc-sched-out.expected.filtered'
```
These are tests whose expected output we changed, as part of the PRs that you're somewhat reverting with your PR, so it's not unexpected to have to update the output again.

For `bsc.evaluator` and `GetUrgency` in `bsc.scheduler`, we're reverting back to the expected behavior before any of your changes.  

For `IBuffer2`, we're reverting back some of the changes (the reordering of terms in method conditions), but there's still a difference, which seems to just be a change in unique numbers? (The diff is supposed to filter out unique suffix numbers, but these appear in the middle of def names created from RHS expressions.  I'd like to understand where these came from, to be sure that the difference is really benign.)

For `bsc.names`, only a slight difference is reverted.  That is, an `x == y` def that wasn't inlined before, but did get inlined with your first changes, but now is again not inlined:
```
x_wget__hNNNN :: Bit 8;
x_wget__hNNNN  = rg.wget;
-- IdProp x_wget__hNNNN[IdP_keep]
```

The failure in `b1894` is because a def wasn't inlined, but this is new behavior.  Before all of your changes, the expression was more complicated but contained `DEF_AVMeth_s_m`.  Now, the expression has been simplified, but that def is replaced with a def that's assigned to it:
```
DEF_v__h172 = DEF_AVMeth_s_m;
DEF__0_CONCAT_cond_CONCAT_s_m___d5 = 2047u & (((((tUInt32)((tUInt8)0u)) << 3u) | (((tUInt32)(DEF_cond__h160)) << 2u)) | (tUInt32)(DEF_v__h172));
```
I would not have expected this as a consequence of your change!  I'm also surprised that nothing between IExpand and the C++ output inlined a def of the form `x = y` (that's only used in this one place).

PR #324, opened 2021-02-22
Typofix BRAM error message

thoughtpolice:


quark17:
I notice that the module is actually called `mkBRAMAdapter`, not `mkBRAMServerAdapter`.  Do you think this message and the earlier `errorM` message should be updated to say `mkBRAMAdapter`?

thoughtpolice:
You're probably right, actually. Though, in either case I just immediately grepped for that exact string and would have found it in either case, so whatever it is should just be consistent!

PR #328, opened 2021-02-25
fix(lib): mark TriState with no_clock/no_reset

thoughtpolice:
Otherwise, the compiler won't allow stateless wrapper modules to use `no_default_clock`, etc.

### DIFF
--- a/src/Libraries/Base2/TriState.bsv
+++ b/src/Libraries/Base2/TriState.bsv
@@ -37,8 +37,8 @@ import "BVI" TriState =
 module vMkTriState#(Bool oe, t i)(TriState#(t))
    provisos(Bits#(t, st));
 
-   default_clock clk();
-   default_reset rst();
+   default_clock no_clock;
+   default_reset no_reset;
 
    parameter     width = valueof(st);
 


quark17:
Can you explain the situation that you're trying to fix?

The current declaration, with empty parentheses, indicates that the interface has an associated clock (and reset), but there is no port for it -- which also indicates that the module does not use the clock and it is therefore okay to be "no clock" (and "no reset").

The benefit of declaring an associated clock like this is that BSC will still do checking of clock domains: BSC will check that the places it is connected to are also declared with the same clock.  If you don't want the interface to have an associated clock (turning off BSC's checking), then you can instantiate the module with "clocked_by noClock" (and "reset_by noReset").

Your changes cause the module to never have checking, which I don't think is what we want.

thoughtpolice:
Oh, I was doing some testing of the FFI and used the TriState module as an example (I am using tristate support but that was unrelated), and realized the `CLK` and `RST_N` lines for the module seemed superfluous, so they could be removed. Even then I forgot to update `TriState.v`, so this patch is wrong anyway (that the CI doesn't catch that, though, is perhaps another issue.)

quark17:
There wouldn't be anything to patch in the Verilog file -- there are no `CLK` and `RST_N` lines here.  In a declaration like this:
```
default_clock clk();
```
The name `clk` is a bookkeeping name -- it's BSC's name for the clock of the module, separate from any ports that the module might have.  In this case, there are no Verilog ports.  If there was a `CLK` port, the statement would look like this:
```
default_clock clk(CLK);
```
And if there was also a gate port, it would look like this:
```
default_clock clk(CLK, CLK_GATE);
```
So when you changed the bookkeeping `clk` to `no_clock`, that does not change what ports are declared on the module.

Not to further throw information at you, but the use of `default_clock` with parentheses is actually a shorthand combining two statements into one statement: a declaration of an `input_clk` and then a declaration of that input clock as the `default_clock`.
```
input_clock clk(CLK);
default_clock clk;
```

PR #284, opened 2020-12-13
Implement and use a generics mechanism based on GHC.Generics

krame505:
Major changes here:
* Add the `Generic` type class and associated representation types in `Prelude.bs`. 
* Implement deriving for `Generic`. 
  * `Generic` is auto-derived on all data, struct and interface types.
  * Handling of types containing polymorphic fields required generating wrapper structs for the fields to use as their representation types.  
  * Numeric-sized collection types (`Vector` and `ListN`) have custom instances in the standard library that use `Vector` as their representation type.
* Re-implement the built-in type classes `PrimMakeUndefined`, `PrimMakeUninitialized` and `PrimDeepSeqCond` in using generics, and remove the existing deriving implementations for them.  
  * Note that this also required deriving `Generic` instances for primative types, and replacing the `PrimDeepSeqCond` instance generated by `GenWrap.hs` with a more general `Generic` instance that uses the original interface type as the representation of the wrapper.  
  * Documentation comments about these type classes were moved (and updated) from `Deriving.hs` to `Prelude.bs`.  
* Add a library `CShow` providing a Classic (Haskell)-syntax version of `FShow`.

This branch passes the testsuite with only a few minor changes, a corresponding pull request on that repository is coming shortly.

There are a few things that should be discussed before merging this:
*  For the moment I have added a `WExperimental` warning to uses of the `Generic` in a context as we may decide to tweak the API somewhat in the near future.  This is currently being done on the `ctxRedCQType'` function in `CtxRed.hs`.  Is this a reasonable place for this check?  
* There are also several new data types added to the prelude for encoding metadata, which we may like to change in the future as well, which someone could hypothetically use in their code without actually referencing the `Generic` type class.  However this seems like kind of a strange corner case, is skipping an additional `WExperimental` warning on these reasonable for now?
* Where should the `CShow` library live?  For now I added it to `Base1` since this is a generally-useful feature that has been requested in the past, but I could move it to `bsc-contrib` (or somewhere else) if you don't think that it belongs in the standard library.  Tests are included in my `bsc-testsuite` pull request.
* What should I do about documentation for generics (and `CShow`) in the future?  Ravi wasn't sure if the special doc comments in the standard library still worked for building the TeX document, and regardless information about how to write them doesn't seem to have been open-sourced.

PR #341, opened 2021-03-23
Performance fix, reinstate getQInsts short-circuit for Generic type class

krame505:
This reintroduces the performance workaround that previously existed for the prim type classes and was removed in #284, for the `Generic` type class.

According to my preliminary performance data, this seems to be around an 18% improvement in running the testsuite, and only about 17% slower than before merging #284 (that change lead to a ~33% slowdown.)  I am working on getting more accurate performance measurements.

### DIFF
--- a/src/comp/Deriving.hs
+++ b/src/comp/Deriving.hs
@@ -20,6 +20,7 @@ import PreIds(
               -- classes that the compiler can derive
               idEq, idBits, idFShow, idBounded, idDefaultValue,
               -- classes that are auto-derived
+              autoderivedClasses,
               idGeneric,
               -- internal classes defined in terms of Generic but still occasionally auto-derived
               idClsUninitialized, idUndefined,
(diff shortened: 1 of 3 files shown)


quark17:
As mentioned on the PR that removed this short-circuit, the code was added with the following comment:
```
Exempt built-in, auto-derived classes from overlap-checking
This limits the impact of the O(n^2) scaling issues because of
the O(n^2) instance sort / overlap check. Unfortunately, it
isn't an asymptotic fix.
```

I think I would prefer that `idGeneric` not be hardcoded here, but that the arm should instead match IDs that are elements of `autoderivedClasses`.  (Which is what the removed code had done, although it was called `requiredClasses` at the time.)

quark17:
And maybe add a short comment saying that this is for performance, because there is a Generic instance for every type and the overlap checking has O(n^2) sort

krame505:
Done.  Also had to move `autoderivedClasses` back to PreIds.hs to avoid a circular import resulting from importing Deriving.hs.

PR #347, opened 2021-04-01
Handle primitive types in Classic-syntax show

krame505:
Apparently handling for primative types was overlooked in classic-syntax show; I added an instance for `ConcPrim` in the generic implementation, as well as a regular instance for `Int`.

### DIFF
--- a/src/Libraries/Base1/CShow.bs
+++ b/src/Libraries/Base1/CShow.bs
@@ -17,7 +17,7 @@ class CShow a where
 instance CShow (UInt a) where
   cshow = fshow
 
-instance CShow Bool where
+instance CShow (Int a) where
   cshow = fshow
 
 instance (Generic a r, CShow' r) => CShow a where
@@ -32,6 +32,10 @@ class CShow' a where
 instance (CShow a) => CShow' (Conc a) where
   cshow' (Conc x) = cshow x
   cshowP' (Conc x) = cshowP x
+
+instance (FShow a) => CShow' (ConcPrim a) where
+  cshow' (ConcPrim x) = fshow x
+  cshowP' (ConcPrim x) = fshow x
 
 -- Note that below there are more specific instances for
 -- CShow' (Meta (MetaConsNamed ...)) and CShow' (Meta (MetaConsAnon ...))
--- a/src/Libraries/Base1/CShow.bs
+++ b/src/Libraries/Base1/CShow.bs
@@ -23,8 +23,32 @@ instance CShow (UInt a) where
 instance CShow (Int a) where
   cshow x = $format "%d" x
 
+instance CShow Real where
+  cshow x = $format (realToString x)
+
 instance CShow Char where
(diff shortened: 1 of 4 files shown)


quark17:
Using the FShow instances for Bit/UInt/Int results in values being printed in a form that isn't BH/Classic syntax (with `'h` for example).  Is that OK?  This is the syntax used by Verilog, so maybe it can be argued that designers are expected to know it?  Does it matter that the output can't be read back in as BH/Classic expressions?  (The FShow output often can't, so maybe that says it's ok.)  I would think that we'd ideally want explicit instances of CShow for these type that use code like `$format "0x%h" value` (instead of FShow's `$format "'h%h" value`).

I would also tend to not even fall back on FShow in any instances in this package (such as the ConcPrim generic instance).  I think I'd rather discover that an instance is missing, via an error, so that we go and fill it in (with a proper CShow definition) than silently print it using FShow -- although I can see benefit to the other way.

In any case, I'm happy to check in an incremental change and file an issue for any further cleanup, if you prefer.  But it might be good to extend the testcases to include more types, just so that we're regressing the current behavior on as much as we can.  Specifically, I think we're not covering: Char, String, Fmt, PrimUnit, Maybe, Either, tuples, and Real (for which there is no FShow instance?).

krame505:
> Using the FShow instances for Bit/UInt/Int results in values being printed in a form that isn't BH/Classic syntax (with 'h for example). Is that OK? This is the syntax used by Verilog, so maybe it can be argued that designers are expected to know it? Does it matter that the output can't be read back in as BH/Classic expressions? (The FShow output often can't, so maybe that says it's ok.) I would think that we'd ideally want explicit instances of CShow for these type that use code like $format "0x%h" value (instead of FShow's $format "'h%h" value).

Good point - I will fix that.

> I would also tend to not even fall back on FShow in any instances in this package (such as the ConcPrim generic instance). I think I'd rather discover that an instance is missing, via an error, so that we go and fill it in (with a proper CShow definition) than silently print it using FShow -- although I can see benefit to the other way.

I suppose that sounds reasonable. 

> In any case, I'm happy to check in an incremental change and file an issue for any further cleanup, if you prefer. But it might be good to extend the testcases to include more types, just so that we're regressing the current behavior on as much as we can. Specifically, I think we're not covering: Char, String, Fmt, PrimUnit, Maybe, Either, tuples, and Real (for which there is no FShow instance?).

I can try and cover more of these types in the testsuite - not sure how soon I will have time, though.

krame505:
I added explicit `CShow` instances and test cases for more primitive types, and got rid of the default using `FShow`.  Also added support for showing tuples of arbitrary length using another helper type class.

nanavati:
Side comment that just occurred to me: The relatively new sized literals Classic means that 'h 'b and 'd numbers are now supported. Aesthetically I'm not sure we'd want to use them because that's not what Haskell would do, but I thought it was worth noting.

PR #350, opened 2021-04-13
Include type arguments in generic representations

krame505:
This is needed for my work on generating C representations of types, for handling polymorphism.  

The kind-* type arguments to a polymorphic struct or data type are now included as a tuple in the `MetaData` metadata type constructor.  Numeric and string types are also included in this tuple wrapped in the `NumArg` and `StrArg` type constructors.  Other kinds of type parameters (e.g. `* -> *`) are omitted from the tuple - I'm not sure that there is a better option here, since B-lang doesn't support kind polymorphism.  

Note that GHC.Generics in Haskell doesn't include any information about type arguments in the metadata, so there is really nothing to compare with the approach here.


quark17:
Your example been updated to use this?  I want to try to understand it being used in action, to be able to judge it.  I was surprised that the meta data included the actual variable -- but that could be that I've forgotten the discussion that lead to this, so I'll have to try to review the example and remember.  I am a bit concerned that this ignores type variables of function kind.

krame505:
Yes, for example https://github.com/krame505/bsc-contrib/blob/gen-c-repr/Libraries/GenCRepr/GenCRepr.bs#L209

The metadata doesn't exactly "include the actual variable", only in the same sense that when a polymorphic data type is instantiated with some specific type arguments, the representation type created by the corresponding `Generic` instance is instantiated with the same types.  

I primarily see this being useful in computing a unique string identifiers corresponding to types of kind `*`.  That doesn't really make sense for arbitrary constructor kinds (in the same way that it doesn't make sense to print a function.)  There is no way to include arbitrary-kinded types in a generic representation without some significantly more advanced type features (possibly kind-level polymorphism?)  

Additionally constructor kinds as data type arguments are relatively uncommon - the only use that I have encountered is monad transformers, which we don't currently support and would be rather strange to use in conjunction with generics in a way that this would matter.  .

quark17:
By "include the actual variable" I meant that it references an actual type instead of, say, containing a string representation of the type.

For example, the `tyargs` parameter could be a tuple of string kinds -- although I guess the built in `printType` function returns a value, not a type, so I don't know how you would create the string-kind type.  And we could make a new type constructor, but it would be specific to the kind of its argument, so you'd need separate ones for each kind (unless we introduced pseudo-type-constructors?).  In any case, the advantage of storing the type is that the user of the Generic representation can decide what they want to do with it (rather than only being given the result of printing the name as a string).

Ah, what if the MetaData value constructor had an argument which is a list of String?  The instance would populate it using `printType` and then instead of using the MetaData type like this:
```
typeName' _ = stringOf name +++ typeNames (_ :: ta)
```
you'd get the value:
```
typeName' (MetaData typestrs) = stringOf name +++ concatWith insertUnderscore typestrs
```
or something like that?

You're probably right that mostly only star and numeric types will be encountered.  In theory, someone could write a type like this:
```
struct MyStruct c =
   f1 :: c 8
   f2 :: c 16
```
intending to use it for Bit, UInt, Int, etc.

krame505:
I'm not sure I follow how adding a value parameter to `MetaData` would help - recall that the various `Meta*` types are used to tag the representation type with `Meta` (e.g. `(Meta (MetaData ...) r)`, and don't actually appear in the value.  

But I think you can achieve the same thing (turning a tuple of types into a list of strings) with a type class accepting a proxy argument:
```
class TypeStrs a where
  typeStrs :: a -> List String

instance TypeStrs (a, b) where
  typeStrs _ = typeStrs (_ :: a) `append` typeStrs (_ :: b)

instance TypeStrs a where
  typeStrs _ = printType a :> nil
```

I had forgotten about `printType`, I suppose that could be used instead of generics to somewhat simplify things here.  But we may also want to do something a bit smarter in case two types with the same name from different packages are used, since package names are only accessible via generics.

krame505:
I made the changes we discussed yesterday, and also added a generic version of `printType` for kind-* types as a test case.  @quark17 does this look good to merge?

krame505:
@quark17 have you had a chance to look if this is good to merge yet?

src/Libraries/Base1/ListN.bs
```diff
@@ -1028,6 +1028,6 @@ mapPairs f g (L xs) = L (List.mapPairs f g xs)
 --X \end{itemize}
 
 -- The generic representation of a ListN is a vector of generic representations
-instance Generic (ListN n a) (Meta (MetaData "ListN" "ListN" 1) (Vector n (Conc a))) where
+instance Generic (ListN n a) (Meta (MetaData "ListN" "ListN" (NumArg n, a) 1) (Vector n (Conc a))) where
```
quark17, line 1031: Should `a` here be changed to `StarArg a`?

src/Libraries/Base1/Vector.bs
```diff
@@ -1347,7 +1347,7 @@ instance (DefaultValue t) => DefaultValue (Vector n t)
     defaultValue = replicate defaultValue
 
 -- The generic representation of a vector is a vector of generic representations
-instance Generic (Vector n a) (Meta (MetaData "Vector" "Vector" 1) (Vector n (Conc a))) where
+instance Generic (Vector n a) (Meta (MetaData "Vector" "Vector" (NumArg n, a) 1) (Vector n (Conc a))) where
```
quark17, line 1350: Should `a` here be changed to `StarArg a`?

testsuite/bsc.typechecker/generics/GenericTests.bs
```diff
@@ -111,15 +119,21 @@ quxFooRepr :: (Generic Qux r) => TypeEq r
...
+   (Meta (MetaConsAnon "UInt" 0 1)
+    (Meta (MetaField "_1" 0) (Conc (Bit 8)))))
+uint8Repr = refl
+
 bit8Repr :: (Generic (Bit 8) r) => TypeEq r (ConcPrim (Bit 8))
 bit8Repr = refl
 
 barVec3Repr :: (Generic (Vector 3 Bar) r) => TypeEq r
-  (Meta (MetaData "Vector" "Vector" 1) (Vector 3 (Conc Bar)))
+  (Meta (MetaData "Vector" "Vector" (NumArg 3, Bar) 1) (Vector 3 (Conc Bar)))
```
quark17, line 132: Should `Bar` here (and four lines later) be changed to `StarArg Bar`?

src/Libraries/Base1/ListN.bs
krame505: Yes, good catch.

PR #360, opened 2021-05-12
Add classic equivalents to some .bsv-specific testsuite tcl commands

krame505:
This is needed by the tests added in my bsc-contrib pull request https://github.com/B-Lang-org/bsc-contrib/pull/10


quark17:
We should rethink the interface for this before the explosion of functions gets out of hand (and at the very least it might be nice to rename the BH/Classic procedures to end with `_bh`), but that's not your problem to deal with today, so I'll go ahead and accept this.

PR #363, opened 2021-05-19
Pointer to Libera.Chat IRC (vs Freenode)

dnm:
Modified the mention in the README.md to point to #bluespec on Libera.Chat instead of Freenode. This should only be accepted & merged if Bluespec maintainers and the community agree with the idea of moving the IRC channel to Libera.Chat and off of Freenode.

I happened to unintentionally create #bluespec on Libera.Chat by just being the first to join the channel there. I've registered the channel with ChanServ to prevent against disconnects & netsplits and corresponding loss of ops; I'm happy to transfer channel "founder" status/"ownership" to whoever else (e.g., whoever started the channel on Freenode, etc.) and/or add others to ChanServ, in-channel ops, etc.


quark17:
Thank you!

PR #373, opened 2021-06-20
Autogenerate version names

quark17:
Rather than hardcode a version string in `Version.hs`, get the string from the generated file `BuildVersion.hs` where its value is created using `git describe --tags`, which will generate a name based on the current or nearest tag.  I hope to shortly tag a commit on `main` as `2021.07.1` -- the string generated for that commit will be the exact tag; if someone builds from source at a point after that, the string will be `2021.07.1-<count>-g<hash>` with the count of commits since the tag and the unique hash for the current commit.  When there is not an earlier tag (such as now), we construct the string `untagged-g<hash>`.

The source can be exported from Git by generating `BuildVersion.hs` and then setting `NOUPDATEBUILDVERSION=1` in `Makefile`, so that there is no attempt to regenerate the file.  I will add this info the README files, but there is already a pending PR to reorganize the READMEs, so I will do it after that has been merged.  (We may eventually want to provide a script that automates the exporting of source.)


quark17:
The Bluesim API includes a `get_version` function that models must implement, using a struct with three fields named `year`, `month`, and `annotation` -- these are when possible (if the version string had the form `X.Y` or `X.Y.STR`), and otherwise, if the string didn't match that form, `year` and `month` are set to zero and the whole string is put in the `annotation` field.  This PR replaces those three fields with a single `name` field that always contains the string, whatever form.  (The struct also contains other fields such as the `build` number -- aka the git commit hash -- and the creation time of the model.)

The Bluesim kernel also stores a copy of this version struct, statically, and uses that to check whether an instantiated model is compatible with the kernel, by checking for equality of the year/month/annotation (now equality of just the one name).  However, if the model was compiled with `-no-show-version` (as in the testsuite) then fields in the model in a `NULL`.  This hasn't caused a problem in the testsuite because the version string was hardcoded as empty, until now.  But now, we have to exempt such models from the version check.  (This PR makes that fix.)  The `-no-show-version` flag is used so that the generated files don't unnecessarily change from version to version of the compiler; we use it in the testsuite so that we can cache the results of compiling the Bluesim objects, and speed up the CI on new commits that don't otherwise change the generated C++.  For `.bo` and `.ba` files, the flag prevents recording the exact version, but there is still an API identifier that's encoded in the objects, and that is compared.  That identifier is only updated when changes are made in the file format.  We ought to do the same thing for Bluesim: only check that the API identifier matches, not that the compiler version matches.  (This PR does not yet make that improvement.)

The BDW tests are failing, probably because they were not correctly allowing for non-null version strings.  I'll have a look.

PR #402, opened 2021-08-28
fix typo in MIMO.tex (enqReadyN -> deqReadyN)

chenm001:


quark17:
Thank you!  I also notice some issues with the "arguments" column of that table, which I will push a fix for.

Issue #117, opened 2020-03-11
Investigate smtlib for bsc constraint solving


thoughtpolice:
### Setup

Right now, Bluespec links directly into Yices or STP for doing solving of various bitvector-like constraints. Or something like that. The default for now is Yices, and STP is optional, and we link to their object files, and that's really what matters.

However, doing this has a number of downsides:

  - **We must vendor `yices` and possibly other packages like `step`**. We should be moving away from this. It requires tooling like submodules, makes build integration more complicated, and ultimately third party systems already package things like this.

  - More substantially, **linking to the Yices API means the Bluespec compiler is a derived work under the GPLv3**, and also, any codebase derived from `bsc` will be, too. It's important to note that the code of `bsc` itself can fully be BSD2, but it is the *linking* part that actually produces the derived work (and implies a requirement to the source code, for resulting binaries, under the derivative/GPL clause). IANAL, but this is my understanding.

The first is problematic for a number of technical reasons (submodules, etc) but the second really hurts. People are also prone to get strange about licensing issues and may find this "deceptive" (because you advertise the codebase as BSD3, but some results are under the GPLv3 by law.)

---

### Proposed solution

I think it should be possible to avoid all this by doing something else instead: most modern SMT solvers (including Yices, Z3, STP from what I can tell, Boolector, and more) support formats like SMT-Lib. This is a project-neutral standard format you can feed a solver and have it produce answers for you.

The most prominent library for this is **[sbv](http://hackage.haskell.org/package/sbv)** in my opinion:

  - It's well-maintained, and Levent is very responsive.
  - It's seen successful, widespread use (in both commercial and open source projects, such as Cryptol, and others I can't speak much about.)
  - It doesn't require **ANY** FFI integration or libraries. Rather, it does the tricky work of portably launching the solver programs (`yices.exe`, `z3.exe`) in a cross platform manner, feeding them problem sets, and returning the results encoded as Haskell values.

These combination of features, IMO, make it the most worthy route to investigate, and feature 3 in particular not only breaks the GPLv3-by-linking problem, but opens the door to other solvers (`z3`, `boolector`) etc as well.

For anyone who wants to tackle this, I suggest looking for uses of `Yices` in the compiler codebase. The actual amount of uses of the Yices FFI library are small, so I'd suggest trying to surgically replace what we have with SBV, to the extent possible. We can then just fix the SBV solve function to use Yices for now.

---

### Upsides

[...]

Issue #51, opened 2020-02-11
Unvendor stp


bpfoley:
...and make it an optional dependency of Bluespec.

NB stp's build system is based on CMake, and currently there aren't packages for Debian/Ubuntu or Homebrew for Mac OS X. I'm working on the latter at the moment.

bpfoley:
STP homebrew Formula: https://github.com/Homebrew/homebrew-core/pull/50020

cbiffle:
Note that STP *also* appears to be GPL as a result of its Bit::Vector dependency. Unless I've missed something.

quark17:
I have not looked to see if the license situation has changed in the new STP available on GitHub, but for the snapshot here, the Bit::Vector library uses LGPL (as indicated in `extlib-constbv/constantbv.h` and reproduced in `LICENSES/LICENSE.stp_components`).  The BSC binaries dynamically link with the library built from the STP snapshot.

quark17:
PR #31 is related to this and has discussion.

cbiffle:
You are probably correct, I may have misread the LGPL as the GPL.

Issue #50, opened 2020-02-11
Unvendor yices2 


bpfoley:
...eventually, at a time where it's suitable for us.

Yices have
* formulas suitable for use with Homebrew on MacOS X,
* PPA packages suitable for use on Debian/Ubuntu
* Binary releases statically linked against the necessary libraries for other x86 Linuxes.

https://yices.csl.sri.com/doc/install-binaries.html

Un-vendoring means we won't have to maintain yices's dependencies, and probably makes bsc more suitable for packaging on Debian/Homebrew. We may have to do some legwork to get yices made into non-third party packages on Debian/Homebrew.

bpfoley:
NB also that yices2 is GPLv3.

cbiffle:
Just noticed that license myself -- any sense for how long Bluespec has been in violation of the GPL? I assume this component wasn't added in the open source release.

quark17:
As part of making this BSC repo, to put on GitHub, I extracted code from various places and created Makefiles and cleaned things up, etc.  I also decided to update the use of Yices, which predated Yices being open source (and GPLv3); so this was in fact just added, for this release.  (I had also wanted to update STP, but didn't get around to it.)  And, yes, it does mean that any distribution of BSC from this repo needs to respect the GPLv3 license or excise the use of Yices.

cbiffle:
Thanks for the clarification!

thoughtpolice:
Note that it's unclear if a newer version of Yices is needed or not for Bluespec -- without the test suite or real users of the FOSS version reporting in just yet, it's hard to know what behavioral differences there may be between now and then. They may or may not need to do a release for us to reliably unvendor things.

That said I'm hopeful if we needed it, and someone asked them nicely, they would cut a new release. :)

PR #404, opened 2021-08-29
Use rpath to find libstp.so

trixirt:
bsc depends on its wrapper to set the LD_LIBRARY_PATH to
it's private version of libstp.so.  Link with an rpath
to this library and bsc can be run without the wrapper.

Signed-off-by: Tom Rix <trix@redhat.com>

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -48,7 +48,7 @@ PARSEC_HS = ../Parsec
 # STP
 STP_HS = ../vendor/stp/include_hs
 STP_INC_FLAGS = -I../vendor/stp/include
-STP_LIB_FLAGS = -L../vendor/stp/lib -lstp
+STP_LIB_FLAGS = -L../vendor/stp/lib -optl-Wl,-rpath,'$$ORIGIN/../../lib/SAT' -lstp
 
 # Yices
 YICES_HS = ../vendor/yices/include_hs


quark17:
Thank you for contributing this.  However, I believe that it is already covered by PR #163.  That PR addresses a few issues that are not addressed here: the Makefile needs to also work for macOS, and the Yices library also needs to be found.

That PR hasn't been merged because it seemed like a better approach would be for the BSC executable to load the library with `dlopen` from a known location (rather than linking at runtime, and without hardcoding the BLUESPECDIR in an additional location).  No one has written that code yet, though.  But I will add it to the wishlist for the next release.

It's worth noting that the wrapper is still needed, to set BLUESPECDIR.

trixirt:
yices is in the same dir as stp and has stp as its dependency so the single rpath should work, though duplicating it would be fine too.
dlopen usually defers link errors to runtime when that happens to a user, they are normally stuck.
The BUILDSPECDIR is still need.

quark17:
BSC uses a SAT/SMT solver for a number of purposes.  Yices and STP are two separate implementations that the user can choose from, by providing the flags `-sat-yices` or `-sat-stp` on the command line.  (Yices is the default if no flags are provided.)

The very first thing that BSC does when executed is to check the flags.  During that, it checks which solver has been selected; and then it checks to see whether the linked library is of the right version.  (Note that this repo supports building a stub in place of either STP or Yices; if, for some reason, you can't build that on your system, or don't want to.  So BSC also needs to check that the library is not a stub!)  For Yices, there is a function in its API that returns the version number.  If the user has selected Yices, but the version does not match the expected version, then BSC warns the user and attempts to use STP instead.  STP does not have a version function in its API; but we still want to check that it is a not a stub, so we look for a non-NULL pointer returned by the init function.

There have been several proposed improvements to this situation:

* Don't offer a choice of solver.  Instead, remove STP and just use Yices.  In that case, we could even statically link with Yices, perhaps.  (See PR #31)
* Unvendor STP and Yices (see issues #50 and #51).  Then, it would be the user's responsibility to have the right version installed, and in the linker's search path
* Have BSC use a standardized API (such as SMT-lib), to which any solver (that supports the format) can be plugged in (see issue #117)
* Use `dlopen` (discussed on PR #163).  The BSC repo would not need to support building stubs in this case, because that could be implemented by just not having a library of the given name in the SAT directory, as BSC would check for that first before attempting to call `dlopen`.

Unvendoring Yices/STP and using a standardized API (SMT-lib) might be my preferred choice.  Although it could make BSC a little harder to install, because users would have to install an SMT solver first, and there are not convenient packages for those.

Applying an RPATH in the meantime is fine with me.  But there are still two problems that would need to be addressed: (1) Your change doesn't work for macOS; you'll need an if-else in the Makefile to use a different syntax for mac.  Your change doesn't remove the LD_LIBRARY_PATH from the wrapper, so the CI won't be testing it. (2) The relative location of BLUESPECDIR to the binaries (`../lib/`) should not be hardcoded in more than one place, if we can avoid it.  It's worth noting that, with the current installation of BSC, someone can move the BLUESPECDIR to wherever they want and just change the wrapper file; with RPATH, they would also now need to change the RPATH in the binaries.  (Probably not a big deal, but it'd be worth noting the new requirement in the build documentation.)

PR #408, opened 2021-09-06
Makefile: Add quotations around variables

omasanori:
It makes the build recipes robust against environment variables containing white spaces. For instance, `$PATH` on WSL contains `/mnt/c/Program Files/...` things by default, which causes build failures on that platform.


Makefile
```diff
@@ -28,25 +28,25 @@ help:
...
 rem_build:
-	rm -fr $(BUILDDIR)
+	rm -fr "$(BUILDDIR)"
 
 # -------------------------
 
 .PHONY: install-src
 install-src:
-	$(MAKE)  -C src  PREFIX=$(PREFIX)  install
+	"$(MAKE)"  -C src  PREFIX="$(PREFIX)"  install
```
jrtc27, line 41: Technically, yes, but the GNU make documentation lists the unquoted form as an example and if make is ever somewhere where that matters then you're going to have a bad time, nobody quotes `$(MAKE)`. Some people even abuse that with `MAKE="make ..."` to pass extra options so quoting it would break that.

Makefile
```diff
@@ -1,8 +1,8 @@
-PWD := $(shell pwd)
-TOP := $(PWD)
+PWD := "$(shell pwd)"
```
jrtc27, line 1: Quotes are meaningless in variables, they only have semantic meaning once the shell executes the line. As far as make is concerned, if `$(shell pwd)` expands to `foo bar` then this sets PWD to the variable containing `"foo` and `bar"` separated by a space. Moreover this chaining of variables leads to a mess:
```
rm -fr "$(PREFIX)"
```
is
```
rm -fr ""$(TOP)/inst""
```
which is
```
rm -fr """$(PWD)"/inst""
```
which is
```
rm -fr """"$(shell pwd)""/inst""
```
which ends up doing nothing useful because all the quotes are paired (in this case it's a 50:50 whether it works or not based on the parity of the number of levels of nesting).

Quoting should therefore always be done once, and at the last possible place, for make. Normally that means only in the shell lines themselves, with notable exceptions being for things like CFLAGS where you _want_ the spaces to be spaces and so need to quote individual arguments as you build the variable (e.g. `CFLAGS += -I"$(FOODIR)"`).

Makefile
omasanori: I see. Revert this part.

Makefile
omasanori: Indeed. I will fix.

omasanori:
Thank you for your accurate review, @jrtc27! I have revised the patch based on your comments.

quark17:
@omasanori I apologize that there was no movement on this!

I don't use Windows, but I see that I can produce a failure on macOS by cloning the repo into a directory with a space in the name and attempting to build.

omasanori:
@quark17 I am happy to hear the change was still relevant. It can not be applied cleanly to the current codebase, so I rebased it and proposed as https://github.com/B-Lang-org/bsc/pull/487

Issue #318, opened 2021-02-10
Use SMT to resolve numeric provisos where substitions are learned


quark17:
Often, BSC will report an error, that a numeric proviso (`Add`, `Mul`, `Div`, `Log`) is needed, but the user can see that the proviso is implied by provisos that are already given.  For example, the user wrote `Mul n 2 m` but BSC asks for `Add n n m`.  This specific example no longer arises, because a first step at fixing this has been implemented; however, many situations still occur, so a second fix is needed (details below).

This is bug 782 in Bluespec Inc's pre-GitHub bug database.  Bluespec Inc has collected dozens of examples that users have encountered.  I believe that most of the examples are in the testsuite, in `bsc.typechecker/numeric/`.  If any are not yet recorded there, I will add them there, rather than attaching them here in the issue ticket.

Details:

For resolving numeric typeclasses, BSC started out with just an ad hoc list of rules (specified in `StdPrel.hs`).  For example, if typechecking infers that `Add x y (TAdd x y)` is needed, then BSC has a rule that says this is trivially true and can be resolved.  Or, if typechecking infers that `Add x y t` is needed, where `t` is not a "bound" variable (that is, a variable that type inference created and not a variable that the user wrote in an explicit type signature), then BSC has a rule that says this is true when `t` is `TAdd x y`, so BSC can resolve this proviso and add `t = TAdd x y` to its knowledge (specifically, a substitution map).

These ad hoc rules only apply to one proviso.  But sometimes we want to resolve a proviso based on other provisos that we have been given or inferred.  For example, if typechecking infers that `Add n n m` is needed, and the user explicitly wrote the proviso `Mul n 2 m`, then BSC should determine that the `Add` proviso is met and can be resolved.  With multiple provisos, the combinations are too numerous for ad hoc rules to be feasible.  When fast SMT solvers were added to BSC, it became possible to use the solvers to detect these situations -- to detect when a proviso is satisfied given other provisos.  (This occurs in `TCMisc.hs` using an API defined in `SATPred.hs`.)

[...]

PR #410, opened 2021-09-13
Bug fixes for mkDivider and mkNonPipelinedDivider with non-zero iterations per cycle.

jonwoodruff:
Fix bugs with non-zero iterations per cycle for both mkDivider and mkNonPipelinedDivider.
Also improve style using a struct and functions and eliminate provisos.
Also improve the testbench to use random testing.


src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -6,76 +6,78 @@ import FIFO ::*;
...
       rule work;
-	 //match {.d, .q, .r} <- toGet(fThis).get;
-	 let x <- toGet(fThis).get;
-	 Int#(TAdd#(1,n)) d = tpl_1(x);
-	 Int#(TAdd#(1,n)) q = tpl_2(x);
-	 Int#(TAdd#(2,m)) r = tpl_3(x);
-	 Int#(TAdd#(2,m)) bigd = cExtendLSB(d);
-
+	 DivState#(n) f <- toGet(fThis).get;
+      	 Int#(TAdd#(2,TAdd#(n,n))) bigd = zeroExtendLSB(f.d);
```
jrtc27, line 51:
```suggestion
	 Int#(TAdd#(2,TAdd#(n,n))) bigd = zeroExtendLSB(f.d);
```

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,214 +124,181 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-      end
-
-      if (rg_index == fromInteger(valueOf(n) / s)) begin
-	 rg_busy <= False;
-	 crg_done[1] <= True;
+        if (!done(count)) begin
+           if (f.r >= 0) begin
+               f.q = (f.q << 1) | 1;
+               f.r = (f.r << 1) - bigd;
+            end
```
jrtc27, line 147: Inconsistent whitespace here (and seemingly later) too

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: Thanks @jrtc27!  I've made a pass on whitespace.  At first I tried to preserve the tab/space behaviour for lines that we didn't touch, but it looks like we touched all the lines with that split anyway, so I just went for all spaces.  This is committed to the source branch now.

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: I've attempted to resolve this one in the source branch as well.  Thanks again @jrtc27!

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +124,180 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
       endmethod
    endinterface
-
 endmodule
 
-typedef 11 MBits;
-typedef 22 NBits;
+typedef 7 NBits;
 
+(*synthesize*)
```
quark17, line 222: I would prefer not to include a synthesize attribute here.

A number of libraries follow the convention of including a testbench at the end.  We could discuss whether this is worthwhile -- maybe we should just move that code into the testsuite, yes?

If the code is going to remain here, the idea has typically been to not have a synthesize attribute.  That way, if a developer wants to run a test, they can do it by manually providing the `-g mkTb` flag; but, otherwise, there isn't a synthesized module sitting in the BSC libraries.  (There is still the module definition, in the library.  I'd be in favor of removing that, too.  Unless you think it serves a purpose to end users?)

The place for tests for libraries is `testing/bsc.libs/`.  You'll see that there are subdirectories for some of the other Math packages, like Complex and FixedPoint and FloatingPoint.  There's no directory for Divide, but we could create one.  I do noticed that the FloatingPoint test includes division, using the Divide library.

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +124,180 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
+   Server#(Tuple2#(Int#(TAdd#(NBits,NBits)),Int#(NBits)),Tuple2#(Int#(NBits),Int#(NBits))) sdiv_mod <- mkSignedDivider(2);
+   FIFO#(Tuple2#(Int#(TAdd#(NBits,NBits)),Int#(NBits))) sdivs <- mkSizedFIFO(16);
 
-   function Action testDividePipe(Integer n, Integer d);
+   function Action testDividePipe(UInt#(TAdd#(NBits,NBits)) ni, UInt#(NBits) di);
       action
-	 UInt#(NBits) ni = fromInteger(n);
-	 UInt#(MBits) di = fromInteger(d);
-	 UInt#(MBits) q = fromInteger(quot(n,d));
-	 UInt#(MBits) p = fromInteger(rem(n,d));
```
quark17, line 275: This code was computing results statically (using Integer quot/rem operators in the elaborator) and comparing the HW results against that.  You've removed this, and are only comparing two HW dividers against each other (pipelined vs non-pipelined).  It would still be good to compare them against a static result.

The callers of this function (like `testDividePipe(1,2)`) were passing in Integer values, so that was why static results could be computed.  You've replaced those with a HW loop that provides random values.  There's no reason why we can't have both -- some known static values and loop of random values -- particularly if there are some specific corner case inputs that we want to provide.  However, we could just have the HW loop, and instead of comparing against a statically computed result, we compare the two HW dividers against a simple sim-time computation: if not a Verilog quot or rem operator, then we could import C functions to compute it at sim time.  That would give me more confidence in the results than just comparing our two HW dividers against each other.

(Same comments apply for `testSignedDividePipe` below.)

quark17 (changes requested):
Thanks!  I am not familiar with the hardware that this library creates, so I'm unable to give an authoritative review on your HW implementation changes, but I've given it a shot.  And, of course, I can comment on cosmetic things.

In addition to the inline comments, I will add this general comment (on the use of TAdd type constructor vs Add proviso):

BSC was asking for provisos like these:
```
Add#(b__, n, m),
```
because we'd written the relationship between `m` and `n` like this:
```
Mul#(2, n, m)
```
instead of writing it like this:
```
Add#(n, n, m)
```
BSC should know that the first implies the second, but there's a known limitation in BSC's numeric solving, which can eventually be fixed (see issue #318).

You've essentially replaced the Mul proviso with the Add proviso, but you've written it implicitly by putting `TAdd#(n,n)` in the type.  I'm nervous about writing the type like this; BSC ought to be converting it to an `Add` proviso, but I'm nervous that bugs might exist.  I would prefer if you removed `TAdd` from the base types and put back `m`, with the proviso `Add#(n, n, m)`.

For example, the module `mkSignedDivider` only needed the `Mul` proviso changed to `Add` and then remove the other two provisos, and nothing else in the module should have needed to change.

Also, in the test module at the end, I would reinstate `MBits` and remove all the uses of `TAdd#(NBits,NBits)` everywhere.  If you want to define `MBits` as `TAdd#(NBits,NBits)`, in one place, that would be fine.  (I do note that previously `NBits` and `MBits` were the opposite of `m` and `n` in the module definitions, which is maybe confusing.)

src/Libraries/Base3-Math/Divide.bsv
quark17: The tests in `testing/bsc.lib/SquareRoot/` might be the closest example to what we could do for Divide.

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -6,76 +6,78 @@ import FIFO ::*;
...
+                         });
    endrule
 
-   FIFO#(Tuple3#(Int#(TAdd#(1,n)),Int#(TAdd#(1,n)),Int#(TAdd#(2,m)))) fThis = fFirst;
-   FIFO#(Tuple3#(Int#(TAdd#(1,n)),Int#(TAdd#(1,n)),Int#(TAdd#(2,m)))) fNext;
+   FIFO#(DivState#(n)) fThis = fFirst;
+   FIFO#(DivState#(n)) fNext;
 
-   for (Integer i = 0; i < (valueOf(n)/s+1); i = i + 1) begin
+   for (countT i = 0; !done(i); i = i + fromInteger(s)) begin
```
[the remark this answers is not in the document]
jonwoodruff: We certainly needed the `i = i + s`. 

The purpose of the `done` function was to ensure that the exit condition for the outer loop and the inner loop were obviously the same, rather than `i < (valueOf(n)/s+1)` vs. `(i + j) <= valueOf(n)`.

The `countT` is not necessary in the pipelined version.  It was needed in the UnPipelined version as we needed a register to hold the iterator. I've moved to Integer for the pipelined version while retaining the same style between the two as much as possible.

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +124,180 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
+   Reg#(countT) rg_count <- mkReg(0);
 
-   rule work(rg_busy);
-      Int#(TAdd#(1,n)) d = rg_d;
-      Int#(TAdd#(1,n)) q = rg_q;
-      Int#(TAdd#(2,m)) r = rg_r;
-      Int#(TAdd#(2,m)) bigd = cExtendLSB(d);
+   function Bool done(countT cmp) = (cmp > fromInteger(valueOf(n)));
 
+   rule work (rg_busy && !done(rg_count));
```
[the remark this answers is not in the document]
jonwoodruff: I've moved the update to count out of the condition, which does cause the hardware to collapse to `assign MUX_div_dut_rg_count$write_1__VAL_1 = div_dut_rg_count + 4'd3 ;`, for example.
I don't know if being precise with `rg_count ` was necessary in some previous iteration, but this works fine now.

quark17:
Thank you for your patience!

Functional correctness is one thing, but there are also considerations around how many clock cycles things take (is it n, n+1, or n+2, say) and whether putting a new request can happen on the same clock as getting a result, and the size of the design (in state bits and/or logic).  I don't have a sense of whether the current modules were written the way they are in order to meet requirements in these areas. It would be easier to accept changes that are just to the types, and to the computing of indices and done conditions; but in the non-pipelined module, you have also changed how the communication happens between the rule and the methods (and I think there may have also been a place that changed a register to a FIFO etc).  I didn't know if these changes are necessary, so I took some time to play around with it myself; I attempt to just apply the changes to the indices/done and to the provisos (and using your new test module).  The test module is only testing for functional correctness; if we had tests to check for regressions in the other considerations (clock cycles, simultaneous enq/deq, area), that would help me feel better about accepting changes.

In experimenting, I did discover that the existing non-pipelined module can give different results, depending on which clock cycles you put/get!  I have a mockup in which I only applied the index/done changes, and didn't change the use of `CReg` etc.  I discovered that when comparing this against a pipeline model (`div_mod`) of `mkDivider(1)`, that the test reported a difference, but if I compared against `mkDivider(3)` that the test said they matched!  And if I compared `mkDivider(1)` and `mkDivider(3)` against each other, they also matched.  My hypothesis is that, when using `mkDivider(3)` as the model, the timing of put/get is different, and that leads to different result from the non-pipelined module.

So clearly the existing module is broken and does need a change in the architecture.  If your version doesn't exhibit this problem, that's a good starting point.  And I think, now that we're a public repo and people can speak up if they are using this module and if their expectations about it are broken, that I am comfortable accepting a PR.  It would help to know if you have analyzed these aspects though (the clock cycle count, the put/get simultaneity, and any change in longest logic paths) and I might suggest adding some comments in the code.  (I should also check if the Library Guide document says anything about these properties of the module, that we should respect.)  We can also look into adding tests for these other features -- I can try to do that, if you don't have more time to put towards this.

However ... having said all that ... I now realize that there is a `NonPipelinedMath` library in `bsc-contrib`, with modules that came from Cambridge and look about like what you're contributing.  Looking back in the history I see that, in May 2014, Bluespec updated the `FloatingPoint` divide and squareroot modules to take the divider/sqrt module as an argument (so the user can swap in different implementations) -- based on a request by Cambridge -- and Bluespec also accepted the non-pipelined versions, but put them in a `Contrib` directory at the time (not in the `Divide` and `SquareRoot` libraries).   That's where I thought things were left, so I am surprised to see that in Sept 2014, Bluespec did add non-pipelined versions to Divide and SquareRoot -- and these appear to be based on the contributed modules, but have the architectural changes that you see now.  Presumably someone had a reason for the modifications.   That was committed by @darius-bluespec, so I'll see if he remembers.  In any case, if we get this resolved, we can also remove `NonPipelinedMath` from `bsc-contrib`.

Thank you again for your patience as I work through this!

PS: Note that there's no need to define `zeroExtendLSB`, as that exists in the `BUtils` library as `zExtendLSB`.  (In the Sept 2014 commit, some truncate/extend/pack etc was replaced with uses of `zExtendLSB` and `cExtendLSB`.)

quark17:
> My hypothesis is that, when using `mkDivider(3)` as the model, the timing of put/get is different, and that leads to different result from the non-pipelined module.

The other possibility is that the test is not respecting the interface of the existing module.  It does appear that the current non-pipelined module allows one to put a new request, without taking a response, causing that response to be lost.  But I'd expect that, if the test was losing a response, it would hang -- the rule to get results (from model and dut) would block because the dut has none, and the rule to put requests would block because the model's pipeline has filled with untaken responses.  And we're not seeing that.

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: I've got the beginnings of a test now in my test suite locally.  It has both testing against the mkDivider(1) case, and also comparing against results from the built-in "/" and "%" functions.

The big problem we ran into with only comparing against the "/" and "%" functions is that these are only valid if the input to the divider is all zeroes in the top of the numerator, and this was all that was tested by the previous test.  This begged the question why the numerator is twice the size of the denominator anyway, and we made a version that required one less iteration and zeroed the upper bits on input, and passed the old test.  Unfortunately the floating point divider that imported mkDivider broke.  It turns out that the floating point divider ONLY initializes the top half of the numerator.  Someone surely knows what mathematically is going on there, but it's not a simple divide.  The only way we found to properly test was to compare the full input and output with the mkDivider(1) case.  At any rate, we certainly don't lose anything by making sure that we can do a normal division, even if the primary use of this divider is to do something else.

I'll finish filling out this test and commit tomorrow.

darius-bluespec:
Looking back through the logs, it seems the reason for the non-pipelined modules in Math was to reduce some control logic and registers, as well as resolve some duplicate or conflicting packages and types as compared to the version in contrib.  I didn't review this merge request in detail, but it does seem like it will be unnecessarily increasing at least the registers, compared to the existing implementation.  It's a little bit hard to follow because there seems to be multiple disjoint changes happening in some of the commits, and not all changes are accurately reflected in the commit descriptions.

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: I've now pushed tests to the testsuite in bsc.lib/Divide.  These test mkDivider and mkNonPipelinedDivider at iterations-per-cycle from 1 to 5, both performing division (vs. builtin "/" and "%") and performing whatever happens when you have random bits in the top of the numerator against the "single iteration per cycle" case.

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: I've removed this test, as I've now committed a better one to the testsuite modeled on the SquareRoot test, as suggested.

jonwoodruff:
It's not at all correct to say that this proposed commit is more similar to the version from bsc-contrib (from Cambridge, derived from the existing mkDivider code). The bsc-contrib version preserved flow control from mkDivider which was perhaps not desirable for a NonPipelinedDivider. The adapted version of the mkNonPipelinedDivider in bsc used registers for flow control more in the spirit of an iterative implementation.

This current pull request also uses registers for flow control and is cycle-for-cycle equivalent to the bsc version. (Checked this morning.) There are style changes to reduce redundant state (such as crg_done) and gratuitous registers (rg_d...) which are always read and written together, and so are more cleanly expressed as a struct in a single register. If anyone genuinely believes there are more registers in this design, it would be nice to get some specific analysis.  We could further reduce register use or reduce the number of cycles required, but this would affect timing in difficult-to-predict ways, so I've left it identical to the original.

The only FIFO that was introduced is in the mkNonPipelinedSignedDivider where the signs were previously stored in two Reg#(Bool)s, and these are changed to a FIFO#(Tuple2#(Bool,Bool). This previous design did not enforce its own flow control, but  relied on the server it was wrapping not allowing a new request until the old one was dequeued. To be safe, it would have needed it's own busy/done scheme, at which point you might as well have a FIFO. I've changed this to a FIFO1 to ensure there are no extra registers.

src/Libraries/Base3-Math/Divide.bsv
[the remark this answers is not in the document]
jonwoodruff: I've now moved to rg_index (and count to index in general) to closer align with the original.

jonwoodruff:
In reference to throughput, I've done a bit of a study this afternoon (and updated the tests to pack the pipelines), and the original NonPipelinedDivider had an advantage of a cycle due to concurrent get/put behavior using a CReg.  I've now replicated this behavior in the pull request.

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +117,99 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-   Reg#(UInt#(TLog#(TAdd#(n,1)))) rg_index <- mkRegU;
-
-   rule work(rg_busy);
-      Int#(TAdd#(1,n)) d = rg_d;
-      Int#(TAdd#(1,n)) q = rg_q;
-      Int#(TAdd#(2,m)) r = rg_r;
-      Int#(TAdd#(2,m)) bigd = cExtendLSB(d);
+// non-restoring divider
+// n+3 cycle latency
+module mkNonPipelinedDivider#(Integer s)(Server#(Tuple2#(UInt#(TAdd#(n,n)),UInt#(n)),Tuple2#(UInt#(n),UInt#(n))))
```
quark17, line 122: Can you replace `TAdd#(n,n)` in the interface type with `m`?

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +117,99 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-	       q = (q << 1) | 1;
-	       r = (r << 1) - bigd;
-	    end
-	    else begin
-	       q = (q << 1);
-	       r = (r << 1) + bigd;
-	    end
-	 end
-      end
+   Reg#(DivState#(n)) fReg <- mkRegU;
```
quark17, line 125: Can you call this `rg_f`, to follow the naming convention for other state? (where `fName` is used for FIFOs and `rg_name` is used for registers)

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +117,99 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-
-   rule work(rg_busy);
-      Int#(TAdd#(1,n)) d = rg_d;
-      Int#(TAdd#(1,n)) q = rg_q;
-      Int#(TAdd#(2,m)) r = rg_r;
-      Int#(TAdd#(2,m)) bigd = cExtendLSB(d);
+// non-restoring divider
+// n+3 cycle latency
+module mkNonPipelinedDivider#(Integer s)(Server#(Tuple2#(UInt#(TAdd#(n,n)),UInt#(n)),Tuple2#(UInt#(n),UInt#(n))))
+   provisos(Add#(n, n, m), Alias#(UInt#(TAdd#(TLog#(n),1)), indexT));
```
quark17, line 123: I would tend to put the provisos on separate lines, for readability

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +117,99 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-	 crg_done[1] <= True;
-      end
+   function Bool done(indexT cmp) = (cmp > fromInteger(valueOf(n)));
+   Bool div_done = done(rg_index);
 
-      rg_index <= rg_index + 1;
-      rg_d <= d;
-      rg_q <= q;
-      rg_r <= r;
+   rule work (!div_done);
```
quark17, line 132: The condition of this rule used to come directly from a register.  (Code inside the rule would check the done condition and set the register, in the last execution of the rule.)  You've moved this logic to now be in the rule condition.  Do we know if this matters (does it contribute to the critical path, say)?

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +117,99 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-	 if (r < 0) begin
-	    q = q - 1;
-	    r = r + cExtendLSB(d);
-	 end
-	 UInt#(TAdd#(1,n)) qq = unpack(pack(q));
-	 UInt#(TAdd#(1,n)) rr = cExtendLSB(r);
-
-	 crg_done[0] <= False;
-	 return(tuple2(truncate(qq),truncate(rr)));
+      method ActionValue#(Tuple2#(UInt#(n),UInt#(n))) get if (crg_busy[0] && div_done);
```
quark17, line 165: The computation of the done expression has also been added to this ready condition.  Again, if this were stored in a register, then the logic path here would be shorter (though, again, I don't know if that matters).

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -128,215 +117,99 @@ module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#
...
-      Add#(a__, n, m),
-      Add#(1, m, TAdd#(n, b__))
-      );
+   provisos(Add#(n, n, m));
 
    Server#(Tuple2#(UInt#(m),UInt#(n)),Tuple2#(UInt#(n),UInt#(n))) div <- mkNonPipelinedDivider(s);
-
-   Reg#(Bool) rg_asign <- mkRegU;
-   Reg#(Bool) rg_bsign <- mkRegU;
+   FIFO#(Tuple2#(Bool,Bool)) fSign <- mkLFIFO;
```
quark17, line 184: Can you explain why we can't rely on `div` for flow control?  I have no objection changing this to a FIFO though.

testsuite/bsc.lib/Divide/divide.exp
```diff
@@ -0,0 +1,25 @@
+# Tests for the Divide package
+
+# These tests can't use 'test_c_veri_bsv' because that attempts to
+# reuse the Verilog backend's .ba files for Bluesim, and that can't
+# be done for these tests, since they use the ConstrainedRandom
+# package, which has different implementations for the two backends.
+
+proc do_test { name } {
+    test_c_only_bsv     $name  sys$name.c.out.expected
```
quark17, line 9: In your tests, the Bluesim and Verilog sim outputs are the same, so there's no need to have separate files named `.c.out.expected` and `.v.out.expected`.  Just have one file named `sys$name.out.expected`.

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: I've replicated this behavior now.  Given all the wide subtractions and things we're doing, it's not likely to be a big factor, but it doesn't hurt.

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -84,19 +78,14 @@ module mkDivider#(Integer s)(Server#(Tuple2#(UInt#(m),UInt#(n)),Tuple2#(UInt#(n)
...
-      );
+module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(TAdd#(n,n)),Int#(n)),Tuple2#(Int#(n),Int#(n))))
+   provisos(Add#(n, n, m));
 
    FIFO#(Tuple2#(Int#(m),Int#(n))) fRequest <- mkLFIFO;
    FIFO#(Tuple2#(Int#(n),Int#(n))) fResponse <- mkLFIFO;
 
    Server#(Tuple2#(UInt#(m),UInt#(n)),Tuple2#(UInt#(n),UInt#(n))) div <- mkDivider(s);
-   FIFO#(Tuple2#(Bool,Bool)) fSign <- mkLFIFO;
+   FIFO#(Tuple2#(Bool,Bool)) fSign <- mkSizedFIFO(valueOf(TAdd#(n,2)));
```
quark17, line 88: The argument to `mkSizedFIFO` is the maximum size needed, regardless of the value of `s`.  Could we compute the actual space needed, using `s`?  (I assume that it'd involve dividing `n` by `s` but rounding up, and then still adding the 2, which I guess is to account for the put/get on either end?)

quark17 (commented):
I just wanted to check in here.  Aside from a few tweaks mentioned before and some I noticed just now, the code looks about ready, except that the testsuite is failing:
```
Running ./divide.exp ...
FAIL: `sysTest_mkNonPipelinedDivider.c.out' differs from `sysTest_mkNonPipelinedDivider.out.expected'
FAIL: `sysTest_mkNonPipelinedDivider.v.out' differs from `sysTest_mkNonPipelinedDivider.out.expected'
```
I can look into this, unless you are already doing that?

Another small thing I noticed is that on lines 45, 135, and 193, the type `TAdd#(n,n)` is explicitly written, where `m` could have been used.  It's fine either way, but I wanted to mention, in case you prefer using `m`.

src/Libraries/Base3-Math/Divide.bsv
```diff
@@ -84,19 +78,14 @@ module mkDivider#(Integer s)(Server#(Tuple2#(UInt#(m),UInt#(n)),Tuple2#(UInt#(n)
...
 endmodule
 
-module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(m),Int#(n)),Tuple2#(Int#(n),Int#(n))))
-   provisos(
-      Mul#(2, n, m),
-      // per request of bsc
-      Add#(a__, n, m),
-      Add#(1, m, TAdd#(n, b__))
-      );
+module mkSignedDivider#(Integer s)(Server#(Tuple2#(Int#(TAdd#(n,n)),Int#(n)),Tuple2#(Int#(n),Int#(n))))
```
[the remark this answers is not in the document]
jonwoodruff: I've now replaced the last two instances where TAdd#(n,n) was replaceable with m.

src/Libraries/Base3-Math/Divide.bsv
jonwoodruff: Good catch!  I've done this now (over 3 commits).  I also messed up the +2 logic the first time, so thanks for recording your thinking here as I was able to catch it in the end.

jonwoodruff:
> I just wanted to check in here. Aside from a few tweaks mentioned before and some I noticed just now, the code looks about ready, except that the testsuite is failing:
> 
> Running ./divide.exp ...
FAIL: `sysTest_mkNonPipelinedDivider.c.out' differs from `sysTest_mkNonPipelinedDivider.out.expected'
FAIL: `sysTest_mkNonPipelinedDivider.v.out' differs from `sysTest_mkNonPipelinedDivider.out.expected'

There was a flow control bug in NonPipelinedDivider; not sure how it snuck through.  This bug is fixed now.

Also, I've made a pass at factoring out shared logic between the two implementations so there should be a lot easier to fix/maintain in the future.

jonwoodruff:
@quark17 Many thanks for the tutorial on how to merge with upstream!  I've done this now.

testsuite/bsc.lib/Divide/Test_mkDivider.bsv
```diff
@@ -26,7 +26,9 @@ interface TesterIfc #(numeric type n);
 endinterface
 
 module mkTester #(Integer maxstages, UInt #(32) numtests) (TesterIfc #(n))
-  provisos (Add#(n, n, m), Add#(1, m, TAdd#(n, a__)));
+  provisos (Add#(n, n, m),
+            Add#(1, m,
+            TAdd#(n, a__)));
```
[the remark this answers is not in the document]
jonwoodruff: I've done this now. You're right, the second proviso wasn't needed at all.

quark17:
Apologies again for the delay.  I need to review your latest changes (factoring out shared logic) and then I'll merge it.  Is it OK if I merge it as one squashed commit on `main`?

jonwoodruff:
Squashing the commit is probably a good idea to avoid a confusing pile of commits in the history.

quark17:
Oh shoot.  I wasn't paying attention and did a rebase merge rather than squash.

PR #411, opened 2021-09-13
Support building on Arm-based Macs

jrtc27:
No substantive changes needed other than yices's build system needing yet more
hand-holding.

This also includes a fix to avoid needing to delete Homebrew's tcl-tk's tcl.h,
which hopefully also works on Intel-based Macs.


jrtc27:
It turns out that the tcl.h deleted during CI comes from the pre-installed copy of R in the image; Homebrew deliberately doesn't link tcl-tk into default search paths to avoid shadowing the system copy (a so-called "keg-only" package) but R just does what it wants and happily breaks things. Not sure if there's a good way to work around that (beyond filing a bug report against R to not break the system), but I don't use R so not something I care about and this is sufficient to be able to build bsc myself.

quark17:
Thanks!

This PR adds a workaround for an issue with the Yices build.  Is that issue fixed at a later point in the Yices repo?  Would the workaround still be needed if we updated the submodule to a later point?

PR #260 includes an update of the Yices submodule.  I'd like to merge that PR, as soon as the remaining issues there are addressed.

PR #260 also unifies the determination of the Tcl include directory into `platform.sh` (and then has the two Makefiles call that, to set TCL_INC_FLAGS).  The computation there does seem to correctly determine that TCL_INC_FLAGS should be empty, although it could be made more explicit.

jrtc27:
> Thanks!
> 
> This PR adds a workaround for an issue with the Yices build. Is that issue fixed at a later point in the Yices repo? Would the workaround still be needed if we updated the submodule to a later point?
> 
> PR #260 includes an update of the Yices submodule. I'd like to merge that PR, as soon as the remaining issues there are addressed.

I don't believe any of the workarounds become obsolete. Looking at the large diff, almost all the changes are non-buildsytem things, with the build system changes just being minor additions, nothing relevant to the workarounds, unfortunately.

> PR #260 also unifies the determination of the Tcl include directory into `platform.sh` (and then has the two Makefiles call that, to set TCL_INC_FLAGS). The computation there does seem to correctly determine that TCL_INC_FLAGS should be empty, although it could be made more explicit.

Chasing through the rather more complicated way in which TCL_INC_FLAGS gets computed there I agree it looks like all the checks will fail so it will end up being empty, but as you point out it's rather hard to reason about and will currently exit with a non-zero exit code (which, curiously, gets ignored in `$(shell ...)`...) so isn't quite sufficient in its current form.

PR #412, opened 2021-09-16
Tweak INSTALL.md to use bluetcl for version information

omasanori:
`git describe --tags`  lets us avoid versioning on installation by hand.


quark17:
Thank you for the suggestion!  I wonder if a better way would be this:
```
BSC_VERSION=$(echo 'puts [lindex [Bluetcl::version] 0]' | inst/bin/bluetcl)
```
Git may not always be available (if the build was in an exported snapshot of the source, for example) and this command asks the BSC installation itself to provide the version.

omasanori:
Thank you for your suggestion, @quark17! I was unaware of the way to print the version with built programs. Updated.

PR #430, opened 2021-11-06
Fix lookup of Tcl via pkg-config

thotypous:
This fixes a regression introduced by PR #427.

Build is currently broken on Arch Linux and possibly other distros. The Tcl package may not have a version suffix.


quark17:
Thanks!

The CI failures in Ubuntu build might to be due to newer GHC version in those VMs.  I will try building with newer GHC, to see if there are flag or code changes needed.

PR #448, opened 2022-02-02
Fix FloatingPoint modules to not require aggressive conditions

### LINKED ISSUE #271: FloatingPoint::mkFloatingPointDivider deadlocks when dividing zero by something.
FloatingPoint::mkFloatingPointDivider deadlocks when dividing zero by something.  Looking at the code and the VCD file, it appears that `s1_stage` detects the zero and sets `out` to a valid value.  `s2_stage` checks for the valid `out` and forwards it without triggering the integer divider.  But then `s3_stage` never fires for reasons I've been unable to unearth.  I'm new enough to Bluespec SystemVerilog that my ability to debug runs out there.  

```
package DivTB;

import FloatingPoint::*;
import Divide::*;
import ClientServer::*;
import GetPut::*;

(* synthesize *)
module mkDivTB();

   Server#(Tuple2#(UInt#(56),UInt#(28)),Tuple2#(UInt#(28),UInt#(28))) int_divider <- mkDivider(1);

   Server#(Tuple3#(Float, Float, RoundMode), Tuple2#(Float,Exception)) fp_divider <- mkFloatingPointDivider(int_divider);

   Reg#(Bool) issued <- mkReg(False);

   rule rl_issue(!issued);
      // works:
      //fp_divider.request.put(tuple3(1, 1, defaultValue));
      // deadlocks:
      fp_divider.request.put(tuple3(0, 1, defaultValue));
      issued <= True;
   endrule

   rule rl_retire;
      let r <- fp_divider.response.get();
      $display("result:", fshow(r));
      $finish;
   endrule

   Reg#(UInt#(12)) watchdog <- mkReg(0);
   rule rl_watchdog;
      watchdog <= watchdog + 1;
      if (watchdog == maxBound) begin
         $display("watchdog hit");
         $finish;
      end
   endrule

endmodule

endpackage
```

quark17:
This fixes #271.

The divide and square root modules, if compiled with conservative implicit condition lifting (the default), would deadlock when given invalid inputs: the first stage would detect the error and determine the result; such that the second stage didn't need to dispatch a request to the integer divider/squarerooter; but the third stage would expect a response, as part of its implicit conditions.  The fix here is to manually split the third stage into two rules: one expecting a response and one not.  It would have been simpler to insert a `split` attribute to have BSC split the rule automatically, but BSC lifts the aggressive conditions before splitting, so that doesn't help. If rule attributes were supported for specifying condition lifting for an individual rule that would also be a simpler fix here.  The modules did have an `options` attribute attached to them, but BSC ignores these on modules that are not separately synthesized.

Testcases are added.

### DIFF
--- a/src/Libraries/Base3-Math/FloatingPoint.bsv
+++ b/src/Libraries/Base3-Math/FloatingPoint.bsv
@@ -2278,7 +2278,6 @@ endmodule: mkFloatingPointMultiplier
 ////////////////////////////////////////////////////////////////////////////////
 /// Floating point divider
 ////////////////////////////////////////////////////////////////////////////////
-(* options = "-aggressive-conditions" *)
 module mkFloatingPointDivider#(Server#(Tuple2#(UInt#(nbits),UInt#(dbits)),Tuple2#(UInt#(dbits),UInt#(dbits))) div)(Server#(Tuple3#(FloatingPoint#(e,m), FloatingPoint#(e,m), RoundMode), Tuple2#(FloatingPoint#(e,m),Exception)))
    provisos(
       Add#(e,1,ebits),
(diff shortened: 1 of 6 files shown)


quark17:
> If rule attributes were supported for specifying condition lifting for an individual rule that would also be a simpler fix here.

I'm not sure why I said this, because there is an attribute `aggressive_implicit_conditions` that can be applied to rules.

Issue #304, opened 2021-02-03
Generate appropriate empty argument lists for Verilog system tasks/functions


quark17:
When system tasks appear in the generated Verilog, this code in `Verilog.hs` omits the parenthesis when there are no arguments:
```
-- no parens when calling a task if it has no arguments                                                                                   
pPrint d p (VTask task []) = pPrint d 0 task <> text ";"
```
This was probably added because that is how the Verilog standard says that `$finish` should be written.  However, for `$fflush`, the standard says that it must have empty parentheses, and some tools will be sticklers and issue a warning if the parentheses are missing.

The `pPrint` code will probably need to check the task name, to handle each specifically.

PR #455, opened 2022-02-11
 Generate appropriate empty argument lists for Verilog system tasks/functions

benycze:
Hello,

I want to help with BSC development and I am starting with my first good first issue. This simple change is trying to address the issue #304.

I am aware just of the `$fflush` function which needs parens even if no argument is provided (as it is mentioned in the issue description).

Do you want me to add/correct something else (test, etc.)?

### DIFF
--- a/src/comp/Verilog.hs
+++ b/src/comp/Verilog.hs
@@ -513,6 +513,7 @@ instance PPrint VStmt where
         pPrint d p (Vdumpvars level vars) = text "$dumpvars(" <> sepList dvargs (text ",") <> text ");"
             where dvargs = (pPrint d 0 level):(map (pPrint d 0) vars)
 -- no parens when calling a task if it has no arguments
+        pPrint d p (VTask task []) | (getVIdString task) == "$fflush" = pPrint d 0 task <> text "();"
         pPrint d p (VTask task []) | isTaskVId task = pPrint d 0 task <> text ";"
         pPrint d p (VTask task es) = pPrint d 0 task <> text "(" <> commaList d es <> text ");"
 


quark17:
Thank you!  Your interest in contributing to BSC is appreciated!  I apologize for the delay in responding.

This line already prints empty parentheses (when `es` is empty):
```
pPrint d p (VTask task es) = pPrint d 0 task <> text "(" <> commaList d es <> text ");"
```
The line before it handles the exceptions:
```
pPrint d p (VTask task []) | isTaskVId task = pPrint d 0 task <> text ";"
```
So instead of adding a third line (exceptions to the exception), I would just change the condition on the exception line:
```
pPrint d p (VTask task []) | isTaskVId task && not (taskRequiresParens task) = pPrint d 0 task <> text ";"
```
Or combine it into a single function, if you can think of a good name for it:
```
pPrint d p (VTask task []) | isTaskWithNoEmptyParens task = pPrint d 0 task <> text ";"
```

At the moment, I am only aware of `$fflush`.  My hope was that someone would go look at the Verilog or SystemVerilog spec and check the grammar for every system task, and report back whether there are any other such tasks.  For example, in the IEEE 1800-2017 spec, Section 21 is about tasks, and it includes grammar like this:
```
file_flush_task ::= $fflush([mcd|fd]) ;
```
The brackets indicate which part is optional, and they are around the argument inside the parens, so the parens are not optional.  In contrast, there is this grammar:
```
simulation_control_task ::=
    $stop [ ( n ) ] ;
  | $finish [ ( n ) ] ;
  | $exit [ ( ) ] ;
```
This has the brackets around the parentheses for `$finish`, so there are no parens when the task has no argument.  For `$exit`, either way is allowed, but there is no difference in behavior, so we can treat either way, as convenient.  For some tasks, there is never any parentheses:
```
time_function ::=
    $time
  | $stime
  | $realtime
```
If you're not able to look through the spec, it would be fine to just except `$fflush` for now, and we can add more later (if they are discovered).  It just means that the bug should remain open, until someone does the work of confirming that all the cases are handled.

Also, FYI, you've edited the place where system _tasks_ are printed, but there are also system _functions_ (that appear as expressions, not statements) and those are printed by this code (elsewhere in `Verilog.hs`):
```
pPrint d p (VEFctCall f []) | isTaskVId f = pPrint d 0 f
pPrint d p (VEFctCall f es) = pPrint d 0 f <> text "(" <> commaList d es <> text ")"
```
It might be worth changing the guard on this too, perhaps to a separate function `isFunctionWithNoEmptyParens`?  At the moment, I'm not aware of any functions that would be excluded.  (As `$fflush` is a task.)

And yes, it would be good to include a test.  I would add a test example in `testsuite/bsc.verilog/tasks/` that generates Verilog containing the system task/functions that can take no arguments, and then set up the test script (the `*.exp` file) to look at the generated Verilog and confirm that each has parens or no parens as appropriate (probably by using a `grep` or `regexp` function -- see, for example, how the regexp is used for `ForeignGroupTest.bs` at the end of the script).  The text example could be named something that indicates that it is for testing empty parens in the ouput.  I see that there are already test files `EmptyTasks.bsv` and `EmptyTasksWorkaround.bsv` that are testing for parens in the BSV source (and only for the sim control tasks).  So pick another name, like `EmptyArgumentsInVerilog.bsv` or `NoArgsInVerilog.bsv`.

I also notice that there are test directories `bsc.misc/format/` and `bsc.misc/fwrite/` which are testing system tasks, and should maybe be consolidated with the others.  But I would ignore those for now, and just work in `bsc.verilog/tasks/`.  (It's also unclear if `bsc.verilog/` is the right place for these tests, since they are not specific to the Verilog backend.)

PR #461, opened 2022-03-04
Makefile: Build bluetcl with RTSFLAGS

bpfoley:
All the other binaries are built with RTSFLAGS, so for consistency
we should build bluetcl with it too.

It gives us the (probably not hugely useful) ability to run bluetcl
with -rtsopts, but also gives us a place to inject linker flags and
libraries that are only valid when linking the final executable.
(This happens to be necessary for linking static binaries at Google
due to a slightly unusual build of GHC here).

### DIFF
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -342,7 +342,7 @@ showrules: $(SOURCES)
 bluetcl: bluetcl.hs bluetcl_Main.hsc $(SOURCES)
 	$(PREBUILDCOMMAND)
 	$(BUILDCOMMAND) $(BUILDFLAGS) -c
-	$(BUILDCOMMAND) $(BUILDFLAGS) $(BLUETCL_BUILDLIBS) \
+	$(BUILDCOMMAND) $(BUILDFLAGS) $(BLUETCL_BUILDLIBS) $(RTSFLAGS) \
 		-o $@ \
 		-no-hs-main \
 		-x c bluetcl_Main.hsc


quark17:
My understanding is that we were intentionally not providing it because, when we had been, we got a warning that it was being ignored -- since the top-level of BSC is not Haskell.  The way that these values are passed in is via a C file, as indicated in the comment above the line:
```
# If you modify this, also update rts_opts in ../vendor/htcl/haskell.c
RTSFLAGS = "-with-rtsopts=-H256m -K10m -i1"
```
And I do see this in that file:
```
conf.rts_opts = "-H256m -K10m -i1";
```
I also see that `bluetcl_Main.hsc` makes a call to `htcl_initHaskellRTS`, which is where that code gets executed.

Are you saying that you don't use `bluetcl_Main.hsc` for your build, and therefore need to do something else?

quark17:
If you're just using the RTS variable to provide additional flags to the build command that are unrelated to RTS, then I'd rather find a clean way for you to do it.  If necessary, create a new variable `USER_BUILDLIBS` to put on all the command lines.  Although, if this is for providing flags to do a static link, then maybe just add that feature to the Makefile (by adding the necessary flags to the build flags, or providing hooks that have `STATIC` in the name)?

quark17:
I left this open, in case you disagreed with my assessment.  There's been no response, so I'm closing it as not accepted (due to not being the right way to implement the possible features that it could be used for).

PR #463, opened 2022-03-31
doc: remove wording that suggests BRAMCore is closed-source.

danderson:
I suspect this is a leftover that predates the open-sourcing of bsc and the base libraries. Both the Verilog and Bluesim source for BRAMCore seem to be available now.

Signed-off-by: David Anderson <dave@natulte.net>


quark17:
Thank you!

Issue #464, opened 2022-03-31
Docs: clarify that BRAM modules can do 1-cycle operations


danderson:
This is a minor suggestion for improving the library reference documentation. I'm happy to send a PR to implement it, but first I wanted to make sure my proposal makes sense.

I'm still a relative newbie to BSV and the libraries. Reading the documentation for the BRAM library, it wasn't clear to me at a glance that the modules support 1-cycle accesses. In particular, the wording of the docs for the `BRAM_Configure` struct implied to me that BRAM modules had a minimum access time of 3 cycles, because `latency` can be 1 or 2 depending on data registration, and then `outFIFODepth` adds another 2 to that.

What I'd missed, of course, is that the BRAM's output FIFO is a `BypassFIFO`, so as long as the reader keeps up, the FIFO depth is irrelevant and the latency is dictated purely by whether I choose to register the output data.

I was further confused because the `BRAM_Configure` struct has another field that mentions bypassing, `allowWriteResponseBypass`. So, naively, I assumed that "bypass FIFO" was in reference to this write bypassing behavior.

I'm sure none of this is confusing to a more experienced hardware designer, especially one already familiar enough with BRAM design, or intimate enough with the Bluespec libraries to recognize that BypassFIFO is a reference to the SpecialFIFOs package. But it took me more time than I'd like to admit, and digging into BRAM's source code, to convince myself of the actual behavior.

My proposal would be to edit the BRAM library documentation slightly:
 - Explicitly state that the `latency` field of `BRAM_Configure` completely defines the access latency, and `outFIFODepth` only matters when faced with slow readers
 - Cross-reference mentions of `BypassFIFO` to the `SpecialFIFOs` package, to make it more obvious that an unfamiliar reader should skip forward in the reference doc and go understand what a `BypassFIFO` does (this is admittedly a small lie, because BRAM privately implements its own bypass FIFO, I assume for historical reasons - but afaict, the semantics are identical to the bypass FIFOs in `SpecialFIFOs`)

Do you think these changes would make sense?

quark17:
Thank you!  And, yes, this looks good.

[...]

PR #467, opened 2022-04-08
BRAM: replace custom bypass FIFO adapter with call to SpecialFIFOs.

danderson:
Checked with:

```
make release
make check-smoke
(cd testsuite/bsc.lib/BRAM && make localcheck)
```

I currently don't have dependencies set up quite right to make it through the full `make check-suite`, but hopefully the above is enough to have some confidence that the SpecialFIFOs behavior is the same as the previous bypass adapter.

Updates #464

### DIFF
--- a/src/Libraries/Base3-Misc/BRAM.bsv
+++ b/src/Libraries/Base3-Misc/BRAM.bsv
@@ -6,6 +6,7 @@ import ClientServer ::*;
 import FIFOF :: *;
 import DReg :: *;
 import GetPut :: *;
+import SpecialFIFOs :: *;
 
 //  Export  section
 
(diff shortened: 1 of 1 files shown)


danderson:
Ran the full suite overnight. The only failures were CHILDPROC errors in the bsc.interra suite, but the failures gave no more information than that. I suspect this is just something else missing in my dependency stack to make these tests work?

quark17:
Thanks!  I held off merging because the macos CI was failing -- due to VM issues, nothing to do with this PR -- and I wanted to resolve that first.  It looks like they're passing now on the main branch, without me having to do anything, so I'll merge this now.

The `mkBypassFIFOAdapter` function could have been moved to the `SpecialFIFOs` library, if we think it might be generally useful to people. But since we don't have any tests for it (and I haven't checked that it has the right scheduling, for example), I think it's OK to just eliminate it (and not create any maintenance burden by keeping it).

PR #482, opened 2022-06-06
Fix build error after upgrade yices to 2.6.4

chenm001:
Fix build issue on Linux platform.
It is not good idea to modify bsc Makefile, but Yices is upstream submodule, so I want not to modify it.

### DIFF
--- a/src/vendor/yices/Makefile
+++ b/src/vendor/yices/Makefile
@@ -7,6 +7,7 @@ PREFIX?=$(TOP)/inst
 
 VER_MAJ = 2
 VER_MIN = 6
+PATCH_LEVEL = 4
 VERSION = v$(VER_MAJ).$(VER_MIN)
 
 all: install
@@ -15,6 +16,7 @@ ifeq ($(OSTYPE), Darwin)
 SNAME=libyices.$(VER_MAJ).dylib
 else
 SNAME=libyices.so.$(VER_MAJ).$(VER_MIN)
+SNAME_FULL=libyices.so.$(VER_MAJ).$(VER_MIN).$(PATCH_LEVEL)
 endif
 
 install:
@@ -22,6 +24,9 @@ install:
 	ln -fsn $(VERSION)/include
 	ln -fsn $(VERSION)/lib
 	ln -fsn $(VERSION)/include_hs
+ifneq ($(OSTYPE), Darwin)
+	ln -fs  $(SNAME_FULL) lib/$(SNAME)
+endif
 	install -m 755 -d $(PREFIX)/lib/SAT
 	install -m 644 lib/$(SNAME) $(PREFIX)/lib/SAT
 


quark17:
Thank you for contributing!  I have some questions, though, because I'm not sure I understand the changes.

Can you explain the build problem you are having?  I can build on Ubuntu and Debian Linux, with yices 2.6.4.  Are you building on a different Linux distribution?

The BSC install process creates a directory `INSTALLDIR/lib/SAT/` and places the file `libyices.so.2.6` in that directory.  It seems like you are suggesting that we add a symlink from the name `libyices.so.2.6.4` to `libyices.so.2.6`, in case linking is looking specifically for that patch version.  In my builds on Debian and Ubuntu, the `bsc` binary does not specify the patch level.  I can see this by using `ldd` to display the name that it is looking for.  Do you have a binary that needs to link with the full name?

Your PR isn't putting the symlink in the BSC installation directory, though.  It's putting the symlink in the build directory.  I would expect this to have no effect, because nothing is looking for libraries in that directory.  Are you sure that the command is correct?

The Makefile in `bsc/src/vendor/yices/` intentionally doesn't mention the patch level (only the major and minor versions), because it assumes that the Yices C API will not change from patch to patch -- that a change in the API would require a bump in the minor version at least.  We can revisit that if the assumption is invalid.

quark17:
Thank you for contributing!  I have some questions, though, because I'm not sure I understand the changes.

Can you explain the build problem you are having?  I can build on Ubuntu and Debian Linux, with yices 2.6.4.  Are you building on a different Linux distribution?

The BSC install process creates a directory `INSTALLDIR/lib/SAT/` and places the file `libyices.so.2.6` in that directory.  It seems like you are suggesting that we add a symlink from the name `libyices.so.2.6.4` to `libyices.so.2.6`, in case linking is looking specifically for that patch version.  In my builds on Debian and Ubuntu, the `bsc` binary does not specify the patch level.  I can see this by using `ldd` to display the name that it is looking for.  Do you have a binary that needs to link with the full name?

Your PR isn't putting the symlink in the BSC installation directory, though.  It's putting the symlink in the build directory.  I would expect this to have no effect, because nothing is looking for libraries in that directory.  Are you sure that the command is correct?

The Makefile in `bsc/src/vendor/yices/` intentionally doesn't mention the patch level (only the major and minor versions), because it assumes that the Yices C API will not change from patch to patch -- that a change in the API would require a bump in the minor version at least.  We can revisit that if the assumption is invalid.

chenm001:
Please allow me explain more details.
First at all, I build on RHEL8 platform.

```
@src/vendor/yices/Makefile
SNAME=libyices.so.$(VER_MAJ).$(VER_MIN)
install -m 644 lib/$(SNAME) $(PREFIX)/lib/SAT

```
it try to use **libyices.so.2.6**

```
@src/vendor/yices/v2.6/yices2/Makefile
MAJOR = 2
MINOR = 6
PATCH_LEVEL = 4

YICES_VERSION = $(MAJOR).$(MINOR).$(PATCH_LEVEL)

@src/vendor/yices/v2.6/yices2/Makefile.build
install-linux install-unix: install-default
	$(LDCONFIG) -n $(DESTDIR)$(libdir) && (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(YICES_VERSION) libyices.so)

```
It symbol link  **libyices.so** to **libyices.so.2.6.4**

In here, I got error message that **libyices.so.2.6** does not found.
In my view, we have two method to fix, one is modify upstream yices project to make extra libyices.so.2.6, the other way is modify bsc to use libyices.so.2.6.4, my workaround use second way.

quark17:
Yes, the link for `libyices.so.2.6` is created by the call to `ldconfig` in this target:

>  ```
> @src/vendor/yices/v2.6/yices2/Makefile.build
> install-linux install-unix: install-default
> 	$(LDCONFIG) -n $(DESTDIR)$(libdir) && (cd $(DESTDIR)$(libdir) && $(LN_S) -f libyices.so.$(YICES_VERSION) libyices.so)
> ```

If I run the command, and add `-v` to turn on verbose output, then I see that it creates the link (on Debian and on RHEL7) (and where `lib` is `bsc/src/vendor/yices/v2.6/yices2-inst/`):
```
$ /sbin/ldconfig -v -n lib
lib:
	libyices.so.2.6 -> libyices.so.2.6.4 (changed)
```

Can you confirm that you can run `ldconfig` and that it does not create the link on RHEL8?  I will try to find a RHEL8 system that I can test it on, myself.

quark17:
A build on Centos 8-Stream earlier this year was fine.  It's not the same as RHEL 8, but I don't see why `ldconfig` it wouldn't work on RHEL.  And if BSC was building fine for you on RHEL 8 with older Yices, the upgrade to 2.6.4 wouldn't have introduced the problem (because the Makefile has used `ldconfig` for years).  If you did `git pull` in the repo, but didn't clean out the old build files, perhaps try removing them and starting with a fresh repo?

PR #487, opened 2022-07-19
Makefile: Add quotations around variables

omasanori:
It makes the build recipes robust against environment variables containing white spaces. (https://github.com/B-Lang-org/bsc/pull/408, take two)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -28,25 +28,25 @@ help:
 
 .PHONY: rem_inst
 rem_inst:
-	rm -fr $(PREFIX)
+	rm -fr "$(PREFIX)"
 
 .PHONY: rem_build
 rem_build:
-	rm -fr $(BUILDDIR)
+	rm -fr "$(BUILDDIR)"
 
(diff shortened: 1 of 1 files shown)


quark17:
I apologize for getting your hopes up.  I investigated this issue and it appears that `make` cannot (and will not) support spaces (or colons) in filenames.  See issue 712 in the GNU database: https://savannah.gnu.org/bugs/?712

Your changes only address the top-level Makefile.  For BSC to build in a directory with space in the path, we would need to add quotes to each of the Makefiles in the subdirectories that are recursively entered.  If you try this on the first subdirectory, `src`, you find that the path is used in an `include` statement, and this causes a problem.  I could not find a way to quote the path in that statement.

I think we just have to accept that it is a limitation of the current build process, that it cannot be performed in a directory whose absolute path includes a space or colon.  You'll have to do the building in a different directory.  We can document this in the INSTALL instructions.

omasanori:
Thank you for your insightful investigation!

PR #490, opened 2022-07-26
First cut at incorporating BH Language Ref Guide into bsc/doc

rsnikhil:
... good enough as a baseline for 2022-07 release.


kenta2:
PR #492 includes everything in this PR, plus more things that make build, install, and release work better, so this PR can probably be closed without merging.

PR #492, opened 2022-08-11
Bluespec Haskell guide with fixes for install and release

kenta2:


kenta2:
This PR subsumes #490, so #490 can be deleted.

PR #496, opened 2022-08-12
Fix typo

pbing:


kenta2:
I have verified that, in `src/Libraries/Base1/Clocks.bs`, the method is called `setGateCond`, so am Merging the PR.

PR #510, opened 2022-10-28
add support for haskell language server

ThePerfectComputer:
Add support for haskell language server with proper hie.yaml generator, unit tests, and documentation.

I've begun poking through ``src/com`` and things go faster when jump to definition works and when I have access to type inference on hover.

I imagine this would help others also getting started reading bluespec sources.


ThePerfectComputer:
seems to fail on CI only because of MacOS 10.15 brownouts.

quark17:
Thank you for this contribution!  It is very clean and I'm happy to merge it!

I do wonder whether there's a better way to put it into the documentation.  You've added a section under "Using the Bluespec compiler" for "Editors", and that was definitely needed, as a place to mention the editor extensions for BH/BSV.  However, the HLS for the sources are for developers and not users, so I wonder if it's worth mentioning that separately, maybe in a new "Developing the Bluespec compiler" section.  (The util directory also contains other subdirectories that could be of interest to users and one for building.)  I don't have any good suggestion on what specifically to change, so I'm also happy to accept it as it is, unless you have any ideas.

ThePerfectComputer:
I'm currently working on the language server for BSC, so I'll have a more concrete idea of any suggestions W.R.T to a separate develop.md like file in the future. But right now, the "Editors" sections seems small enough to not be too problematic W.R.T being inside Install.md

PR #516, opened 2022-11-17
docs: have Cntrs.tex reference mkCount

Aaron-Hartwig:
While the code samples are correct in using `mkCount`, there are a handful of references to `mkCounter` (which does not exist) in the `Cntrs` section of the BSV reference documentation.

### DIFF
--- a/doc/libraries_ref_guide/LibDoc/Cntrs.tex
+++ b/doc/libraries_ref_guide/LibDoc/Cntrs.tex
@@ -41,6 +41,9 @@ comparison operations.
 
 {\bf Interfaces and Methods}
 
+\index{Count@\te{Count} (interface)}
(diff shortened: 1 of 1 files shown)


quark17:
Thank you for contributing a fix!  I think there was some confusion because there also exists a `Counters` package (under `src/Libraries/Base1/`), which does export `mkCounter`.  This library (in `Base3-Misc`) seems to be a newer API for the same underlying Verilog module.  So perhaps there was some copy-paste issues in creating the documentation for the new version.

I noticed that even though you corrected `mkCount` in one of the `\index` statements, the statement still had `Counters` as the package name.  That needs to be corrected to `Cntrs` too.  I've pushed a commit that makes that change, but it also adds some indexing for the other module and interfaces -- which were not being added to the index.

Anyway, thank you again!

PR #521, opened 2022-12-07
export getEvalPosition

pwang7:


quark17:
This defines `getEvalPosition`, but doesn't add it to the export list.  I assume you'll also want to add it, to line 56 or insert a new line 57.

pwang7:
> This defines `getEvalPosition`, but doesn't add it to the export list. I assume you'll also want to add it, to line 56 or insert a new line 57.

Just added.

PR #527, opened 2022-12-30
Update .gitattributes

ThePerfectComputer:
some basic highlighting for bluespec haskell code

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,4 @@ doc/**  linguist-documentation
 # with the info they need
 #
 src/comp/update-build-version.sh export-subst
+*.bs linguist-language=haskell


quark17:
I'm guessing that you're addressing the fact that, if you view a file like `Prelude.bs` on GitHub, it isn't highlighted.  GitHub uses a tool called Linguist to detect the language of files in the repo, and the result of that is used for a number of things, including deciding whether to apply a syntax highlighter.  Your PR tells Linguist that all `*.bs` files are Haskell files, and so it uses the Haskell syntax highlighter (which mostly works for BH, though not perfectly).  This is lying to Linguist (and may have consequences aside from highlighting), so my inclination is to reject this.  The correct way to fix syntax highlighting is to add BH to Linguist's list of known languages (and the highlighters for those languages).

Linguist already knows about BSV.  See, for example, line 642 in [`lib/linguist/languages.yml`](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) in the `github/linguist` repo.  I would suggest filing a PR to change that file to include BH.  (I'm not sure if BSV and BH should be separate entries or one entry.  Often it'd be better for GitHub to report that a repo is some percent "Bluespec" code, rather than X% BSV and Y% BH.  But if they are considered one language, then you might need a single highlighter that is smart enough to detect which highlighting to use.)

Detecting the language is just the first step of what GitHub does .  Its second step is to lookup the detected language in a table to find the syntax highlighter for that language.  Again, GitHub already knows of a syntax highlighter for BSV.  A list of the highlighters is given in [`vendor/README.md`](https://github.com/github/linguist/blob/master/vendor/README.md) in the LInguist repo.  You'll see that "Bluespec" is listed and the highlighter that is used is one maintained by @thotypous.  This support was added in 2015 with [this PR](https://github.com/github/linguist/pull/1945) in the Linguist repo.

I notice that in @thotypous 's PR, there was this change:
```
-  tm_scope: source.verilog
+  tm_scope: source.bsv
```
I believe this means that, before the change, Linguist and GitHub treated BSV files as if their grammar was Verilog.  And so, it may not be necessary to have a BH syntax highlighter, to add BH to Linguist.  The `tm_scope` field for BH could be `source.haskell`, for now.

So that's my suggestion: file a PR on the Linguist repo, adding BH (declaring it as Haskell grammar for now, if needed).

I see that, in June 2022, a language called Bikeshed was added to Linguist, with `.bs` as its extension.  And GitHub is now reporting that the BSC repo is 6.1% Bikeshed.  So it'll be good to get BH added to Linguist, and fix that in the language detection for our repos.  I don't think it'll be a problem for the same suffix to be used for multiple languages -- because I see that `.bs` is also listed for a language called BrighterScript.  Hopefully, the grammar is used to try to disambiguate.  If not, we can always add a line to `.gitattributes` specifying BH for all `*.bs` files (like you've done in this PR, but telling it the right language).

Does that seem reasonable?

ThePerfectComputer:
Seems reasonable. Something to add to my ever growing list of TLC needed for BSC.

quark17:
Ok, I will close this PR and open an issue (for adding BH to Linguist).

The first step (which gets syntax highlight) is just a matter of submitting a 6-line patch to Linguist (adding an entry for BH with values for a few fields that Linguist needs).  It just requires figuring out what the field values should be.  I might go read the header comment that explains the fields and see if I can submit something.

PR #538, opened 2023-01-22
update front page README for latest release

kenta2:


quark17:
It looks like `latest` works.  You just had `download` and `latest` swapped in order in your link.  If I switch them, I am able to download the documentation: https://github.com/B-Lang-org/bsc/releases/latest/download/bsc_user_guide.pdf

quark17:
The github documentation does say to use `releases/latest/download/`.  I do see why you got it backwards, though: for named releases, the link is `releases/download/2023.01/`.  If you reverse it, `releases/2023.01/download/`, that doesn't work.  So you can't just replace a name with `latest` and have it work.  I'm not sure why the URLs aren't consistent.

kenta2:
OK thanks for catching that.  I'm going to abandon this PR and create a new one that doesn't have a bunch of commits going back and forth on this edit.

PR #541, opened 2023-02-02
Complex: Add support for complex conjugate function.

edowson:
This pull request includes a set of commits that:
- add support for a complex conjugate function,
- minor fixes to the test-suite readme for executing the test-suite.

### DIFF
--- a/doc/libraries_ref_guide/LibDoc/Complex.tex
+++ b/doc/libraries_ref_guide/LibDoc/Complex.tex
@@ -190,6 +190,21 @@ function Complex#(a_type) cmplxSwap( Complex#(a_type) cin ) ;
 \end{tabular}
 \end{center}
 
+\index{cmplxConj@\te{cmplxConj} (complex function)}
+\begin{center}
+	\begin{tabular}{|p{.7 in}|p{4.8 in}|}
+		\hline
+		&\\
+		\te{cmplxConj}&Negates the imaginary part.\\
+		&\\
+		\cline{2-2}
+		&\begin{libverbatim}
+			function Complex#(a_type) cmplxConj( Complex#(a_type) cin ) ;
+		\end{libverbatim}
+		\\ \hline
+	\end{tabular}
+\end{center}
+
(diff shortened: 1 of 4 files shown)


quark17:
Hi Elvis, thank you for the contribution!

I'm happy to merge the second commit (the changes to Complex).  Though we will also need to update the documentation in `./doc/libraries_ref_guide/LibDoc/Complex.tex`.

The first commit (the change to the testsuite README) needs some discussion.  The command given in the README is the correct command when run from the top directory of the test suite (the `./testsuite/` directory).  The command you changed it to is the correct command when run from the top level of the BSC repo (the `./` directory).  For historical reasons, the test suite is documented as its own unit.  (You'll note that the README is in the `testsuite` directory and, in the same paragraph where it mentions `make check`, it refers to the default installation directory as `../inst/` and not `./inst/`.)  But this has obviously created confusion.  I don't think that just changing the target (from `check` to `check-suite`) is going to fix this confusion.  I think the README needs to clarify the directory where the top-level commands are run.  Plus, there are other `make` commands given in the README (`smoke`, `fullcheck`, `checkparallel`, `fullparallel`) that are only available when run from the `testsuite` directory, as they are not targets in top level BSC makefile.

edowson:
Hi Julie,

I have made the following changes:
- added documentation for cmplxConj and tested the changes
- removed the commit for modifying the testsuite documentation

quark17:
Thanks again!

By the way, it looks like you submitted a PR from the `main` branch of your fork repo.  Because of that, you will probably need to delete your repo and create a new one, the next time you want to submit a PR.  This is a lesson that I learned the first time I created a PR.  The recommended way of operating is to not push any commits to `main` and instead leave that branch to track the upstream repo; when you want to develop or submit a PR, create a _new_ branch, push your commits to that branch, and submit the branch as a PR.  Then you can delete the branch when the PR is merged.  And you can continue to pull upstream changes into `main`, without any conflicts.  That also allows you to have multiple PRs open or to continue development on something else while the PR is pending, because each will be on their own branch.

PR #549, opened 2023-02-24
util/vim: Fix indentation for interfaces and if guards

romangraef:
This commit fixes two issues with the vim indentor.

 * Indentation of methods in interfaces:

```bsv
interface SomeInterface;
    method Action someMethod();
    // Previously this would indent the next method once more
    method Action someOtherMethod();
endinterface : SomeInterface
```

 * If guards on methods being treated as if statements, therefore only indenting the next statement

```bsv
method Action someMethod() if (ready);
    a <= 0;
    // Previously this next line would not be indented at all.
    b <= 0;
endmethod
```

### DIFF
--- a/util/vim/indent/bsv.vim
+++ b/util/vim/indent/bsv.vim
@@ -82,16 +82,6 @@ function GetBSVIndent()
       echo vverb_str "De-indent after a multiple-line comment."
     endif
 
-  " Indent after if/else/for/case/always/initial/specify/fork blocks
-  elseif last_line =~ '`\@<!\<\(if\|else\)\>' ||
-    \ last_line =~ '^\s*\<\(for\|case\%[[zx]]\|do\|foreach\|randcase\)\>' ||
-    \ last_line =~ '^\s*\<\(always\|always_comb\|always_ff\|always_latch\)\>' ||
-    \ last_line =~ '^\s*\<\(initial\|specify\|fork\|final\)\>'
-    if last_line !~ '\(;\|\<end\>\)\s*' . vlog_comment . '*$' ||
-      \ last_line =~ '\(//\|/\*\).*\(;\|\<end\>\)\s*' . vlog_comment . '*$'
-      let ind = ind + offset
-      if vverb | echo vverb_str "Indent after a block statement." | endif
-    endif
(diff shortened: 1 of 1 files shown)


exhq:
yo this pr bussin frfr u should 100% merge bro fr on god 💯

quark17:
This is awesome, thank you for contributing!  I'm not familiar with vim code, though, so I don't know how to judge if this is worth merging.  It would be good if we could develop some tests, that would assure that the changes are not breaking something else in the process.  Do you have any thoughts on how we could test it?  Is there a command-line that can be run to cause vim to autoformat some input text, and then we could compare the output against a golden reference?

romangraef:
The formatter should automatically run when typing new code in vim, but if you want to run it on an existing file you can open the file in vim and type `gg=G` (`gg` goes to the beginning of the file, `=` equals autoindent and `G` means till the end of the file). To automatically invoke vim to format an entire file you can use `vim input.bsv +"normal\! gg=G" "+wq output.bsv"` (if `output.bsv` is not specified then the change is done in place by overwriting the original file). Also keep in mind that this is an auto indenter more than an auto formatter, so this would probably fail a bunch of tests that a normal auto formatter would properly run on (both before and after this change). I am unsure about how to implement an automatic test for this with your current test infrastructure.

quark17:
Sorry for the long delay in responding! And thank you again for the contribution!  I'm OK accepting the PR as it is.  We'll just manually try it out, and then merge it.

Also, I notice that @mtikekar has a Vim plugin for BSV, that seems to have more features? https://github.com/mtikekar/vim-bsv

quark17:
This is definitely an improvement, so I'm happy to merge it.

An example where the indentation still fails (if you or anyone else wants to look at it) is `src/Libraries/Base3-Misc/CRC.bsv`.  The `endmodule` on the first module fails to unindent, so everything else in the file is indented by one (and the contents of those later modules is not further indented), and the `endpackage` is still indented by one.

Also, I have no attachment to the vim-mode in this repo over the one maintained by @mtikekar.  I'm happy for that to be the official one, or for them to be merged, or whatever people want to do.

mtikekar:
I have not used my vim plugin in a few years but happy to have it be merged here and help however needed.

PR #554, opened 2023-05-23
Replace egrep with grep -E

Vekhir:
`egrep` is deprecated and emits lots of warnings during compilation, and that is quite distracting from the warnings which gcc emits. With that in mind, I set out to create this PR.

This PR seeks to replace `egrep` with the recommended alternative `grep -E`\
No changes to final behaviour.

With this PR applied, I didn't see any warnings about egrep anymore, allowing me once again to serenely watch as the gcc warnings go by.


quark17:
Oh, yes, this should be changed.  I didn't realize that `egrep` was still being used in places.  I will merge this as soon as the testsuite finishes.  Thank you!

PR #555, opened 2023-05-24
Remove CI jobs for macos-10.15 and ubuntu-18.04

quark17:
GitHub no longer provides runners for `macos-10.15` and `ubuntu-18.04`.  When CI is run, these jobs will hang waiting for a server to take the job, which will never happen.  The jobs might time out, but I haven't waited long enough.  Better to remove them.  This also allows us to remove some special code for installing systemC libraries on `ubuntu-18.04`.

The list of environments provided by GitHub is here: https://github.com/actions/runner-images#available-images

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
   build-ubuntu:
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


rsnikhil:
Sounds reasonable to me to remove those OS versions from CI.

PR #560, opened 2023-06-02
Fix tests for iverilog v11 (SFT deprecated warning)

### LINKED ISSUE #344: Icarus verilog v11
As b-lang-org/bsc-testsuite#16 (and related PR b-lang-org/bsc-testsuite#17) show, the way that BSC invokes iverilog for designs with imported foreign functions leads to this warning message:
```
SFT files are deprecated. Please pass the VPI module instead.
```
I suspect that the `bsc_build_vsim_iverilog` script in BSC will need to be updated to check the iverilog version and to use different VPI arguments when the version is 11.   This should be simple to fix, but it requires figuring out what the new invocation should look like -- so I've marked this as a "good first issue".

In the testsuite, in `bsc.options`, the exp-file was updated to filter out this deprecation warning (which was otherwise causing a failure when comparing the expected output).  This filtering can be removed when the exec script is fixed.

There are also a number of places in the testsuite where we expect verilog tests to fail, for certain versions of iverilog, due to bugs in the VPI support. It would be good to check to whether any of these bugs have been fixed and can be enabled when the iverilog version is 11.

Vekhir:
In IVerilog v11+, the use of sft files is deprecated and creates a warning. This warning leads to failing test cases, though behaviour is unchanged.

This PR is based on https://steveicarus.github.io/iverilog/usage/vpi.html and replaces the sft file creation with the -L option to pass the vpi files. The design works for IVerilog v11 and v12.
In addition, a version check is introduced that applies the old behaviour if the version is 10.x. I wasn't able to test with IVerilog v10.x as my distro doesn't provide this version for some time. So that's left for the CI to check.

Fixes #344

### DIFF
--- a/testsuite/bsc.options/options.exp
+++ b/testsuite/bsc.options/options.exp
@@ -289,7 +289,7 @@ if { $vtest == 1 } {
     # iverilog 10.1 has spurious warnings
     if { $verilog_compiler == "iverilog" &&
          [regexp {^\d+\.\d+} $verilog_compiler_version majmin] &&
-         10.0 < $majmin && $majmin < 10.2
+         $majmin == "10.1"
     } {
         append ere { -e {/inherits dimensions from var/d}}
     }
--- a/testsuite/bsc.options/options.exp
+++ b/testsuite/bsc.options/options.exp
@@ -287,12 +287,11 @@ if { $vtest == 1 } {
     set filtfile "$rawfile.filtered"
     set ere {-e /WARNING:\ IVerilog/d -e /not\ guaranteed/d}
     # iverilog 10.1 has spurious warnings
-    if { $verilog_compiler == "iverilog" && $verilog_compiler_version == "10.1" } {
+    if { $verilog_compiler == "iverilog" &&
+         [regexp {^\d+\.\d+} $verilog_compiler_version majmin] &&
+         10.0 < $majmin && $majmin < 10.2
+    } {
         append ere { -e {/inherits dimensions from var/d}}
-    }
-    # Silence deprication warning in iverilog 11.0 for now
-    if { $verilog_compiler == "iverilog" && $verilog_compiler_version == "11.0" } {
-        append ere { -e {/SFT files are deprecated/d}}
     }
     sed $rawfile $filtfile $ere {}
     compare_file $filtfile empty.expected
--- a/src/exec/bsc_build_vsim_iverilog
+++ b/src/exec/bsc_build_vsim_iverilog
@@ -129,7 +129,7 @@ fi
 
 # create SFT for iverilog older than v11 (only v10 supported)
 # v11 and later use the vpi files directly
-if [[ "$(iverilog -V 2>&1 | head -n 1)" =~ "v10_" ]]; then
+if echo "$(iverilog -V 2>&1 | head -n 1)" | grep -q "v10_"; then
     CREATESFT="yes"
 else
     CREATESFT="no"
--- a/src/exec/bsc_build_vsim_iverilog
+++ b/src/exec/bsc_build_vsim_iverilog
@@ -129,7 +129,7 @@ fi
 
 # create SFT for iverilog older than v11 (only v10 supported)
 # v11 and later use the vpi files directly
-if echo "$(iverilog -V 2>&1 | head -n 1)" | grep -q "v10_"; then
+if echo "$(iverilog -V 2>&1 | head -n 1)" | grep -q " 10."; then
     CREATESFT="yes"
 else
     CREATESFT="no"
--- a/src/exec/bsc_build_vsim_iverilog
+++ b/src/exec/bsc_build_vsim_iverilog
@@ -129,7 +129,7 @@ fi
 
 # create SFT for iverilog older than v11 (only v10 supported)
 # v11 and later use the vpi files directly
-if echo "$(iverilog -V 2>&1 | head -n 1)" | grep -q " 10."; then
+if echo "$(iverilog -V 2>/dev/null | head -n 1)" | grep -q "Icarus Verilog version 10."; then
     CREATESFT="yes"
 else
     CREATESFT="no"
--- a/src/exec/bsc_build_vsim_iverilog
+++ b/src/exec/bsc_build_vsim_iverilog
@@ -127,7 +127,13 @@ if [ -n "$BSC_OTHER_ARGUMENTS" ]; then
   exit 1
 fi
 
-CREATESFT="yes"
+# create SFT for iverilog older than v11 (only v10 supported)
+# v11 and later use the vpi files directly
+if [[ "$(iverilog -V 2>&1 | head -n 1)" =~ "v10_" ]]; then
+    CREATESFT="yes"
+else
+    CREATESFT="no"
+fi
 
 if [ "$CREATESFT" = "yes" ]; then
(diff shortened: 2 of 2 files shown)


quark17:
On lines 158 and 170, there's a non-breaking space (NBSP) character (ASCII 0xA0) instead of an ordinary space (0x20) between the `[` and `"` in the start of the if-statement:
```
if [ "$CREATESFT" = "yes" ]; then
```
In my emacs, it shows up as a red underscore.  Displaying it with `more` shows it as a space.  I can delete it in emacs (and replace it with a space), and that fixes the problem.

When the character is there, it causes the following shell errors:
```
+/path/bsc/inst/lib/exec/bsc_build_vsim_iverilog: 158: /path/bsc/inst/lib/exec/bsc_build_vsim_iverilog: [ yes: not found
+/path/bsc/inst/lib/exec/bsc_build_vsim_iverilog: 170: /path/bsc/inst/lib/exec/bsc_build_vsim_iverilog: [ yes: not found

```

Vekhir:
Interesting. Thanks for the comment, that would have been hard to find. Especially since my shell seems unaffected by it.

The latest commit should fix that. After that I'll reintroduce the version check and add the -L option for newer IVerilog. Hopefully without rogue whitespaces :smile:

Vekhir:
This commit series should now fix the issue and give IVerilog the VPI files instead of SFT while preserving support for 10.x

@quark17 Since the tests are passing on main, are there test changes/additions that should be made in order to show that the issue is solved? I'm thinking of suppressed warnings that can be re-enabled or something.

Vekhir:
Getting an Ubuntu 20.04 VM running was easier than I thought.

Apparently, the Ubuntu release does not contain the v10_x tag, so we can't rely on it. Replaced it with the Version 10.x

quark17:
In the testsuite (in `get_verilog_compiler_version` in `testsuite/config/verilog.tcl`), we get the iverilog version by getting the first line of output:
```
iverilog -V 2>/dev/null | head -1
```
and then extracting the first group of non-space characters following `Icarus Verilog version`:
```
regexp -- {^Icarus Verilog version ([^ ]+).*$} $line1 tmp verilog_compiler_version
```
This could be of form `x.y.z` and might even have some tag information at the end, I think (as alphanumeric and underscores)?  The testsuite is just interested in the major and minor version, when concerned about version, so in those places we get the "maj.min" string with tcl regexp, which we can then compare numerically.  For example in `testsuite/bsc.verilog/tasks/tasks.exp`:
```
if { $verilog_compiler == "iverilog" &&
     [regexp {^\d+\.\d+} $verilog_compiler_version majmin] &&
     $majmin < 11
 } {
```
Although I do see one place in `tasks.exp`, where we are matching the whole version string, which is probably bad (and should be changed to using the extracted `majmin` and numerical comparison operators):
```
if { $verilog_compiler == "iverilog" &&
     ($verilog_compiler_version == "10.1" || $verilog_compiler_version == "10.2") } {
```
(Also, I'm a bit concerned that the Tcl is using `-1` instead of `-n 1`, but it seems to work?  And I don't know if your piping of stderr to stdout is the best way to handle errors?)

I wonder if your grep should be checking for the whole pattern, `Icarus Verilog version 10.`?

> Since the tests are passing on main, are there test changes/additions that should be made in order to show that the issue is solved? I'm thinking of suppressed warnings that can be re-enabled or something.

If we want to add a test, it could probably be done in `testsuite/bsc.codegen/foreign/`.  In `foreign.exp`, there are examples containing foreign functions that are being compiled and run.  After one of those, you could add a check that `SFT` (or `sft`) doesn't exist in the output of the BSC linking stage.  You'd want to guard it, so that the check only occurs for iverilog and only for versions 11+.  So maybe something like this:
```
if { $verilog_compiler == "iverilog" &&
     [regexp {^\d+\.\d+} $verilog_compiler_version majmin] &&
     $majmin >= 11
 } {
```
There are test procedures like `find_n_strings` and `find_regexp` etc defined in `testsuite/config/unix.exp` that can be used.  Maybe `find_regexp_fail`, which expects to not find the pattern.  However, if iverilog's error message changes and doesn't contain `SFT`, then this test would continue to pass, even if there was a problem.  So a stronger possibility would be to compare the entire output against an expected file (with `compare_file`), but iverilog also outputs a warning about VPI support being incomplete, so you'd need to include that for now.  Just checking for SFT is probably fine.

Vekhir:
The great restriction for version checking is that `bsc_build_vsim_iverilog` is run using the POSIX shell `sh`, so e.g. regexp is not available. We only need to check for existence anyway, so grep should be fine.\
I have nonetheless changed the error pipe to /dev/null, and the entire prefix up to "10." is checked as it is unlikely to change for this version.

As for tests, I found that the warning is suppressed in `bsc.options/options.exp` for IVerilog v11, which explains why _I_ get the error, but the CI runs fine (I have v12). I think that removing that check is good enough for a test.

While at it, I also changed the 10.1 check to your suggested syntax using numerical comparison instead of string comparison. There are also workarounds for VPI issues in `bsc.codegen/foreign/foreign.exp` and `bsc.codegen/foreign/battery/battery.exp` which I can confirm are still needed for v12.

Vekhir:
GitHub seems to take its time updating this page. The commits are available on [my fork](https://github.com/Vekhir/bsc/tree/fix-tests-iverilog-v11).

Vekhir:
@quark17 I'm happy with these commits. Is there anything you'd like to see before merging?

testsuite/bsc.options/options.exp
```diff
@@ -287,13 +287,12 @@ if { $vtest == 1 } {
     set filtfile "$rawfile.filtered"
     set ere {-e /WARNING:\ IVerilog/d -e /not\ guaranteed/d}
     # iverilog 10.1 has spurious warnings
-    if { $verilog_compiler == "iverilog" && $verilog_compiler_version == "10.1" } {
+    if { $verilog_compiler == "iverilog" &&
+         [regexp {^\d+\.\d+} $verilog_compiler_version majmin] &&
+         10.0 < $majmin && $majmin < 10.2
```
quark17, line 292: I don't think this use of comparison operators works, if the minor version increases to 10 and above -- since 10.10 would be between 10.1 and 10.2 numerically.  Checking for exactly "10.1" wasn't OK on the full version string, but I think it should be OK to do once we have extracted just the majmin part of the version string.

Vekhir: How likely do you think will a version 10.10 exist? But okay, we can do direct comparison.

Do you mean something like:
```
-    if { $verilog_compiler == "iverilog" && $verilog_compiler_version == "10.1" } {
+    if { $verilog_compiler == "iverilog" &&
+         [regexp {^\d+\.\d+} $verilog_compiler_version majmin] &&
+         $majmin == "10.1"
```
Only the last line changed to compare the extracted majmin version

Vekhir:
@quark17 I've implemented your suggestion.

I can only check that nothing breaks, as can the CI since Ubuntu 18.04 was dropped.

quark17:
Thank you for your patience!  Looks good, I'll merge.  Also, sorry that I forgot about your other issue asking about v12 failures and the testsuite.  I'll look at that and respond now.  Thank you again!

Vekhir:
Your guidance throughout this PR is also much appreciated. I'm glad I could be helpful!

PR #563, opened 2023-06-10
Replace vendored STP with system package

Vekhir:
As much as the goal of #554 was to see the gcc errors again, I have decided that I don't actually like them. Since most of them come from compiling STP, this PR seeks to remove the vendored version within BSC and instead require it as a pre-installed dependency.

It was surprisingly easy to achieve this goal of removing STP. The first two commits deal with removing all references outside of vendor/stp to STP, and replace them with the locations of system installed packages (here /usr/lib/ and /usr/include/). The foreign function interface definition for Haskell is kept.

The third commit removes all code inside vendor/stp/ except for the aforementioned Haskell FFI. Most of the ~560 files changed are indeed deletions for this reason.

In all honesty, I didn't expect all that to even work first try. But whatever test I throw at it, it refuses to fail during compilation or testing. It does however fail if I remove the system package, so it clearly uses that.

The CI will need to be adapted to this change too. Incidentally, Homebrew provides an STP package, Ubuntu however does not. Afaict, only Arch Linux, Fedora, FreeBSD, and OpenSuse actually provide a system package for STP by default.\
On the other hand, BSC already needs to be compiled by almost everyone, so maybe we can get away with saying: `Compile STP yourself`?

Tasks:
- [ ] Get MacOS CI working
- [ ] Get Ubuntu CI working


Vekhir:
A few issues I encountered is that STP is required at runtime for some reason. Without it, BSC won't even start. I'd guess that something in the stub STP file avoids this issue.

I tried some ideas around making STP truly optional during build and runtime, but I don't know enough Haskell to succeed. Still, this PR shows that the current C API isn't too different from the vendored one, so in theory, STP could be replaced by a system package without changing every bit of the implementation.

As mentioned in other places (#31, #404), implementing a standard interface via smt-lib et. al. is probably the best solution to then remove the vendored STP. There might be some value in keeping the vendored Yices, in order to have an easier default installation for distros where Yices is not available as a package.

Anyway, I'm closing this as I won't be working on this anymore. I was just surprised how well my simple idea to just point to the system package worked.

quark17:
Also note that you can make BSC with `STP_STUB=1` (or any non-empty string) and the repo will just build stub code, not the actual STP library.  BSC uses Yices by default, so as long as you don't provide the `-sat-stp` flag, it'll be fine.  If you did give the flag to choose STP, BSC will notice that it has loaded a stub and report and error.

PR #566, opened 2023-06-23
bs-classic: Swap the precedence level of `:=` and `$`

rossc719g:
In bluespec classic, this allows `$` to be used on either side of a `:=`.
It should not have any impact on bluespec verilog, where `$` is not a valid operator (AFAIK).

E.g., something like this would be syntactically legal:
`select foo $ f1 a := map f2 $ map f3 v`

Wanting to use `$` on the lhs of an assignment is probably uncommon, but the rhs is more sensible.
E.g., something like:
`foo := map f $ map g v`
is currently an error, because, it would try to pass `map g v` as an argument to `foo := map f`, which is not the intention.
I think the result of a `:=` is always `Action` which can't take an argument, so I do not believe the existing behavior needs to be preserved.

Alternatives that might be worth considering:
Setting both operators to the same precedence level works to allow `$` on the rhs of `:=`.
It feels more natural to me to swap them though.


quark17:
Thank you for contributing this PR!

PR #576, opened 2023-07-15
Generate VPI files in the same dir with Verilog

quark17:
This addresses an issue reported in Discussion #575.

Generated Verilog files are placed in the `-vdir` directory or, if no `-vdir` is specified, in the same directory as the source file being compiles.  VPI wrapper files are also placed in the `-vdir`, but if no `-vdir` is specified, they were being placed in the current directory (where BSC is invoked).  These should be consistent, so the location for VPI wrappers is fixed to match the generated Verilog.

Simple testing is added to `bsc.options`.

### DIFF
--- a/src/comp/Error.hs
+++ b/src/comp/Error.hs
@@ -1118,6 +1118,7 @@ data ErrMsg =
         | EMissingUserFile String [String]
         | EUnrecognizedCmdLineText String
         | EUnknownVerilogSim String [String] Bool
+        | EMissingVPIWrapperFile String Bool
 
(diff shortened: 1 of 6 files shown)


quark17:
I have pushed a commit which also changes the BSC link step, to search for VPI wrapper files in the Verilog search path.  This allows BSC to link designs that pull in previously-compiled code from various places, which might each have their own VPI wrappers -- where previously BSC expected the wrappers to be in one directory (either the link step's vdir or the current directory where the link step is invoked).

Issue #584, opened 2023-08-01
SV preprocessor parentheses parsing


quark17:
As mentioned in issue #469, there are warnings about non-exhaustive pattern matching in `SystemVerilogScanner.lhs`, in the function [`scanLinePosDirective`](https://github.com/B-Lang-org/bsc/blob/0e9a5b2624d91d514a2fb586b7e5d8c94050853b/src/comp/SystemVerilogScanner.lhs#L309-L322) which parses `` `line `` directives.

The preprocessor accepts two kinds of `` `line `` directive: one according to the Verilog/SystemVerilog standards (without parentheses) and one with parentheses (which may be something that Bluespec invented, for preserving position information in the pre-processor output?).  The function `scanLinePosDirective` is for parsing the directives with parentheses.  A comment on the function says "One day the gods will smite me for this" -- I don't know if this is for inventing the new syntax or because the function doesn't handle any error conditions!

I wrote a few examples of error situations: closing paren but not enough arguments; EOF before the closing paren; missing closing paren; closing paren (or arguments) on the following line; non-numeric values where numeric arguments expected; etc.  Surprisingly, I found that a directive spread over multiple lines will still parse -- I don't know if that's expected.  (This is not high priority, but I wanted to record the issue.)

Macro definitions (the `` `define `` directive) can also take arguments in parentheses.  This is handled in [`SystemVerilogPreprocess.lhs`](https://github.com/B-Lang-org/bsc/blob/0e9a5b2624d91d514a2fb586b7e5d8c94050853b/src/comp/SystemVerilogPreprocess.lhs#L219) (which also generates non-exhaustive pattern warnings).  I wrote some examples and found that error situations are also not being handled there.  (This is more serious, because it's a user-visible feature and an actual part of the SV spec.)  One interesting quirk is that, if the close paren is missing, the directive will continue to be parsed on later lines until a close paren is found (and extra text is silently ignored).

I will commit the examples into `testsuite/bsc.preprocessor/misc/`.

quark17:
[...]

Issue #469, opened 2022-04-18
Investigate and resolve GHC "incomplete" warnings


quark17:
The GHC warnings `-Wincomplete-uni-patterns` and `-Wincomplete-record-updates` may have been around for a while, but were off by default and not turned on with `-Wall`.  In GHC 9.2, they are now included in `-Wall` (see [here](https://gitlab.haskell.org/ghc/ghc/-/issues/15656)), and when we build BSC with GHC 9.2.2 we now see a ton of warnings.  We should investigate these warnings, to see if any indicate an actual bug, and in either case fix the code so that there is no warning.

I started to do this myself, but there are so many files to look at, that it would help if others can also look.  Maybe comment on this issue to claim a file, to avoid duplicating effort?

The effort here is in inspecting each warning and trying to work out if the missing patterns are reachable.  If they are, then BSC will abort with a GHC error.  At the very least, we want to handle these patterns explicitly, as either a BSC user error (if it's legitimately reachable) or a BSC internal error (if we don't expect it to occur).  For example, `ASchedule.hs` has this code:
```
getRule ids = l
  where (l:_) = ids
```
which is only legitimate if `ids` is not an empty list.  We can certainly eliminate the warning by adding an arm for `[]` and reporting an internal error in that case -- but we should work out whether the list can ever be empty, otherwise we're just ignoring a potential issue.  If we identify a bug, we don't need to fix it now; we can open a GitHub Issue ticket for it.

FYI, the `Util.hs` module has several `*OrErr` versions of common Haskell functions, which give a BSC internal error instead of a GHC error.  So the above code in `ASchedule.hs` would be best re-written as the following (if we decide that the empty list should not occur):
```
getRule ids = headOrErr "addMEAssumps getRule" ids
```
This also points out that BSC may currently be calling library functions like `head` in places, which is equally as dangerous as a missing case-arm for empty list, but is _not_ being warned about by GHC.  If you notice uses like this, please also flag them (or resolve them).

In some cases, the BSC source code has a proper guard on the incomplete case, but GHC doesn't see this, and flags the issue.  For example, in `ASchedule.hs` there are list comprehensions that include this code:
```
isJust m_methods2,
let (Just methods2) = m_methods2,
```
GHC complains that `Nothing` is not handled, but we've checked for that.  We can appease GHC (and make the code cleaner in the process) by replacing it with this:
```
(Just methods2) <- [m_methods2],
```
[...]

PR #585, opened 2023-08-02
Resolve ghc 9.2 warnings

quark17:
This resolves most of the incomplete-pattern warnings that occurred when building with GHC 9.2.8 (Issue #469).  This also adds examples for SV-preprocessor issues that were revealed by the warnings (Issue #584).  That seems to be the only area where the warnings were an indication of an issue.


quark17:
I have added two more commits.  One gets BSC to compile with GHC 9.6, by addressing changes in the `Control.Monad` libraries.  The other removes an old abstraction (`ErrorTCompat`) that hasn't been needed since we dropped support for GHC earlier than 8.

Issue #581, opened 2023-07-29
CI should test BSC with newer GHC versions


quark17:
PR #580 will upgrade the CI to use GHC 9.2.8, which is a stable recommended version, and what we will use for pre-built releases.  There are, however, newer versions of GHC, specifically 9.4.5 and 9.6.2.  We should add jobs to the CI to test BSC with these newer versions as well, so that we can monitor compatibility (and not discover it later when we want to upgrade).

quark17:
I have a prototype on the [`reorg-ci` branch](https://github.com/quark17/bsc/tree/ci-reorg) of my fork, which abstracts the build and test jobs into a call-able workflow, that the top-level CI can call with specific OS and GHC version inputs.  GitHub's display of nested jobs is a little confusing (and different for matrix and non-matrix callers), which you can see in the Actions tab, for example [this run](https://github.com/quark17/bsc/actions/runs/7362252048).

quark17:
FYI, callable (or reusable) workflows is documented here: https://docs.github.com/en/actions/using-workflows/reusing-workflows

And I notice that you can specify a subset of a matrix to run, by not specifying any matrix variables and using `include` to specify only the configurations to run (see [example in the GitHub documentation on matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations)).  Currently, I've defined two separate top-level jobs (one for the OS matrix and one for the latest GHC), but we could just combine those into one job that explicitly lists all the combinations to run.

PR #597, opened 2023-08-11
Update GHC support in INSTALL.md

Vekhir:
The supported GHC versions have changed and need to be reflected in INSTALL.md

The GHC version in CI has been upgraded to 9.2.8 from 9.0.2 (#580),
Support for < 8.0 has been dropped (#585),
Support for 9.4 and 9.6 is in the works (#581), and
GHC 9.6 is known to compile (#585).

These 4 facts are now reflected in the INSTALL.md with outdated information removed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -107,13 +107,13 @@ For example (cabal v3.x only):
     $ cabal v2-install --package-env=default syb old-time split
     $ make GHC="ghc -package-env default"
 
-Bluespec compiler builds are tested with GHC 9.0.1.
-GHC releases older than 7.10.1 are not supported.
+Bluespec compiler builds are tested with GHC 9.2.8.
+GHC releases older than 8.0.1 are not supported.
 
 The source code has been written with extensive preprocessor macros to
-support every minor release of GHC since 7.10, through 9.0. The source
-has not yet been updated for 9.2 or beyond.  Any releases in that
-range should be fine.  The stable releases of 8.10.7 and 9.0.1 (at the
+support every minor release of GHC since 8.0, through 9.2. The source
+has been updated for 9.6 to compile. Any releases in that
+range should be fine. The stable releases of 9.0.2 and 9.2.8 (at the
 time of writing) are known to work.
 
 ## Additional requirements
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -108,13 +108,12 @@ For example (cabal v3.x only):
     $ make GHC="ghc -package-env default"
 
 Bluespec compiler builds are tested with GHC 9.2.8.
-GHC releases older than 8.0.1 are not supported.
+GHC releases older than 7.10.3 are not supported.
 
 The source code has been written with extensive preprocessor macros to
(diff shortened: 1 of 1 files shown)


quark17:
Thanks for contributing this!

FYI, BSC still builds with GHC 7.10.3 (although the I see that I needed to add `mtl` to the `cabal` command given in the build instructions).  I notice that GHCUP still lists 7.10.3 as an option.  It's probably OK to say that GHC 8+ is needed, but it also doesn't hurt to leave in 7.10.  Was there a reason you thought that 7.10 should be dropped?

quark17:
Ah, I see I was sloppy and mentioned 8.0 in the PR that you reference.  I should have said 7.10.

Vekhir:
There is a slight difference between "supported" and "happens to work" which I used to explain the discrepancy between your comment and commit message, the latter stating correctly that 7.10 is unaffected. The situation was still somewhat confusing, so I chose to just say 8.0 for the initial PR.

If 7.10 still qualifies as "supported", then I have no issue with keeping that particular language.

quark17:
We don't have a policy on what we want to support.  Technically, the only thing we can say we support is what we test, which is just 9.2.8 at the moment.  Although I have been experimenting [in my own fork](https://github.com/quark17/bsc/commits/ci-ghc-matrix) with adding testing of more GHC versions to the CI.  I was just looking at adding 9.6.2.  We could add more, but I worry about the CI time getting out of control then?

Unless someone wants to propose a policy ... I'm happy to say that we still support 7.10.3+ until it becomes a hassle to do so ... and I'm happy to just track the versions that GHCUP lists. (There's no need to keep support for more versions than GHCUP, but we could choose to support fewer versions, if someone wants to propose a policy.)  I also notice that the GHC download page lists only 9.2.8, 9.4.6, and 9.6.2 as "stable releases" (and the rest as "older").  I'm happy to say that BSC has two tiers of support: the three GHC "stable release" are more supported, meaning that any issues would be a priority to fix; whereas, for older versions, we just support that the code will compile.

The Makefile and source code still has ifdefs to support 7.10; we haven't removed those.  We probably should update `src/comp/Makefile` so that it doesn't warn about 9.4 and 9.6 not being supported, since I would consider those supported now (9.6.2 compiles and passes the CI tests on Ubuntu).

I might propose the following text (for now until we add more GHC versions to the CI):
```
Bluespec compiler builds are tested with GHC 9.2.8.
GHC releases older than 7.10.3 are not supported.

The source code has been written with extensive preprocessor macros to
support compiling with every minor release of GHC since 7.10, through 9.6.
Any releases in that range should be fine.
The stable releases of 9.0.2 and 9.2.8 are reasonable choices.
```
Although GHC's website doesn't call 9.0.2 a "stable release", so I'm open to a different wording.  I'm also happy to drop the mention of 9.0.2 or to say that 9.2.8 is the recommended version.

Vekhir:
I've updated the PR with another commit. The chosen wording just mentions 9.2 in its latest point release as recommended.

quark17:
Thank you!

Issue #595, opened 2023-08-11
STP_STUB option leads to build failure


Vekhir:
### Issue
When using the build option STP_STUB=1 during building in order to disable STP, BSC fails to build with error
```
/build/bluespec-git/src/bsc/inst/bin/bsc -stdlib-names -bdir /build/bluespec-git/src/bsc/build/bsvlib -p . -vsearch /build/bluespec-git/src/bsc/build/bsvlib -no-use-prelude Prelude.bs
/build/bluespec-git/src/bsc/inst/bin/core/bsc: error while loading shared libraries: libstp_stub.so: cannot open shared object file: No such file or directory
```
Apparently, `bsc` is looking for `libstp_stub.so`, eventhough it should search for `libstp.so.1`, which does exist in `bsc/inst/lib/SAT/`. 

### More information
Error occurs on the main branch (commit 42fb7b6cc030346a149db57ba65c79c5d6885824). Omitting the build option builds just fine. Also tried to build with (the randomly chosen) commit a7f3b2c29dee6ecc59bcaec44b6f7d9a80795c9f, but failed with the same error.

Command: `make STP_STUB=1 GHC="ghc -dynamic" GHCJOBS=4 GHCRTSFLAGS='+RTS -M5G -A128m -RTS' install-src`

### System information
OS: Arch Linux
Kernel: 6.4.9-arch1-1
GHC: 9.0.2-3
BSC: 42fb7b6cc030346a149db57ba65c79c5d6885824 and a7f3b2c29dee6ecc59bcaec44b6f7d9a80795c9f (via [bluespec-git](https://aur.archlinux.org/packages/bluespec-git))

<details>
<summary><b>Full error log</b></summary>

```
make  -C src  PREFIX=/build/bluespec-git/src/bsc/inst  install
make[1]: Entering directory '/build/bluespec-git/src/bsc/src'
make  -C vendor/stp   PREFIX=/build/bluespec-git/src/bsc/inst  install
make[2]: Entering directory '/build/bluespec-git/src/bsc/src/vendor/stp'
make -C src_stub install
make[3]: Entering directory '/build/bluespec-git/src/bsc/src/vendor/stp/src_stub'
cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -c -o stp_stub.o stp_stub.c
[...]

PR #600, opened 2023-08-14
Extend STP_STUB and YICES_STUB flag

Vekhir:
This commit series renames STP_STUB to STP_USE_LIB and extends it to accept config values instead of being a simple boolean flag. The old behavior of STP_STUB=1 is realized with STP_USE_LIB=disable

Two new options are added to the flag, `system` and `arbitrary absolute path`. The first option searches the standard locations to find the system installed version of STP, while the latter takes a path to a user-defined library providing STP.

YICES_STUB will be refactored in a similar manner to ensure consistency between the two flags.

### DIFF
--- a/src/vendor/stp/Makefile
+++ b/src/vendor/stp/Makefile
@@ -5,10 +5,24 @@ PREFIX?=$(TOP)/inst
 
 .PHONY: all install clean full_clean
 
-ifeq ($(STP_STUB),)
-SRC = src
-else
+ifneq ($(STP_STUB),) # If STP_STUB exists, disable STP. Otherwise check STP_USE_LIB
 SRC = src_stub
+else ifeq ($(STP_USE_LIB),)         # default (compile from provided source)
+SRC = src
+else ifeq ($(STP_USE_LIB),included) # compile from provided source
+SRC = src
+else ifeq ($(STP_USE_LIB),disable)  # disable STP at runtime
+SRC = src_stub
+else ifeq ($(STP_USE_LIB),system)   # use system STP
+SRC = src_sys
+STP_LIBPATH = /usr/lib/
+export STP_LIBPATH
+else ifneq ($(shell echo "$(STP_USE_LIB)" | grep -E "^/*"),) # use STP at the absolute path provided by STP_USE_LIB
+SRC = src_sys
+STP_LIBPATH = $(STP_USE_LIB)
+export STP_LIBPATH
+else # unknown option for STP_USE_LIB
+$(error Unknown option for STP_USE_LIB: $(STP_USE_LIB))
 endif
 
 ifeq ($(OSTYPE), Darwin)
--- a/src/vendor/stp/Makefile
+++ b/src/vendor/stp/Makefile
@@ -33,6 +33,9 @@ ifeq ($(STP_USE_LIB),included)      # compile from provided source
 else ifeq ($(STP_USE_LIB),disable)  # disable STP
     SRC = src_stub
     SNAME=libstp_disabled.$(SO)
+else ifeq ($(STP_USE_LIB),placeholder)  # use stub to be replaced by real library later
+    SRC = src_stub
+    SNAME=libstp.$(SO)
 else ifeq ($(STP_USE_LIB),system)   # use system STP
     SRC = src_sys
     ifeq ($(OSTYPE), Darwin)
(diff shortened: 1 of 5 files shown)


Vekhir:
Previous attempts to change the library loading behavior of STP and YICES are listed in [this comment](https://github.com/B-Lang-org/bsc/pull/404#issuecomment-912231615). Ultimately, something like [SBV](https://leventerkok.github.io/sbv/) is the best solution, but there is no ETA and I'm not too familar with Haskell, so waiting for that is hardly an option. 

This PR extends the CLI options at build time to enable usage of the system library for STP, or even a custom library. The changes are limited to the respective Makefiles and INSTALL.md (the latter for obvious reasons). This is done to hopefully quickly merge the changes after the next release.

Vekhir:
Taken from [this comment](https://github.com/B-Lang-org/bsc/issues/595#issuecomment-1676520908) by @quark17:

> [...] I wonder if a build with the system library should just not install any object file and use what's in the system. (I guess that depends on whether such a build wants to require the users to also have the library installed.) 

When specifying `STP_USE_LIB=system`, then the Makefile will search for the system installed library. Note that the library is only copied to `inst` if it is self-compiled (i.e. the included source or the stub), otherwise the user has to provide their library at runtime aswell.

> Also, my sense has been that an installation can change -- like, if it started with a yices stub, but then someone wanted to add yices, they could change what's in the lib/SAT/ directory (either put a real library in that directory or remove the stub and let the one from the system be found) -- and if that's the perspective we want to support, then the bsc binary should always be looking for the generic name (which can be a link to different implementations) and not looking for an implementation-specific name depending on how the installation was built (like libstp_stub.so or libstp_system.so etc).

Given that this is about a build time option, I don't think that changing the library later is a use case that has to be supported. Indeed, changing the name, or even creating a symlink, isn't terribly difficult.\
And in my experience, the system/custom option ultimately take the name of the library from the library itself, no matter how the actual file is called. That is one reason why custom libraries aren't copied to the installation in `inst`, but need to be provided by the user at runtime too.

#### Other proposals

> Also, there have been a lot of proposals of how to replace the current STP/Yices integration with other ways of doing things (see open issues/discussions mentioning STP); if we're putting effort into an overhaul, would it be better to put the effort into one of those bigger overhauls? (At the very least, we might consider using a newer STP. The snapshot we have doesn't have a version function, so we can't even warn the user that the library that's found doesn't match, and I see that the STP in github has a function for getting the version tag.)

Yeah, I know about these discussions, and also know about the great summary I mentioned in the previous comment. But as stated elsewhere, all of that takes even more time and does things that aren't necessary at this point. So the question is `incremental improvement now` or `complete solution sometime in the future`? Basically, what I'm saying is that this PR and all of the other proposals don't take anything away from each other.

> if we're putting effort into an overhaul, would it be better to put the effort into one of those bigger overhauls?

To reply to this point more specifically: This question relies on the assumption that the effort needed for the big proposals is of the same kind (just more of it) as the one needed for this PR. But they are completely different on a technical level, sufficiently different that I don't consider myself able of implementing them. In part, this misunderstanding may have been caused by my usage of "overhaul", as this PR is really just a minor refactor + feature addition. Doesn't alleviate the need for testing, but certainly not the grand vision that the other proposals are.

quark17:
All good.  And thank you for your effort and input on this!

For some background: Yices and STP were added back when BSC was a proprietary tool.  Yices was not open source either at the time, but people could download the pre-compiled library for individual use.  The Bluespec company didn't have the right to distribute the library.  Instead, BSC was shipped with a Yices stub, and users who wanted to use Yices could download the real library and move the stub aside.  So stubbing was introduced exactly for the reason of allowing people to change the library used by their installation of BSC.

When BSC went open on GitHub, there were concerns about compiling/including STP.  The STP stub was introduced, probably as a simple way to satisfy those concerns at the time.  I can't recall if there was a conscious choice to go with a stub (vs just removing STP out of the build with macros).

Because, thinking about it now, the only reason to have a stub is because you want to change that choice later.  On the other github issue, you mentioned in a comment about it being a build option.  If not using STP is a build-time option, then we don't need a stub: the BSC executable could just be built to know that STP is not available and to not go looking for a library at all.

Vekhir:
I see. Up to now, the *_STUB options were only mentioned in the context of disabling, but it makes sense that there were other motivations behind them.

I can implement another option `STP_USE_LIB=placeholder` which essentially covers this use case. It is going to be pretty brittle though, as it happens when you tell users to "just replace the library". As such, it's probably a better choice to use the `system` or `absolute path` options if possible.

quark17:
Thank you for this, and sorry that it was neglected for so long.  I am going to close this PR, as Issue #595 has been resolved by PR #973.

The next step, in my opinion, is to eliminate the need for a stub: when we want to build BSC without STP, we can use preprocessor macros to eliminate any mention of the STP library, and then we don't need to link with it.

Your PR also added the option to link with an STP library from the system, rather than the library compiled from the vendor snapshot.  If someone needs that feature, we can add it, but it should be implemented simply by not building the STP snapshot and not installing it into the `inst` directory, so that it's not in the search path when BSC is linked (only the system library will be found in the search path).

PR #610, opened 2023-09-03
Disallow ° and ´ in BH identifiers

mieszko:
Following up on discussion in PR #601, the ° and ´ characters are no longer allowed to be parts of identifiers in BH; instead, they can now constitute parts of symbols.

### DIFF
--- a/src/comp/Lex.hs
+++ b/src/comp/Lex.hs
@@ -420,11 +420,6 @@ isSym _ = False
 isIdChar :: Char -> Bool
 isIdChar '\'' = True
 isIdChar '_' = True
-
--- \176 (hexB0) (octal 0260) is the degree symbol
-isIdChar '\176' = True
--- \180 (hexB4) (octal 0264) is the prime (acute accent) symbol
-isIdChar '\180' = True
 isIdChar c = isAlphaNum c
 
 isBinDigit :: Char -> Bool
--- a/testsuite/bsc.syntax/bh/DegreePrimeSymbol.bs
+++ b/testsuite/bsc.syntax/bh/DegreePrimeSymbol.bs
@@ -1,0 +1,13 @@
+package DegreePrimeSymbol((°´), (´°), (+´), (+°)) where
+
+(°´) :: a -> b -> (a, b)
+x °´ y = (x,y)
+
+(´°) :: a -> b -> (a, b)
+x ´° y = (x,y)
+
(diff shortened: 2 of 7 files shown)


quark17:
Thank you!  I am happy to merge this when the checks finish.

These symbols were at one time used by BSC when creating new identifiers.  They needed to be accepted by the BH parser back when BSC generated separate `.bi` and `.bo` files, where `.bi` files were a textual representation of the type declarations in the package -- see the `CI*` constructors in the CSynatx `CDefn`, sich as `CItype` and `CIclass` -- and the BH parser was used to read in the `.bi` files (there was a function `pIDefn` in the BH parser, alongside `pDefn`, at one time).  Since `.bi` files were removed (their info was moved into the binary `.bo` files, which has the benefit that additional info like position info is now preserved in identifiers), the parsing of these generated IDs is no longer needed.  So we can remove them from `isIdChar`.

Further, though, BSC no longer generates IDs with these symbols, anyway.  BSC normalizes all type constructors to take one argument, so for constructors of multiple arguments, it generates a struct type to become that one argument; the degree symbol was used in the name of the new struct, `<typename><degree><consname>`.  This is created by `mkTCId` in `Id.hs`, and it uses `fsTyJoin` as the joining string.  That was originally the degree symbol but was changed to `_$` (see `PreStrings.hs`).  It's possible to have a clash with a user declared type name, and we should probably find a better way, but this ought to be rare.

The acute symbol was briefly added in 2004 (but gone shortly after) as a character to go at the end of internally generated names, to guarantee that they wouldn't conflict with user names.  This was abandoned, but internally generated names could be identified by ID properties such as `IdPInternal`, `IdPGeneratedIfc`, `IdPRenaming`, `IdPParserGenerated`, `IdP_TypeJoin`.  Back in 2006, the idea was raised of making the `Eq` instance for `Id` consider IDs to be unequal if one was marked `IdPinternal` and the other was not, as a way to guarantee no name clash; but this didn't work at the time, as the property was getting lost somewhere, possibly when written to a `.bi` file and read back in.  Anyway, you can still find `Acute` mentioned in a few places in the source code, but it is now defined as `__`.  It is used to generate the fresh variable names in typechecking (`a__`, `b__`, etc).  And it's used in the BSV parser for creating renaming variables -- and I see that `CVPrint` has a clause that attempts to merge nested let-statement with "acute" names when pretty-printing, but I don't know if that pattern is still relevant.

Issue #224, opened 2020-07-31
Make .bs and .bsv support all the same pragmas, particularly verilog synthesis pragmas


bpfoley:
The BSV pragmas are defined in the BSV parser in Parser.BSV.CVParser.attribute2modpprop:
https://github.com/B-Lang-org/bsc/blob/master/src/comp/Parser/BSV/CVParser.lhs#L5067

The BS pragmas are more limited and are defined in Parser.Classic.CParser.pPragma:
https://github.com/B-Lang-org/bsc/blob/master/src/comp/Parser/Classic/CParser.hs#L604

These should be unified.

Clqssic supports the interface renaming pragmas, but these aren't documented.

Notably, `clocked_by`, `reset_by`, `default_clock_osc`, `default_reset` and `default_clock_gate` are missing from Classic.

Attached is an example test case.
[ClassicPragmas.bs.txt](https://github.com/B-Lang-org/bsc/files/5009191/ClassicPragmas.bs.txt)

quark17:
See Discussion #616 for some details on addressing this issue

PR #624, opened 2023-10-11
Add support for some clock and reset pragmas to Classic

nanavati:
This includes support for gate_input_clocks, clock_family, clock_prefix, reset_prefix and gate_prefix. Partially addresses #224 and fixes #616


testsuite/bsc.syntax/bh/bh_pragmas/UnknownClockFamily.bs
```diff
@@ -0,0 +1,21 @@
...
+interface Ticked =
+  ticked :: Bool
+
+{-# synthesize mkUnknownClockFamily {
+  gate_input_clocks = { default_clock },
+  clock_family = { default_clock, ungated } } #-}
+mkUnknownClockFamily :: (IsModule m mType) => Clock -> m Ticked
+mkUnknownClockFamily c = module
+  toggle :: Reg Bool <- mkReg False
+  toggle_delay :: Reg Bool <- clockedBy ungated $ mkRegU
```
quark17, line 15: The code also has an error, since it refers to a clock name that doesn't exist.  The pragma checking happens first, so this test works; but I wonder if it's worth keeping the code correct, and making the only error be in the pragma.  This could be either by removing the `clockedBy` or by keeping the `ungated` argument and referring to a different name in the pragma.  If you think it's worth it.

testsuite/bsc.syntax/bh/bh_pragmas/bh_pragmas.exp
```diff
@@ -0,0 +1,25 @@
...
+# (fixes Google issue #78233370)
+test_c_veri_bs_modules Pragmas {}
+
+compile_verilog_pass GateDefaultClock.bs
+compare_verilog sysGateDefaultClock.v
+
+compile_verilog_pass GateExplicitClock.bs
+compare_verilog sysGateExplicitClock.v
+
+compile_verilog_fail_error GateUnknownClock.bs P0182
```
quark17, line 11: For the compiles that fail with errors, it might be nice to check that the names that appear in the error messages are correct, not just the number.  But this is fine, if you prefer to keep it simple.

quark17 (commented):
This is awesome, thanks!

src/comp/PragmaCheck.hs
```diff
@@ -272,37 +272,45 @@ checkModuleArgPragmas pos pps_orig pps vtis =
...
+                clk_prefix = listToMaybe clk_prefixes
                 ps = filter (modifies idDefaultClock) pps
                 default_clock_osc =
                     listToMaybe $ catMaybes $
                       [ lookup idDefaultClock xs | PPclock_osc xs <- ps ]
-            in  case (clk_prefix, default_clock_osc) of
-                    (Just "", Nothing) ->
+            in  case (clk_prefixes, clk_prefix, default_clock_osc) of
+                    (_:_: _, _, _) ->
+                        Just (pos, EMultipleAttribute "clock_prefix")
```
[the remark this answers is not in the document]
nanavati: Done

testsuite/bsc.syntax/bh/bh_pragmas/UnknownClockFamily.bs
nanavati: Fixed. I decided to use the clock c rather than change anything else.

testsuite/bsc.syntax/bh/bh_pragmas/ClockFamily.bs
```diff
@@ -0,0 +1,21 @@
...
+clockedBy :: (IsModule m mType) => Clock -> m a -> m a
+clockedBy c = changeSpecialWires (Just c) Nothing Nothing
+
+interface Ticked =
+  ticked :: Bool
+
+{-# synthesize mkClockFamily {
+  gate_input_clocks = { default_clock },
+  clock_family = { default_clock, ungated } } #-}
+mkClockFamily :: (IsModule m mType) => Clock -> m Ticked
```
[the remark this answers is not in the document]
nanavati: I thought we mainly used `sys` for top-level testbenches and the like. I used `mk` here because I made a component meant to be instantiated in something larger.

testsuite/bsc.syntax/bh/bh_pragmas/bh_pragmas.exp
nanavati: Done

testsuite/bsc.syntax/bh/bh_pragmas/ClockFamily.bs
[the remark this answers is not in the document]
quark17: I will merge the PR as it is now; I don't want to keep making work for you.  But to answer your question: I guess I don't see the difference between this module named `mk` and all the other test files with modules named `sys`.  I don't see how either is more or less a submodule versus a top module.  I guess it really only matters for commands that expect a top-module (like `test_c_veri_bs`) and here we're just using `compile_verilog_pass`, but I tend to use the same structure even when I'm not using `test_*`, of making the top (or only) module in the file be named with `sys`, which is why I found this use of `mk` odd (plus the inconsistency with the other files, which are also not using `test_*`, but still named `sys`).

PR #626, opened 2023-10-13
Add missing features to the {-# properties foo = {...} #-} syntax in Classic

nanavati:


quark17:
Thank you again!  I'm happy to merge this when the checks finish.  A couple observations, though:

I was confused why you were comparing the Verilog for `mkSynthesize`, but it dawned on me that you need something to confirm that the module was synthesized (and decided to do that check this way?).  Maybe a comment would avoid the risk of someone in the future removing that comparison, thinking that it was unnecessary (as I did).

Also, FYI, BSV module attributes for clock and reset are tested in `bsc.mcd/Pragmas/` and `bsc.names/portRenaming/moduleArgs/`.  I don't have a proposal for how we could better organize the tests (to not be in multiple places and asymmetric between BH and BSV), so this current structure is fine.  The BSV tests are mostly testing behavior, so it's OK for the BH tests to just focus on syntax.

nanavati:
Added a comment about the mkSynthesize.v comparison.

PR #635, opened 2023-10-27
Fix type: s/typdef/typedef/

hce:
Fix a very simple typo in the reference guide, should be typedef, not typdef.


quark17:
Thank you!

PR #640, opened 2023-11-03
README.md: add package information about gentoo

vowstar:
- Changes:
    - add gentoo gitweb link about ``sci-electronics/bluespec``
    - add LiGurOS information
    - add repology list for bluespec
- Install ``sci-electronics/bluespec``:
    ```bash
    sudo eselect repository enable guru
    sudo emerge sci-electronics/bluespec
    ```
- View all installed bsc version:
    ```bash
    eselect bluespec list
    ```
- Switch between bsc versions:
    ```bash
    sudo eselect bluespec set bsc-2023.07
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ For the following systems, the Bluespec toolchain is available
 as a package that can be installed with the standard package manager:
 
 * ArchLinux AUR: [`bluespec-git`](https://aur.archlinux.org/packages/bluespec-git/) ([among others](https://aur.archlinux.org/packages/?K=bluespec))
+* Gentoo GURU and LiGurOS: [`sci-electronics/bluespec`](https://gitweb.gentoo.org/repo/proj/guru.git/tree/sci-electronics/bluespec) ([among others](https://repology.org/project/bluespec/versions))
 * Nix/NixOS: [`bluespec`](https://search.nixos.org/packages?channel=20.09&from=0&size=50&sort=relevance&query=bluespec)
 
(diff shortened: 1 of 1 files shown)


quark17:
Awesome, thank you!

For AUR, there are two links given: first it mentions the package for BSC (`bluespec-git`) and then in parentheses it adds "among others", with a link to search results for all AUR packages that are related to Bluespec.  The "among others" is meant to convey that there are other AUR packages besiges `bluespec-git` -- for example, packages for BDW (`bluespec-bdw-git`) and for the contrib libraries (`bluespec-contrib-git`), but there were also other packages for BSC (`bluespec-tools` and `bluespec-docs`), that were for the proprietary BSC and are presumably obsolete or broken now.

In your changes, you added an "among others" link next to the new entry for Gentoo.  However, there is only that one package for Gentoo, not others; and the link doesn't go to a search of Gentoo packages, but to a search of packages across all OSes (including ones mentioned on other lines).  This link (to Repology) is good to include, but since it isn't specific to Gentoo, it shouldn't go next to Gentoo.

I suggest moving that link to the text after the bullet list.  Adding something like, "You can also use the Repology search engine to check for Bluespec packages for your system."

Do you agree, or did I miss something?  And, if you agree, do you want to push a fix, or should I fix it up?  In any case, thanks again!

Also, if the text "among others" isn't clear, I'm happy to reword that.

vowstar:
@quark17 
What you said makes perfect sense. Thanks for reminding me. I have modified it according to your suggestion.

PR #642, opened 2023-11-08
CI: Workaround a PATH/ghc issue on macOS 11 VM

quark17:
Homebrew has stopped providing pre-built packages for macOS 11, so packages are built from source, which requires installing any tools needed to build the source.  Thus, ghc and cabal are now being installed in /usr/local/bin, which is earlier in the PATH than the ghcup bin directory, so the wrong GHC is found.  This workaround unlinks the specific ghc and cabal-install packages.  Another option would be to put ghcup earlier in the PATH.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,6 +129,9 @@ jobs:
         shell: bash
         run: |
           .github/workflows/install_dependencies_macos.sh
+          # Workaround for any packages that installed ghc
+          (brew list ghc > /dev/null 2>&1) && brew unlink ghc
+          (brew list cabal-install > /dev/null 2>&1) && brew unlink cabal-install
           # Don't rely on the VM to pick the GHC version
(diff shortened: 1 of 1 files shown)


quark17:
FYI, the way to prepend a directory to the path (the option that I didn't go with) is the following:
```
echo "$HOME/.ghcup/bin" >> $GITHUB_PATH
```
This affects the path seen by subsequent actions in the job; it does not change the path used by the current action.

quark17:
This was filed as issue 8738 in the actions/runner-images repo.  It has been fixed there, by unlinking ghc and cabal in the Haskell provision script for the VM (for macos11).  There's no need for our workflow to execute an unlink -- the packages should remain unlinked, regardless of any additional package installs that we might do.  The fixed runner is now live on GitHub, so our CI is working again without this change.  So I'm closing this PR without merging.

PR #655, opened 2023-12-26
allow any unqualified classic identifier in interface `arg_names`

### LINKED ISSUE #654: uppercase names not allowed in interface property `arg_names`
the `arg_names` property of interface fields does not allow names starting with uppercase:
```
interface Foo =
    foo :: Bool -> Action {-# arg_names = [FOOBAR] #-}

{-# synthesize mkFoo #-}
mkFoo :: Module Foo
mkFoo = return _
```

this is the error:
```
Error: "Foo.bs", line 4, column 49: (P0005)
  Unexpected "]"; expected "."
```

it seems that the parser thinks that `FOOBAR` is a package name, and expect it to start a qualified identifier. and indeed the relevant fragment of `CParser.hs` uses `pFieldId` for this:

```
    literal (mkFString "arg_names") ..+ eq ..+ lb ..+  pFieldId `sepBy` cm +.. rb `into`
                (\a -> succeed $  [(PIArgNames a)])
```

but of course things with dots don't make sense to generate as a Verilog identifier, and if you replace `FOO` with `FOO.bar` then the compiler indeed rightfully complains about this:

```
Error: "Foo.bs", line 4, column 4: (G0106)
  Method `foo' generates a port with name `foo_FOO.bar' which is not an
  accepted Verilog identifier. It must contain only letters, digits, $ and _,
  but cannot begin with $ or a digit.
```

interestingly `PIArgNames` takes `[Id]` whereas all other properties take `String`; not sure why it's not `[String]` but perhaps there is some deep reason somewhere?

mieszko:
the `arg_names` property in interface type declarations now allows any unqualified bs classic identifier, including identifiers starting with uppercase letters.

minimal fix; perhaps a better fix would be to change `PIArgNames` to take `[String]` rather than `[Id]` (see #654), but that requires a lot more digging in bsc internals.

fixes #654.

### DIFF
--- a/src/comp/Parser/Classic/CParser.hs
+++ b/src/comp/Parser/Classic/CParser.hs
@@ -349,7 +349,7 @@ prefix = literal (mkFString "prefixs") ||| literal (mkFString "prefix")
 pIfcPragmas :: CParser [IfcPragma]
 pIfcPragmas =
     -- arg_name = [id,id]
-    literal (mkFString "arg_names") ..+ eq ..+ lb ..+  pFieldId `sepBy` cm +.. rb `into`
+    literal (mkFString "arg_names") ..+ eq ..+ lb ..+  varcon `sepBy` cm +.. rb `into`
                 (\a -> succeed $  [(PIArgNames a)])
     -- prefix = "str"
     ||!  prefix ..+ eq ..+ varString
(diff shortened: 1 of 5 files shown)


quark17:
This looks good to me.  I'm happy for it to be merged when the checks pass.

PR #656, opened 2023-12-26
Fix scheduling annotations for methods with actions conditional on the arguments

### LINKED ISSUE #641: Method with internal input-dependent "if" mistakenly scheduled as SBR to self
Given this module

```bsv
interface Meow;
    method Action meow(Int#(32) value);
endinterface

(* synthesize *)
module mkPositiveMeow(Meow);
    Reg#(Int#(32)) r <- mkReg(0);

    rule disp;
        $display(r);
    endrule

    method Action meow(Int#(32) value);
        if (value > 0) begin
            $display("actually set");
            r <= value;
        end
    endmethod
endmodule
```

Semantically, `meow(1)` then `meow(-1)` results in `r` being set to `1`, whereas `meow(-1)` leaves `r` unchanged. This means that `meow C meow`. However BSC infers `meow SBR meow`:

```
=== Generated schedule for mkPositiveMeow ===

Method schedule
---------------
Method: meow
Ready signal: True
Sequenced before (restricted): meow
 
Rule schedule
-------------
Rule: disp
Predicate: True
Blocking rules: (none)
 
Logical execution order: disp, meow

==============================================
```

The incorrect behavior can be seen with this testing module:

```bsv
(* synthesize *)
module mkMeowTwice();
    Meow meow <- mkPositiveMeow;

    (* fire_when_enabled *)
    (* execution_order = "meow1, meow2" *)
    rule meow1;
        $display("set 1");
        meow.meow(1);
    endrule

    (* fire_when_enabled *)
    rule meow2;
        $display("set -1");
        meow.meow(-1);
    endrule
endmodule
```

If using `-sim`, or `-verilog` but with `(* synthesize *)` on `mkPositiveMeow` removed, `r` is correctly set to `1` as expected:

```
          0
set 1
actually set
set -1
          1
set 1
actually set
set -1
...
```

With `(* synthesize *)` on `mkPositiveMeow` and `-verilog`, `meow1` and `meow2` are somehow still both schedulable in the same cycle, but `r` incorrectly stays at `0`

```
set 1
set -1
          0
          0
set 1
set -1
set 1
set -1
          0
          0
...
```

Inspecting the generated `mkMeowTwice.v` reveals that it has indeed incorrectly merged `meow(1)` and `meow(-1)` into `meow(-1)` based on the incorrectly inferred `meow SBR meow`:

```verilog
  // submodule meow
  mkPositiveMeow meow(.CLK(CLK),
		      .RST_N(RST_N),
		      .meow_value(meow$meow_value),
		      .EN_meow(meow$EN_meow),
		      .RDY_meow());

  // submodule meow
  assign meow$meow_value = 32'hFFFFFFFF ;
  assign meow$EN_meow = 1'b1 ;
```

bsc version: 2023.07

quark17:
Fixes #641.

There is some reorganization of the code in `ASchedule` that constructs the three conflict graphs (CF, SC, and PC, indicating the scheduling limitations between rules/methods).  This code contained forced evaluations of the graphs, to prevent memory usage from growing too much (due to the graphs being contructed lazily), and the location of that forcing has moved -- the graphs are constructed in steps (first the initial graph based on pairwise method calls then user pragmas then methods with arguments), and the forcing is now after the first step, not the third, but the first step is the most significant, so experiments suggest that forcing there has the same memory footprint.  (It's also probably important to force each graph before starting the next, so all three aren't in memory at once.)  If any issues are detected, we can profile the example and revisit this choice.

There was already a step that was adding conflict edges for ActionValue methods where arguments are used in the return value.  That just needed to be extended to also check for Action/ActionValue methods where arguments are used in the condition of an action.  The effort of both checks is now avoided by first looking for whether a conflict edge already exists for the method relative to itself, and skipping the checks if so.  (This was the motivation for the reorganization: so that an initial SC graph was available for lookup when computing the edges to add.)

Test case added.

There was only one group of existing tests affected by the change.  It was an example where a method was writing into a vector of registers and the specific register index was coming from an argument:
```
Vector#(Size, Reg#(Bool))  rgs <- replicateM( mkReg(False) ) ;

method Action set (Index idx);
  rgs[idx] <= True;
endmethod
```
BSC was inferring SBR for this method, but that would result in incorrect code-generation for a parent module that called the method twice with different argument values: the first call would be ignored and only the register at the second index would be updated.  With this PR, BSC infers C by inserting a conflict edge (for the method with itself) in the SC conflict graph.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,8 +44,8 @@ jobs:
         run: |
           sudo .github/workflows/install_dependencies_ubuntu.sh
           # Don't rely on the VM to pick the GHC version
-          ghcup install ghc 9.4.7
-          ghcup set ghc 9.4.7
+          ghcup install ghc 9.4.8
+          ghcup set ghc 9.4.8
       # Until BSC uses cabal to build, pre-install these packages
       - name: Install Haskell dependencies
         shell: bash
@@ -81,11 +81,11 @@ jobs:
       # reflect the locations of any new haskell sources
       - name: Test Haskell Language Server
         run: |
-          ghcup install hls
+          ghcup install hls 2.5.0.0
           pip3 install pyyaml
           python3 util/haskell-language-server/gen_hie.py
           pushd src/comp
-          haskell-language-server-9.4.7 bsc.hs
+          haskell-language-server-9.4.8 bsc.hs
           popd
       # Check that .ghci has all the right flags to load the source.
       # This is important for text editor integration & tools like ghcid
@@ -138,8 +138,8 @@ jobs:
             export PATH=$HOME/.ghcup/bin:$PATH
           fi
           # Don't rely on the VM to pick the GHC version
-          ghcup install ghc 9.4.7
-          ghcup set ghc 9.4.7
+          ghcup install ghc 9.4.8
+          ghcup set ghc 9.4.8
       # Until BSC uses cabal to build, pre-install these packages
       - name: Install Haskell dependencies
         shell: bash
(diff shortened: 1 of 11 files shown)


quark17:
I realized that the new code (for computing the method-arg conflict edges) should go in a separate function, to abstract it away and reduce clutter.  So I pushed a new commit that does that.

PR #697, opened 2024-05-15
1. update for emacs bsv-mode load fail and some elisp script complaint

microway199x:
1. update for emacs bsv-mode load fail for `which-func-modes default is `t ,
   not a list; add condition for only `user config which-func-modes to list,
   add element to it;
2. updat for emacs complain with bsv-mode, unexpected "?;", except "?\;"


quark17:
Thank you for submitting this!  I am not familiar with elisp, so I'm hoping that maybe @mieszko or someone can look at it and comment.  I think @kenta2 might also have planned to test it out.

kenta2:
Looks good.

PR #700, opened 2024-05-25
Add CI for new ubuntu-24.04 beta VM

quark17:
GitHub now provides an `ubuntu-24.04` runner, in beta, so this adds that to the matrix for Ubuntu testing.

### DIFF
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -20,7 +20,7 @@ jobs:
   build-macOS:
     name: "Build: ${{ inputs.os }} ghc-${{ inputs.ghc_version }}"
     runs-on: ${{ inputs.os }}
-    timeout-minutes: 90
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
       - name: Checkout submodules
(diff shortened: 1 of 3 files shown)


quark17:
I've included a commit to remove `macos-11` from the testing matrix, since the build is now failing due to exceeding the timeout.  A combination of (1) that Homebrew no longer supports installing binary packages for macOS 11, so it had to compile from source; (2) that we added `brew update` to workaround an issue with GitHub's runners; and (3) that more packages have new versions since the `macos-11` runner snapshot was made; means that the `macos-11` runner is now exceeding the 90 minute timeout just in the "install dependencies" stage.  We could increase the timeout, but that would be time spent in our CI each time it runs (and holding up other macOS jobs), so it's probably just time that we remove `macos-11`.

(The reason why so many packages are being installed is because we've run `brew update` before runing `brew install` for a large list of dependencies.  We only needed `update` for the `mactex` package, so an option might be to run `install` for everything else, then run `update`, then `install` just for `mactex` -- which should greatly reduce the number of dependent packages that will get installed at newer versions.  But it's probably not worth the time to experiment with it.)

Issue #698, opened 2024-05-20
GCC 14 emits extra warnings


Vekhir:
Hi,
when linking a BSV project using `bsc` and Bluesim with GCC 14.1, the following warning occurs in several places:
```
In file included from /opt/bluespec/lib/Bluesim/bluesim_primitives.h:6,
                 from ./model_mkFibOne.cxx:7:
/opt/bluespec/lib/Bluesim/bs_prim_mod_reg.h:22:14: warning: template-id not allowed for constructor in C++20 [-Wtemplate-id-cdtor]
   22 |   MOD_Reg<T>(tSimStateHdl simHdl, const char* name, Module* parent,
      |              ^~~~~~~~~~~~
/opt/bluespec/lib/Bluesim/bs_prim_mod_reg.h:22:14: note: remove the ‘< >’
```

### Description
[BDW](https://github.com/B-Lang-org/bdw) provides for development two modes for compilation/simulation: Bluesim and Verilog. The Verilog mode works fine, the Bluesim mode however emits above warnings (and produces a binary which doesn't run, but that's a separate issue).

#### Steps to reproduce
1. Copy [examples/smoke_test/FibOne.bsv](https://github.com/B-Lang-org/bsc/blob/main/examples/smoke_test/FibOne.bsv) into a local directory.
2. Run `bsc -u -sim -simdir . -bdir . -info-dir . -keep-fires -p %/Libraries -g mkFibOne FibOne.bsv`
3. Run `bsc -e mkFibOne -sim -o ./out -simdir . -p %/Libraries -bdir . -keep-fires`

These warnings also lead to test failures when running `make check` whereever the output is tested. Afaict, the failing tests are all affected by this mismatch.
They also show this warning:
```
Bluesim object created: model_mkTbGCD.{h,o}
../../../../src/bluesim/event_queue.cxx:8:24: warning: ‘bk_clock_name’ violates the C++ One Definition Rule [-Wodr]
    8 | extern "C" const char* bk_clock_name(tClock handle);
      |                        ^
../../../../src/bluesim/kernel.cxx:1093:13: note: type mismatch in parameter 1
 1093 | const char* bk_clock_name(tSimStateHdl simHdl, tClock clk)
      |             ^
../../../../src/bluesim/kernel.cxx:1093:13: note: type ‘struct tSimState *’ should match type ‘tClock’
../../../../src/bluesim/kernel.cxx:1093:13: note: ‘bk_clock_name’ was previously declared here
[...]

PR #702, opened 2024-05-27
Bluesim: Fix function call arguments in debug function

quark17:
The `EventQueue::print()` function was calling `bk_clock_name()` with the wrong number of arguments, resulting in the warning reported in #698:
```
Bluesim object created: model_mkTbGCD.{h,o}
../../../../src/bluesim/event_queue.cxx:8:24: warning: ‘bk_clock_name’ violates the C++ One Definition Rule [-Wodr]
    8 | extern "C" const char* bk_clock_name(tClock handle);
      |                        ^
../../../../src/bluesim/kernel.cxx:1093:13: note: type mismatch in parameter 1
 1093 | const char* bk_clock_name(tSimStateHdl simHdl, tClock clk)
      |             ^
```
If the `print()` function were ever called, this would probably segfault.  Fortunately, `print()` is not used anywhere, as it only exists for use when debugging.

Originally, Bluesim only had one global simulation state, but to support having multiple simulations running in the same program, the Bluesim kernel functions were changed to take a pointer to the simulation state (`tSimStateHdl`).  It's likely that, during that change, the `print()` function -- and the `extern` declaration for `bk_clock_name` -- was overlooked.

This PR fixes the `extern` declaration to have the right arguments, by adding a `tSimStateHdl` argument.  The `print()` method then needs to have a `tSimStateHdl` argument, to be able to pass to the call to `bk_clock_name`.

That resolves the warning.  Although we also could have resolved the warning by deleting `print()` and the `extern` declaration, since it's unused.

I notice that many functions in the `EventQueue` class take the `tSimStateHdl` as an argument.  I don't believe that we need a single `EventQueue` to support multiple simulations, so it might make sense for the `EventQueue` constructor to take the simulation state and store it, and then the function calls wouldn't need it as an argument.  That would be a bigger change than I'm prepared to do now, though.

### DIFF
--- a/src/bluesim/event_queue.cxx
+++ b/src/bluesim/event_queue.cxx
@@ -5,7 +5,7 @@
 
 /* we need these for the debugging routines only */
 #include "priority.h"
-extern "C" const char* bk_clock_name(tClock handle);
+extern "C" const char* bk_clock_name(tSimStateHdl simHdl, tClock handle);
 
 
 /* Fundamental heap operations */
@@ -222,7 +222,7 @@ void EventQueue::clear()
 }
 
 /* Print the event queue contents (for debugging) */
-void EventQueue::print() const
+void EventQueue::print(tSimStateHdl simHdl) const
 {
   printf("Event queue:\n");
   for (unsigned int i = 0; i < count; ++i)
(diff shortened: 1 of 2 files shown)


Vekhir:
I can confirm that this commit removes the warning and lets the tests pass.

PR #721, opened 2024-07-21
bsc: remove use of `old-time`

thoughtpolice:
A long time coming (no pun.) This removes all uses of `old-time` from the compiler and replaces it with equivalents from `time`; eliminating a 3rd party dependency needed on top of GHC. Most of the translations are mechanical and rather straight forward.

At the same time, I also:

- Removed some extra use of `TimeInfo` in `SimBlocksToC`
- Removed all references to `ClockTime` (2 extra lines of code in `GenABin`)


src/comp/ABin.hs
```diff
@@ -49,7 +49,6 @@ data ABinModInfo =
         -- the name of the BSV package which defined this module
         abmi_src_name :: String,
         -- time when BSC was called to compile the .ba
```
quark17, line 51: This accompanying comment would need to be removed too

src/comp/Makefile
```diff
@@ -92,7 +92,7 @@ $(warning Building unoptimized to work around a bug in GHC 9.2.1)
 GHCOPTLEVEL ?= -O0
 endif
 
-GHC += -Wtabs -fmax-pmcheck-models=800
+GHC += -Wtabs -fmax-pmcheck-models=800 -freverse-errors
```
quark17, line 95: Does this GHC flag change belong in this PR?

src/comp/SimBlocksToC.hs
```diff
@@ -489,12 +489,15 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
         get_creation_time = function (userType "time_t")
                                      (mkScopedVar "get_creation_time")
                                      []
-        (TimeInfo _ clock_time@(TOD t _)) = if (timeStamps flags)
-                                            then creation_time
-                                            else TimeInfo 0 (TOD 0 0)
-        time_str = calendarTimeToString (toUTCTime clock_time)
+
+        ( time_str, time_secs ) = case mcreation_time of
+          Nothing -> ( "1970-01-01 00:00:00 UTC", 0 )
```
quark17, line 494: Would it be better to create a `UTCTime` value representing this and call `show` on it?

src/comp/SimBlocksToC.hs
```diff
@@ -489,12 +489,15 @@ convertSchedules flags creation_time top_id def_clk def_rst sb_map ff_map
         get_creation_time = function (userType "time_t")
                                      (mkScopedVar "get_creation_time")
                                      []
-        (TimeInfo _ clock_time@(TOD t _)) = if (timeStamps flags)
```
quark17, line 492: Why did you remove the flag check from here and put it in the caller?

src/comp/bsc.hs
```diff
@@ -1339,10 +1339,14 @@ genModuleC errh flags dumpnames time0 toplevel abis =
        start flags DFsimBlocksToC
        let sb_map = M.fromList $ map (\sb -> (sb_id sb,sb))
                                      (simblocks_opt ++ primBlocks)
-           creation_time = time
```
quark17, line 1342: This adds the flag check that was removed from `simBlocksToC`.  But you also added a call to `getCurrentTime` -- what was your thinking for that?

src/comp/bsc.hs
```diff
@@ -1806,10 +1810,8 @@ cxxLink errh flags toplevel names creation_time = do
     unless (quiet flags) $ putStrLnF ("Simulation shared library created: " ++ soFile)
     -- Write a script to execute bluesim.tcl with the .so file argument
     let bluesim_cmd = "$BLUESPECDIR/tcllib/bluespec/bluesim.tcl"
-        (TimeInfo _ (TOD t _)) = creation_time
-        time_flags = if (timeStamps flags)
-                     then [ "--creation_time", show t]
-                     else []
+        time_flags = maybe [] (\t -> ["--creation_time", show t]) mcreation_time
```
quark17, line 1813: Ah, this explains why you moved the flag check: you unified two flag checks into one, which has to be done in the caller, and the `Maybe` result is passed to both places.

This use of `maybe` is less readable than the if-then-else, in my opinion.

quark17 (commented):
At the moment, BSC compiles with all versions of GHC back to 7.10.3 (which GHCUP still offers).  This PR will presumably restrict us to newer GHC versions -- but can you saw what that earliest GHC version is?  I attempted to build with 7.10.3 and got an error about `nominalDiffTimeToSeconds` not in scope, but actually I'm surprised that it didn't say that `Data.Time` doesn't exist, since that's not listed in the base -- so where is it coming from, and at what version?  Anyway, `nominalDiffTimeToSeconds` is available since 1.9.1 of `Data.Time` -- but where do I find what version is available to GHC versions, if not from the base version?

I haven't looked deeply at what you're doing, but have given some comments.  I see that some tests fail, for BSC and BDW.  Do you know what those are about?

Vekhir:
> At the moment, BSC compiles with all versions of GHC back to 7.10.3 (which GHCUP still offers). This PR will presumably restrict us to newer GHC versions -- but can you saw what that earliest GHC version is? I attempted to build with 7.10.3 and got an error about `nominalDiffTimeToSeconds` not in scope, but actually I'm surprised that it didn't say that `Data.Time` doesn't exist, since that's not listed in the base -- so where is it coming from, and at what version? Anyway, `nominalDiffTimeToSeconds` is available since 1.9.1 of `Data.Time` -- but where do I find what version is available to GHC versions, if not from the base version?

FYI, you can see the boot libraries shipped with each GHC version since 7.0.1 here: https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history

Going to the `time` row shows that it was always shipped (1.2.0.3 in 7.0.1), so that's why `Data.Time` exists. However, GHC 7.10 only ships with `time 1.5.0.1`, so `nominalDiffTimeToSeconds` isn't available there; Only 8.8.1+ ship with `time 1.9.1` or later.

PR #722, opened 2024-07-21
comp: remove most imports of `Data.Generics`

thoughtpolice:
99% of the uses of the `Data.Generics` import from `syb` was to qualify the names `Data` and `Typeable`, but these have long since been available directly in `base` along with `DeriveDataTypeable` for many years now. The `syb` exports are in fact just re-exports from base. Migrate most of the code to just use `Data.Data` directly instead.

At the same time, this touches up some of the import lists to be a little more consistent but is otherwise functionally identical.

### DIFF
--- a/src/comp/ISyntax.hs
+++ b/src/comp/ISyntax.hs
@@ -82,9 +82,10 @@ module ISyntax(
 import Prelude hiding ((<>))
 #endif
 
-import System.IO(Handle)
-import qualified Data.Map as M
+import Data.Data
 import Data.List(intercalate)
+import qualified Data.Map as M
+import System.IO(Handle)
 
 import qualified Data.Array as Array
 import IntLit
@@ -112,7 +113,6 @@ import Error(internalError, EMsg, ErrMsg(..))
 import PFPrint
 import IStateLoc(IStateLoc)
 import IType
-import qualified Data.Generics as Generic
 
 -- ============================================================
 -- IPackage, IModule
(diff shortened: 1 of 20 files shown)


src/comp/ISyntax.hs
```diff
@@ -754,11 +754,18 @@ type PSet a = S.Set a
 data PTerm a = PAtom (IExpr a)
              | PIf (IExpr a) (Pred a) (Pred a)
              | PSel (IExpr a) Integer [Pred a]
-        deriving (Eq, Ord, Show, Generic.Data, Generic.Typeable)
+        deriving (Eq, Ord, Show, Data, Typeable)
 
 -- ==============================
 -- IConInfo
 
+-- FIXME: syb includes an orphan to work around the fact Handle needs a Data
```
quark17, line 762: And the only place we have `Handle` in a data structure needing `Data` is here in `IConInfo` (`ICHandle`), so you've chosen to put the instance here?  Ok.

quark17 (commented):
Does "long since" include GHC 7.10.3, or will this increase the minimum GHC version?  If so, do you know what version?

Vekhir:
> Does "long since" include GHC 7.10.3, or will this increase the minimum GHC version? If so, do you know what version?

`Data.Data` exists since `base 4.6.0.1`: https://hackage.haskell.org/package/base-4.6.0.1/docs/Data-Data.html#t:Data
That's the oldest I could find. This corresponds to GHC 7.6 or later, so the current bounds are fine.
Not done here, but the `DeriveDataTypeable` language extension is enabled by default since GHC 7.10 ([source](https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-Typeable.html)). It's mentioned in the OP.

PR #724, opened 2024-07-23
Add a type class for showing values at elaboration time

krame505:
There isn't currently really a great way of printing out structured data at elaboration time for debugging purposes.  I made a type class `SShow` mirroring `CShow`, except that it generates a `String` instead of `Fmt` for printing at elaboration time.


quark17:
I had commented offline that it's unfortunate that this `SShow` library duplicates so much from `CShow` and wondered if one could be defined in terms of the other.  I think I suggested that a function from `Fmt` to `String` could be used to define `SShow`, but I think you responded by pointing out that non-runtime types like `List` wouldn't have `CShow` instances.  So then what about the other direction -- can we define `CShow` as just printing the string from `SShow`?  Or can the choice of `show` versus `$format` be abstracted out, to have a common underlying polymorphic implementation?  If not, this amount of redundancy isn't too bad, so I'm OK with accepting the PR as it is, if that's easiest.

FYI, I notice that there's no instance for the `Array` primitive -- is that possible to add?

quark17:
Oh, also, could you add a test case?

krame505:
There isn't a way to go from a runtime value to an elaboration-time `String`, either.  If we had a typeclass `StringLike` like in Haskell, we could maybe abstract out the generic bit for showing data/structs/vectors, but I'm not sure how to define the primitive instances in that approach.

krame505:
I added the `Array` instance and some test cases.

quark17:
@krame505 Hey, thanks for updating the PR.  However, in the future, can you update with rebase rather with a merge commit?  It shouldn't be a problem on this PR, since we'll probably squash merge it rather than preserve any commit history; but in general I'd like to avoid merge commits.

For accepting a PR, there are three options: create a merge commit; squash and merge; rebase and merge.  This week, I changed the settings for the BSC repo to disallow "Create a merge commit" for accepting PRs.  I also enabled the feature that displays an "Update branch" button on PRs -- perhaps that's what you used to update the PR?  There's a little pull-down arrow next to the button that lets you select the method for merging, either with a commit or by rebase.  Unfortunately, even though merge commits are disable for accepting PRs, the merge-commit option for updating the branch is still enabled, and is the default.  When updating a PR branch, if people could click the pull-down arrow and select "Update with rebase", that'd be appreciated.

krame505:
Sure, will do in the future.  Yes, I just hit "update branch" without thinking about whether it would rebase.  I agree this PR should probably just be squash merged anyway, so it doesn't matter in this case.

Issue #714, opened 2024-07-12
Better control of wrapper generation using type classes


krame505:
This is related, but essentially orthogonal to the idea in #713. This is a bit less well-formed of an idea but I wanted to get some notes down before the meeting today.

The motivation is that the wrapper generation logic is rather complex and inflexible.  Specifically, the inputs and outputs of action methods are not split into separate ports, which makes interfacing with the generated verilog rather tedious.

The high-level idea is to have a type class `Wrapped` that maps interface types into a type-level description of the wrapper that should be generated.  The description types would contain of tuples of the ports that should be generated.  Wrapper generation would then consult the `Wrapped` type class to determine what ports should be generated.
```haskell
data (Port :: $ -> # -> *) name n = Port (Bit n)

class Wrapped a b | a -> b where
    wrap :: a -> b
    unwrap :: b -> a
```
For example, consider the following interfaces:
```haskell
interface Foo =
  a :: UInt 8
  b :: Bool
 deriving (Bits)

interface Bar =
  c :: Foo
  d :: ActionValue Foo
  e :: Foo -> Action
```
The wrapper representation for `Foo` (corresponding to the way wrapper generation currently happens) would be
```
(Port "a" 8, Port "b" 1)
```
and `Bar` would be
```
(Port "c_a" 8, Port "c_b" 1, ActionValue (Port "d" 9), (Port "e_1" 9) -> Action)
```
It should be possible to generate these instances using generics (with a few complications, that I will get to in a moment.)  There could also be an instance of `Wrapped` for `Vector` that adds the current behavior of unfolding vectors of interfaces into separate ports.  Wrapper generation from these representation types is essentially trivial; the `to` and `from` methods can also make use of the `wrap` and `unwrap` methods in the `Wrapped` type class.

The interesting part is that we can define some newtype wrappers for how interface fields/method parameters should be treated, e.g. to indicate that a parameter/action result should be recursively expanded into separate ports:
```haskell
interface Baz =
  f :: ActionValue (Expand Foo)
  g :: (Expand Foo) -> Action
```
The type `Expand` has a special case instance for `Wrapped` that give the representation type for `Baz`:
```haskell
(ActionValue (Port "f_a" 8, Port "f_b" 1), (Port "g_a" 8, Port "g_b" 1) -> Action)
```
which would yield the wrapper
```
interface BazWrap =
  f_a :: Bit 8
  f_b :: Bit 1
[...]

PR #729, opened 2024-08-20
Support splitting up struct method parameters into multiple input ports

krame505:
After some back and forth with @nanavati, we ended up going with a design that is closer to my original proposal in #714. It is really hard to do port splitting in `GenWrap.hs` with the degree of flexibility that we want, and there are some limitations like not being able to resolve type aliases.  Not to mention that code is incredibly tedious and hard to modify, and would only make it harder to implement features like methods with multiple output ports.  Doing this with type classes makes the logic a bit more transparent and easier to modify in the future.

There are several significant changes bundled together in this pull request:
1. A significant refactor to wrapper generation, using a type class to determine the types of fields in wrapper interfaces, replacing some of the hard-coded logic in `GenWrap.hs`;
2. Move port type saving logic out of the evaluator and genwrap into the type class;
3. Pass method argument names by tagging method fields with a new primitive instead of via `BetterInfo` (this allows determining input port names at elaboration time);
4. Move the sanity checks for port name collisions to post-elaboration;
5. Determine how method arguments are split into ports using a type class;
6. Add some port splitting utilities using generics.

## Implementation
I'm not changing the fundamental structure of wrapper interfaces, only changing how field types are determined - flattening of nested interfaces still happens in `genwrap` as usual.  The types of fields in a wrapper interface are determined by the typeclass `WrapField`.  The `toWrapField` method converts from a field in the original interface (e.g. `Int 8 -> ActionValue Bool`) to the type of field in the wrapper interface (e.g. `Bit 8 -> ActionValue_1`), while `fromWrapField` is the inverse of this.  The special cases for `Clock`, `Reset` and `Inout` are also handled by `WrapField`.  

`WrapField` uses a type class `WrapMethod` to compute the wrapped type of a method field.  This type class uses the `SplitPorts` type class to convert the type of a method input into a tuple of `Port`s, and the `WrapPorts`  type class converts this into a tuple of `Bit` values.  (Thus one determines how a method argument type should be split into ports by defining instances of `SplitPorts`.  More on that later.)  The individual tuples of method argument `Bit` values then get turned into a single curried function type for the wrapper interface method.  

These type classes are also used to compute the names of input ports.  This is happening on the value level as lists of strings[^listn], not as type-level strings as I was originally thinking. I hit some sort of snag with this, although I don't remember exactly what it was, and being able to just compute this in the evaluator seemed like less of a pain than dealing with type-level lists of strings, adding type-level number to type-level-string conversion, etc.[^note]  The only reason a type-level string is there in `WrapField` is to give the original field name to generate a better error message when context resolution fails due to a type not being in the `Bits` type class.

[^listn]: Really, this should be using `ListN` to ensure that the list of port names is always the correct length.  But sadly that doesn't exist in the Prelude, and `SplitPorts` needs to be.

The list of argument names are tagged on to the wrapper method function/value with `primMethod`.  The evaluator now expects this primitive to exist on method fields in `iExpandField`.[^vMkRWire1]  We could potentially stick additional metadata that is computed at elaboration-time here in the future.  The field name/`result` pragma and the `arg_names` pragma (if present) are passed as arguments to `toWrapField`, which are used to compute the base names of input ports, which are and tagged on to the converted method value.

[^vMkRWire1]: This required a corresponding tweak in `vMkRWire1`, which is a handwritten interface LARPing as a generated wrapper interface, to be instantiated way later by the scheduler.  

Because port names are now determined at elaboration time, I had to move the port name collision checks to after elaboration.  This is maybe slightly less nice as some error messages show up latter, but this sort of error isn't super common.  It does feel like a more natural place to implement these checks anyway, instead of needing to figure out the port names from the pragmas before type checking.

Saving port types, on both sides of the synthesis boundary, is also handled via these type classes.  See the `saveFieldPortTypes` method in `WrapField` type class.  Calls to this method get inserted in both genwrap and wrappergen.  This method also requires the same field naming arguments as `toWrapField`.  I considered making `toWrapperField`/`fromWrapperField` be in the `Module` monad and do the port type saving too, but that complicates the code generation in genwrap a fair bit as every field value needs to be bound in a giant `do`-block.  

[^note]:  Although in retrospect the various tuple shenanigans I needed were just as complicated, and I could maybe have just stuck the port name in the `Port` type constructor.  So I'm not sure if it ended up being much simpler.  Having the evaluator is more flexible, at least.

## Specifying port splitting
How a method argument type gets split up, and how the resulting ports are named is determined by the `SplitPorts` type class.  There is a default instance that doesn't do any flattening, which preserves the current behavior:
```
instance SplitPorts a (Port a) where
  splitPorts = Port
  unsplitPorts (Port a) = a
  portNames _  base = Cons base Nil
```

If we have a struct
```
struct Bar =
  v :: Vector 3 Bool
  w :: (Bool, UInt 16)
  z :: Foo

interface Top = 
  putBar :: Bar -> Action
```
then for `putBar` to have separate input ports for each field, we need an instance
```
instance SplitPorts Bar (Port (Vector 3 Bool), Port (Bool, UInt 16), Port Foo) where
  splitPorts (Bar { v = v; w = w; z = z; }) = (Port v, Port w, Port z)
  unsplitPorts (Port v, Port w, Port z) = Bar { v = v; w = w; z = z; }
  portNames _ base = Cons (base +++ "_v") $ Cons (base +++ "_w") $ Cons (base +++ "_z") Nil
```
One can write this sort of instance explicitly.  However there are a few ways that this can be done with less boilerplate.  

I added a library `SplitPorts` in Base1 with a couple of utility type classes.  `ShallowSplitPorts` uses generics to flatten out a struct by one level, using the `SplitPorts` instances for each of its fields.  One can use these to define a `SplitPorts` instance:
```
instance (ShallowSplitPorts Bar p) => SplitPorts Bar p where
  splitPorts = shallowSplitPorts
  unsplitPorts = shallowUnsplitPorts
  portNames = shallowSplitPortNames
```
This would be a bit nicer to use if we had `deriving via`.  In fact, I'm wondering if we should make `derive SplitPorts` generate the above sort of instance automatically.  

`DeepSplitPorts` fully flattens a struct, including nested struct, tuple and `Vector`[^vectorsplit] fields, down to primitives and types with multiple constructors.  When using this type class, if one wishes for some nested struct type not to be flattened, they can define a `DeepSplitPorts` instance that does nothing to prevent this.

[^vectorsplit]: Making this work reasonably for large vectors required a fun bit of awesomeness in the `ConcatTuple` type class I added, which converts a vector of tuples to and from a flattened vector.

Sometimes one might wish for a type to be flattened in only some places.  Instead of defining a `SplitPorts` instance, you can insert the `ShallowSplit` or `DeepSplit` "newtype" wrapper on your interface method parameters:
```
interface Top = 
  putBar :: DeepSplit Bar -> Action
```

I added test cases illustrating all these different patterns/approaches.  There are probably more possibilities and I'm not sure what will prove to be the most ergonomic in practice, but these utilities are easy to add/change later.  

## Future considerations
I designed this with support for methods with multiple output ports in mind, which I may or may not attempt next depending on how much time I have.  The `SplitPorts` type class could be reused to also determine how results of value/`ActionValue` methods are split into output ports.  

I'm not quite sure what the wrapper type representation looks like for types with multiple output ports.  Just using a tuple of `Bit ` values for methods with multiple output ports might work for value methods, but `ActionValue_` only accepts a single numeric size parameter.  My current thinking is that we should ditch `ActionValue_` and have
a struct `PrimValue :: (# -> * -> *) n a ` that tags a `Bit n` value onto a chain of output values `a`, ending in a `PrimAction` or `PrimUnit`.  

## Remaining issues
The error message when a method yields a port that isn't in `Bits` is fine, but there is another error message about a `Bits` context that didn't reduce, with unknown position.  See for instance [testsuite/bsc.verilog/noinline/NoInline_ArgNotInBits.bsv.bsc-vcomp-out.expected](https://github.com/krame505/bsc/blob/0ea992e85ae2852a9e204e5bae907c8bfbb3e07d/testsuite/bsc.verilog/noinline/NoInline_ArgNotInBits.bsv.bsc-vcomp-out.expected).  I'm not totally sure where this is coming from or how to suppress it, but it maybe isn't too bad.  

Congrats on making it to the end of this wall of text.  Hopefully @quark17 has time to look this over before the sync meeting in Friday?


src/Libraries/Base1/Prelude.bs
```diff
@@ -4370,3 +4428,213 @@ data (MetaConsAnon :: $ -> # -> # -> *) name idx nfields = MetaConsAnon
...
+  toWrapField :: StrArg name -> String -> List String -> f -> w
+
+  -- Given a proxy value for the field name, converts a wrapper interface field value
+  -- to its synthesized interface field.
+  fromWrapField :: StrArg name -> w -> f
+
+  -- Save the port types for a field in the wrapped interface, given the module name
+  -- and the prefix, arg_names and result pragmas.
+  saveFieldPortTypes :: StrArg name -> f -> Maybe Name__ -> String -> List String -> String -> Module ()
+  saveFieldPortTypes _ _ _ _ _ _ = return ()
```
nanavati, line 4453: Why does saveFieldPortTypes have default implementation that does nothing?

That means the default behavior is to save nothing when the user doesn't remember to override this method. I think an error saying saveFieldPortTypes is undefined is better.

src/Libraries/Base1/Prelude.bs
```diff
@@ -4370,3 +4428,213 @@ data (MetaConsAnon :: $ -> # -> # -> *) name idx nfields = MetaConsAnon
...
+class WrapMethod m w | m -> w where
+  -- Convert a synthesized interface method to its wrapper interface method.
+  toWrapMethod :: m -> w
+
+  -- Convert a wrapper interface method to its synthesized interface method.
+  fromWrapMethod :: w -> m
+
+  -- Compute the actual argument base names for a method, given the prefix and arg_names pragmas.
+  methodArgBaseNames :: m -> String -> List String -> Integer -> List String
+  methodArgBaseNames _ _ _ _ = Nil
```
nanavati, line 4492: Why do methodArgBaseNames and inputPortNames have default implementations? Again it seems like a default of not implementing them would help users get implementations of WrapMethod right.

src/Libraries/Base1/Prelude.bs
```diff
@@ -4370,3 +4428,213 @@ data (MetaConsAnon :: $ -> # -> # -> *) name idx nfields = MetaConsAnon
...
+  -- Save the port types for a method, given the module name, argument base names and result name.
+  saveMethodPortTypes :: m -> Maybe Name__ -> List String -> String -> Module ()
+
+instance (SplitPorts a p, TupleSize p n, WrapPorts p pb, WrapMethod b v, Curry (pb -> v) w) =>
+    WrapMethod (a -> b) w where
+  toWrapMethod f = curryN $ toWrapMethod ∘ f ∘ unsplitPorts ∘ unpackPorts
+  fromWrapMethod f = fromWrapMethod ∘ uncurryN f ∘ packPorts ∘ splitPorts
+
+  methodArgBaseNames _ prefix (Cons h t) i = Cons
+    (if prefix == "" && not (isDigit $ stringHead h) then h else prefix +++ "_" +++ h)
```
nanavati, line 4507: Where does the isDigit case come up? Is this is someone passes in a arg_names pragma that starts with a digit?

src/Libraries/Base1/SplitPorts.bs
```diff
@@ -0,0 +1,193 @@
...
+  shallowUnsplitPorts' = map shallowUnsplitPorts' ∘ unconcatTuple
+  shallowSplitPortNames' _ base =
+    let genElem i = shallowSplitPortNames' (_ :: r) (base +++ "_" +++ integerToString i)
+    in List.concat $ List.map genElem $ List.upto 0 (valueOf n - 1)
+
+instance (ShallowSplitPorts' r p) => ShallowSplitPorts' (Meta (MetaField name idx) r) p where
+  shallowSplitPorts' (Meta x) = shallowSplitPorts' x
+  shallowUnsplitPorts' = Meta ∘ shallowUnsplitPorts'
+  shallowSplitPortNames' _ base = shallowSplitPortNames' (_ :: r) $
+    if stringHead (stringOf name) == '_'
```
nanavati, line 71: Why is this special case needed?

src/Libraries/Base1/SplitPorts.bs
```diff
@@ -0,0 +1,193 @@
...
+  deepSplitPortNames _ base = Cons base Nil
+
+instance DeepSplitPorts () () where
+  deepSplitPorts _ = ()
+  deepUnsplitPorts _ = ()
+  deepSplitPortNames _ _ = Nil
+
+instance (DeepSplitTuplePorts (a, b) p) => DeepSplitPorts (a, b) p where
+  deepSplitPorts = deepSplitTuplePorts
+  deepUnsplitPorts = deepUndeepSplitTuplePorts
```
nanavati, line 109: What is the name `deepUndeepSplitTuplePorts` supposed to mean?

krame505: Find/replace gone wrong... should be `deepUnsplitTuplePorts`.

src/Libraries/Base1/SplitPorts.bs
krame505: Data field names are _digit.  Added a comment.

src/Libraries/Base1/Prelude.bs
krame505: Yeah

src/Libraries/Base1/Prelude.bs
krame505: It kept things a bit cleaner in the below instances.  Users shouldn't be implementing instances for WrapMethod, typically.  But I suppose there's no good reason why not.

src/comp/CVPrint.hs
```diff
@@ -287,7 +287,7 @@ instance PVPrint CDefn where
     pvPrint d p (CprimType (IdKind i k)) =
         t"primitive type" <+> pp d i <+> t "::" <+> pp d k
 
-    pvPrint d p (Cforeign i ty oname opnames) =
+    pvPrint d p (Cforeign i ty oname opnames _) =
```
nanavati, line 290: It might be nice for the pretty-printing of `Cforeign` to tell you if it is a noinline function or not now that it knows.

src/comp/ContextErrors.hs
```diff
@@ -454,6 +459,13 @@ handleCtxRedPrimPort pos (vp, reduced_ps) userty =
     in
         (pos, ECtxErrPrimPort (pfpString userty) poss hasVar)
 
+-- --------------------
+
+handleCtxRedWrapField:: Position -> (VPred, [VPred]) -> FString -> Type -> EMsg
+handleCtxRedWrapField pos (vp, reduced_ps) name userty =
+    (pos, EBadIfcType (getFString name)
+     "This method uses types that are not in the Bits or SplitPorts typeclass.")
```
nanavati, line 467: typeclasses

src/Libraries/Base1/Prelude.bs
krame505: Good point, although I hadn't really considered users writing instances for this type class.

src/comp/CVPrint.hs
krame505: Done.

src/comp/ContextErrors.hs
krame505: Done.

src/comp/GenWrap.hs
```diff
@@ -913,37 +900,31 @@ genIfcField trec ifcIdIn prefixes (FInf fieldIdQ argtypes rettype _) =
...
-                                                cf_default = []
-                                              }
-                       --
-                       -- the ready field
-                       let rdy_field = if (isClock || isReset || isIot) then []
-                                       else mkReadyField trec ifcPragmas ifcIdIn fieldId fi
-                       --
-                       --traceM( "ifc_fields is: " ++ ppReadable ifc_field)
-                       return ((ifc_field : rdy_field), mprops )
+                      let fnt = cTStr (getIdFString fieldIdQ) (getIdPosition fieldIdQ)
```
nanavati, line 903: Does this mean that when you have a nested interface and you get an error about wrapping you'll get an error about the method name in the subinterface without a reference to what it is contained in?

src/comp/GenWrap.hs
```diff
@@ -1490,7 +1444,7 @@ mkNewModDef _ (def,_,_,_) =
 -- This is the part of "genWrapInfo" which makes the DefFun,
 -- a continuation function which does the final wrapper computation.
 
--- type DefFun = VWireInfo -> VSchedInfo -> VPathInfo -> [VPort] -> SymTab -> [VFieldInfo] -> [Id] -> [Id] -> IO CDefn
+-- type DefFun = Bool -> VWireInfo -> VSchedInfo -> VPathInfo -> [VPort] -> SymTab -> [VFieldInfo] -> [Id] -> IO CDefn
```
nanavati, line 1447: Was the type of DefFun in the comment wrong before?

src/comp/GenWrap.hs
nanavati, line 1: I didn't see the deletion of the code that did the old port-name checking in GenWrap. Did I just miss it or is there a reason you're keeping it around?

src/comp/IExpand.hs
```diff
@@ -2122,6 +2113,24 @@ evalString e = do
             _ -> do e'' <- unheapAll e'
                     errG (getIExprPosition e'', EStringNF (ppString e''))
 
+evalStringList :: HExpr -> G ([String], Position)
+evalStringList e = do
+  e' <- evaleUH e
+  case e' of
+    IAps (ICon _ c) _ [a] -> do
+      a' <- evaleUH a
```
nanavati, line 2121: You should be able to test whether it is cons or nil by looking at the constructor number of the ICCon, rather than hacking around with the containing tuple.

src/comp/IExpand.hs
```diff
@@ -2122,6 +2113,24 @@ evalString e = do
...
+      -- XXX this is a horrible way of pulling apart a list, but I don't think there is a better way:
```
nanavati, line 2122: The IConInfo for the outer application should be an ICCon that tells you which constructor is being applied to the argument. You should be able to test that to figure out whether you're dealing with Cons or Nil.

src/comp/IExpand.hs
```diff
@@ -4091,6 +4105,10 @@ getBuriedPreds (IAps ic@(ICon i_sel (ICSel { })) ts1 [e])
     | (i_sel == idAVValue_ || i_sel == idAVAction_) = do
   --traceM("getBuriedPreds: AV sel")
   getBuriedPreds e
+getBuriedPreds (ICon _ (ICMethod _ _ eb)) = do
```
nanavati, line 4108: This surprises me. Is there a testcase that means we need to peek inside of ICMethod to surface implicit conditions?

src/comp/VModInfo.hs
```diff
@@ -67,7 +67,7 @@ getVNameString (VName string) = string
 
 -- convert Bluespec identifier to Verilog names
 id_to_vName :: Id -> VName
-id_to_vName i = VName (getIdString i)
+id_to_vName i = VName (getIdBaseString i)
```
nanavati, line 70: Why did this need to change?

testsuite/bsc.mcd/Misc/ClockCheckCond.bsv.bsc-vcomp-out.expected
```diff
@@ -6,7 +6,7 @@ Error: "ClockCheckCond.bsv", line 6, column 8: (G0007)
   Method calls by clock domain:
     Clock domain 1:
       default_clock:
-	the_y.read at "ClockCheckCond.bsv", line 2, column 18,
+	the_y.read at "ClockCheckCond.bsv", line 2, column 10,
```
nanavati, line 9: Do you know why this position changed?

testsuite/bsc.names/portRenaming/conflicts/clock/ClockEnable.bsv
```diff
@@ -6,4 +6,6 @@ endinterface
 
 (* synthesize *)
 module mkClockEnable(Ifc);
+    method m = noAction;
```
nanavati, line 9: What was happening before that you needed to edit this?

src/comp/GenWrap.hs
krame505: It was removed, in `PragmaCheck.hs`.

src/comp/GenWrap.hs
krame505: Yeah, it was out of date.

src/comp/GenWrap.hs
krame505: Good point, I tweaked it to include the subfield path from the prefixes.

src/comp/VModInfo.hs
krame505: I think this was previously only getting passed unqualified Ids, but something changed causing qualified names to srtart showing up in the Verilog from `vMkRWire1`.

testsuite/bsc.mcd/Misc/ClockCheckCond.bsv.bsc-vcomp-out.expected
krame505: I'm not sure I understand why this error message had the position that it did in the first place. It is still pointing to the same interface field at least.

src/comp/IExpand.hs
krame505: I think I added this when I was debugging some issues with implcit conditions not being surfaced, but it turns out not to have actually been needed.

testsuite/bsc.names/portRenaming/conflicts/clock/ClockEnable.bsv
krame505: The errors were previously being reported by genwrap, now the errors are reported after elaboration.  Having the fields undefined was causing an elaboration-time error before getting to the port name conflict check.

nanavati:
I think it would be good to have tests for the error messages you get if SplitPorts instances are wrong (wrong number of names, name conflicts between generated ports and any others you can think of).

src/comp/IExpand.hs
krame505: Done.

src/comp/PreIds.hs
```diff
@@ -118,10 +118,11 @@ idInvalid = prelude_id_no fsInvalid
 idValid = prelude_id_no fsValid
 idEmpty = prelude_id_no fsEmptyIfc
 idFile = prelude_id_no fsFile
-idEither, idLeft, idRight :: Id
+idEither, idLeft, idRight, idPreludeCons :: Id
 idEither = prelude_id_no fsEither
 idLeft = prelude_id_no fsLeft
 idRight = prelude_id_no fsRight
+idPreludeCons = prelude_id_no fsCons  -- idCons isn't qualified
```
nanavati, line 125: Why did you need a qualified version of idCons? Are the existing uses of idCons wrong?

krame505: Yeah the existing `idCons` is unqualified for some reason. I am comparing equality with the id coming out of `ICon`, which is qualified.  I suppose I could just compare the base string but comparing with the whole id seems better?

quark17: I agree with Ravi that instead of defining `idPreludeCons`, we should fix `idCons` (and `idNil`) to be qualified.  (I'd also remove the neighboring `idConcat`, since it's unused, and remove `idNothing`, replacing its few uses in the BSV parser with `idInvalid` -- and remove their associated definitions in `PreStrings`.)  That could be done in a separate PR.

src/Libraries/Base1/PreludeBSV.bsv
```diff
@@ -88,15 +88,17 @@ interface VRWireN#(numeric type n);
...
-      return(toPrimAction(_rw.wset(v)));
-   endmethod
-   method wget = _rw.wget;
-   method whas = pack(_rw.whas);
+   function rw_wset(v);
+      return toPrimAction(_rw.wset(v));
+   endfunction
+   method wset = primMethod(Cons("v", Nil), rw_wset);
+   method wget = primMethod(Nil, _rw.wget);
+   method whas = primMethod(Nil, pack(_rw.whas));
```
quark17, line 101: If the user writes a `conflict_free` attribute for two rules, then BSC will add logic to dynamically check that any conflicting methods between the two have not been called at the same time.  BSC does this by instantiating an `RWire` for each method called by each rule, and inserting a `wset` action at the place in the rule where it calls the method; then a $display is inserted into the module, which prints an error message if the wires for two conflicting methods is ever true.

This occurs in BSC after scheduling, when the module is in `ASyntax`.  To add a submodule instantiation at that point, BSC needs an `AVInst` value.  The reason that `vMkRWire` exists is so that BSC can get its hands on the `AVInst` for `RWire`.  It does this by simulating the entire compilation flow for `vMkRWire` -- first typecheck, then convert to ISyntax, then elaboration, then convert to ASyntax -- and then looking for the sole submodule in the resulting ASyntax.

The interface for `vMkRWire` is entirely unneeded -- BSC just wants to get the `AVInst` structure for the `_rw` submodule instantiation.  We might consider changing this module to have an empty interface, and then you don't need to worry about adding `primMethod` to any methods.

The reason that you needed to add `primMethod` is because BSC's simulated compilation flow starts at typecheck, without any `GenWrap` stage.  It used to be OK for BSC to skip `GenWrap`, because later stages only required that the interface be raw types (bit and PrimAction).  But you've changed the evaluator to expect that all methods evaluate to `IMethod`, and that only happens when there's a call to `primMethod`, and that gets inserted by the `toWrapField` call that `GenWrap` now inserts on the interface.  So `AAddSchedAssumps` would need to simulate the `GenWrap` step by inserting `toWrapField`; but if that's not done, then you would need to manually insert the `toWrapField`, or manually write the result of that (by inserting `primMethod` on each method).

I think it might be better to write `toWrapField` on `_rw`, rather than explicitly spell out the interface; and better than that would be to update `AAddSchedAssump` to insert it; but better still would be to eliminate the interface on `vMkRWire1` altogether -- but actually, I would prefer that we eliminate the simulated compile flow altogether and just hardcode an `AVInst` value for RWire in `AAddSchedAssump`.

However, all of this does make me wonder: For imported modules in BH code, we have written them as an import of the raw interface and then a wrapper module (that converts from the raw interface and inserts calls to primitives to save types etc) -- for example, `mkReg` is a wrapper around `vMkReg`.  Can we simplify those wrappers now with just a call to `toWrapField`?

quark17: Also, it looks like someone added a `vMkUnsafeRWire1` version of this module, when they were creating `Unsafe` versions of all the `RWire` modules, but there's no need for that, so I'd delete that module.

src/Libraries/Base1/Prelude.bs
```diff
@@ -253,7 +254,10 @@ package Prelude(
         -- Generics
         Generic(..), Conc(..), ConcPrim(..), ConcPoly(..),
         Meta(..), MetaData(..), StarArg(..), NumArg(..), StrArg(..), ConArg(..),
-        MetaConsNamed(..), MetaConsAnon(..), MetaField(..)
+        MetaConsNamed(..), MetaConsAnon(..), MetaField(..),
+
+        primMethod, WrapField(..), WrapMethod(..), WrapPorts(..),
```
quark17, line 259: I assume that `primMethod` here is only exported so that it can be used in `PreludeBSV` for `vMkRWire1` and can be removed when that's resolved.

src/Libraries/Base1/PreludeBSV.bsv
krame505: To be honest I haven't dug into `AAddSchedAssump` suffieciently to understand what is going on there.  Although constructing an `AVInst` directly does seem a lot nicer than simulating the whole compilation flow, if doing so is reasonable.  

I did try changing the `primMethod` to `toWrapField`, but that doesn't work because `VRWireN.wset` returns a `PrimAction`, but `toWrapField` uses `ActionValue_ 0`.  The implementation of `ActionValue_` is not exported from the Prelude, and even if it was, we can't manipulate the result of `toWrapField` since it is wrapped in `primMethod`.  I suppose we could make `toWrapMethod` use `PrimAction` instead of `ActionValue_ 0` for wrapping  `ActionValue`s?  But this is likely to change again anyway in the future when we add support for methods with multiple output ports.

I suppose it is possible to use `toWrapMethod` in the manually-written wrapper modules for imported Verilog.  Although in those cases the wrapped interface field types need to be written down anyway.  Also this would require fixing the above-mentioned `PrimAction`/`ActionValue_ 0` disparity.

src/Libraries/Base1/Prelude.bs
```diff
@@ -171,6 +171,7 @@ package Prelude(
         Tuple6, tuple6, Has_tpl_6(..),
         Tuple7, tuple7, Has_tpl_7(..),
         Tuple8, tuple8, Has_tpl_8(..),
+        AppendTuple(..), AppendTuple'(..), TupleSize(..),
```
quark17, line 174: Does `AppendTuple'` need to be exported?

src/Libraries/Base1/Prelude.bs
```diff
@@ -2589,6 +2593,23 @@ curry f x y = f (x, y)
 uncurry :: (a -> b -> c) -> ((a, b) -> c)
 uncurry f (x, y) = f x y
 
+-- Polymorphic, N-argument version of curry/uncurry
+class Curry f g | f -> g where
+  curryN :: f -> g
```
quark17, line 2598: I'm trying to understand: Can any current use of `curry`/`uncurry` be replaced by `curryN`/`unCurryN`?  That is -- error message concerns aside -- could the current `curry` and `uncurry` definitions be removed and the typeclass methods be named `curry` and `uncurry`?  (I understand that Haskell's `Data.Tuple.Curry` names them `curryN` and `uncurryN`, so you might be following that.)

src/Libraries/Base1/Prelude.bs
```diff
@@ -3372,6 +3393,43 @@ tuple7 a b c d e f g = (a,b,c,d,e,f,g)
...
+
+instance (AppendTuple' a b c) => AppendTuple' (h, a) b (h, c) where
+  appendTuple' (x, y) z = (x, appendTuple' y z)
+  splitTuple' (x, y) = case splitTuple' y of
+    (w, z) -> ((x, w), z)
+
+class TupleSize a n | a -> n where {}
+instance TupleSize () 0 where {}
+instance TupleSize a 1 where {}
+instance (TupleSize b n) => TupleSize (a, b) (TAdd n 1) where {}
```
quark17, line 3431: Hm, would the size of `(x,())` be reported as 1?  Do we think it should be 2?

src/Libraries/Base1/Prelude.bs
krame505: For instance resolution to work correctly, any type classes referenced by an instance need to be exported.  E.g. we need to export `PrimDeepSeqCond'(..)` in addition to `PrimDeepSeqCond(..)`.

src/Libraries/Base1/Prelude.bs
krame505: I did use the names from Haskell.

It should be possible to replace them, I think.  The downside is that in principle this could create type ambiguities when the curried function type isn't constrained.  There would also be slight breaking change in the behavior if someone wrote `uncurry f` where `f :: a -> b -> c -> d`; the result would now have type `(a, b, c) -> d` instead of `(a, b) -> c -> d`.

src/Libraries/Base1/Prelude.bs
quark17: I will have to create an example to see what the behavior is.  I thought that BSC had access to all instances inside imported files, and that the exporting was only about making the name available to users to write code that mentions the typeclass -- and the `(..)` only needed for code that mentions the method names (thus there's probably no need to have `(..)` on `TupleSize` since it has no methods?).

src/Libraries/Base1/Prelude.bs
krame505: Hmm, I suppose it would be reported as 1.  

I don't think it's too unreasonable to treat tuples that end in `()` as one element smaller.  Note that `AppendTuple` also would drop the `()` if the first tuple being appended ends in `()`, and the same for `CurryN`.

quark17: I recall that there is a general issue with tuples: because they are build from nested pairs, the last element can't be of a tuple type, otherwise it's indistinguishable from one large tuple.  For example (and this works in BSV too), if you have a variable of type `Tuple2 a (Tuple2 b c)`, you can assign to it with `tuple3` -- and the current implementation of `TupleSize` considers it 3.  Here's code that you can try:
```
package TupleSizeBH where

type T3 = Tuple2 Bool (Tuple2 Bool Bool)

mkTupleSizeBH :: (TupleSize T3 3) => Module Empty
mkTupleSizeBH =
  module
    let x :: T3
        x = tuple3 True False True
```
However, if we have a variable of type `Tuple3 a b ()`, that _cannot_ be assigned with `tuple2` -- BSC gives a type error unless you use `tuple3`.  However `TupleSize` says its 2.  This inconsistency seems wrong to me.  

The issue is that the expression`(e1,e2)` parses into `PrimPair e1 e2`, and `(e)` parses the same as `e`, and `()` parses into `PrimUnit`.  Longer tuples are nestings of `PrimPair`, but `PrimUnit` is not a zero of `PrimPair` -- this is perhaps clearer in BSV, where you can't write `()`, it has to be written as `void`.

The inconsistency can be fixed for `TupleSize` by defining it like this:
```
instance TupleSize a 1 where {}
instance TupleSize (a,b) 2 where {}
instance (TupleSize b n) => TupleSize (a, b) (TAdd n 1) where {}
```

But you're saying that `AppendTuple` and `curryN` also use `PrimUnit` as a zero.  Here is an example showing that appending a `Tuple3` and `Tuple2` will give `Tuple4` if the `Tuple3` ends with `PrimUnit`:
```
package AppendTupleBH where

type TA = Tuple3 Bool Bool ()

type TB = Tuple2 Bool Bool

-- type TC = tuple5 Bool Bool () Bool Bool
type TC = Tuple4 Bool Bool Bool Bool

mkAppendTupleBH :: (AppendTuple TA TB TC) => Module Empty
mkAppendTupleBH =
  module
     let x :: TA
         x = tuple3 True False ()
         y :: TB
         y = tuple2 True False
         z :: TC
         z = appendTuple x y
```
This seems to be because you want to support `()` as a zero in calls to `appendTuple` and `splitTuple`, and so there need to be instances for `()`, but you have also written your typeclass so that it reuses the top-level instances for sub-elements of the tuple.  If you separate the processing of the sub-elements from the top-level, I think the issue goes away.  Of course, there's also a complication that BSC won't allow overlapping instance of the form `T () x` and `T x ()`, even if a more explicit instance for the intersection `T () ()` is given -- which is unfortunate, and maybe something that can be fixed one day?  You got around that by adding a second level of typeclass.  However, that second level is still conflating top-level `()` with sub-element `()`.  I think you just need to add a third level of typeclass, that's only for processing once top-level `()` has been eliminated -- and in fact I confirmed that this works:
```
class AppendTuple a b c | a b -> c where
  appendTuple :: a -> b -> c
  splitTuple :: c -> (a, b)

instance AppendTuple a () a where
  appendTuple x _ = x
  splitTuple x = (x, ())

instance (AppendTuple' a b c) => AppendTuple a b c where
  appendTuple = appendTuple'
  splitTuple = splitTuple'

class AppendTuple' a b c | a b -> c where
  appendTuple' :: a -> b -> c
  splitTuple' :: c -> (a, b)

instance AppendTuple' () a a where
  appendTuple' _ x = x
  splitTuple' x = ((), x)

instance (AppendTuple'' a b c) => AppendTuple' a b c where
  appendTuple' = appendTuple''
  splitTuple' = splitTuple''

class AppendTuple'' a b c | a b -> c where
  appendTuple'' :: a -> b -> c
  splitTuple'' :: c -> (a, b)

-- Top-level () has been handled before AppendTuple''
-- so no instance is needed for () because occurrences
-- can only be sub-elements at this point and are handled
-- by the instance for sub-elements of any type

instance AppendTuple'' a b (a, b) where
  appendTuple'' a b = (a, b)
  splitTuple'' = id

instance (AppendTuple'' a b c) => AppendTuple'' (h, a) b (h, c) where
  appendTuple'' (x, y) z = (x, appendTuple'' y z)
  splitTuple'' (x, y) = case splitTuple'' y of
    (w, z) -> ((x, w), z)
```

I haven't looked closely at `curryN`, but I assume that a layer of hierarchy could be added, with `()` not handled in the second layer.

src/Libraries/Base1/Prelude.bs
quark17: I confirmed that BSC requires you to export `AppendTuple'`, however the export of its methods with `(..)` is not required.  I suggest removing `(..)` from `AppendTuple'(..)`.

In theory, we could enhance BSC to not require this explicit export -- and somehow implicitly pull out what is needed -- but I guess, in general, the user may see typecheck error messages that refer to helper typeclasses, and so the user can't be buffered from having to know about the existence of these helpers, so it might as well be an explicit typeclass (though the user doesn't need to know about its methods).  But this brings up a point: for any of these typeclass-based functions that you've added, if it's something that users might use (not just generated code), then we should consider what the error messages will look like (in various misuse examples), and may possibly need to add cases to `src/comp/ContextErrors.hs` to give more user-friendly messages when they fail.  (It'd be interesting to know whether there's a source-level mechanism that could be added for that, rather than having to make special handling in the compiler each time -- assuming it's not something that can be generalized in the compiler.)

And maybe consider removing `(..)` from `TupleSize(..)` -- I'm not sure whether it's cleaner to have it or not have it? (not that it matters much).

src/Libraries/Base1/Prelude.bs
quark17: I was worried that there's still the issue that `splitTuple` of `Tuple3 Bool Bool ()` to split off the last element and then `appendTuple` to put it back won't result in the original tuple.  But actually, both definitions are allowed -- by your implementation and mine!  Unless I'm missing some reason why this example compiles:
```
package TestSplitAppend where

type TA = Tuple3 Bool Bool ()

mkTestSplitAppend :: Module Empty
mkTestSplitAppend =
  module
     let x :: TA
         x = tuple3 True False ()
         p :: ((Bool,Bool), ())
         p = splitTuple x
         y1 :: Tuple2 Bool Bool
         y1 = case p of
                (a,b) -> appendTuple a b
         y2 :: TA
         y2 = case p of
                (a,b) -> appendTuple a b

     let e1 :: (Bool, Bool)
         e1 = (True, False)
         e2 :: ()
         e2 = ()
         e31 :: (Bool, Bool)
         e31 = appendTuple e1 e2
         e32 :: (Bool, Bool, ())
         e32 = appendTuple e1 e2
```
And it's not because the types are considered the same, because adding `y1 == y2` causes a type error.

There's a dependency on the typeclass, so for the same arguments, only one result type should be possible:
```
class AppendTuple a b c | a b -> c where
```

Separately, isn't there something missing in the dependencies you declared?  Shouldn't it also have `a c -> b` and `b c -> a`?

src/comp/ContextErrors.hs
```diff
@@ -454,6 +459,13 @@ handleCtxRedPrimPort pos (vp, reduced_ps) userty =
     in
         (pos, ECtxErrPrimPort (pfpString userty) poss hasVar)
 
+-- --------------------
+
+handleCtxRedWrapField:: Position -> (VPred, [VPred]) -> FString -> Type -> EMsg
+handleCtxRedWrapField pos (vp, reduced_ps) name userty =
+    (pos, EBadIfcType (getFString name)
+     "This method uses types that are not in the Bits or SplitPorts typeclasses.")
```
quark17, line 467: It would be good for the error message to point to the specific types that are a problem.  (Caveat: I haven't yet looked at the testcases yet, to see examples of the error message.)

This `handle*` function is called when `WrapField` didn't reduce; there is probably an instance, but that instance introduces provisos (like `Bits` or `SplitPorts`) which don't reduce.  As some of the other `handle*` functions demonstrate, it's possible to get a list of those leaf provisos, and then you can extract the list of types whose provisos didn't reduce, and can report those in the error message.

src/comp/GenWrapUtils.hs
```diff
@@ -87,4 +88,7 @@ getDefArgs dcls t =
 
 -- ====================
 
+mkProxy :: CType -> CExpr
```
quark17, line 91: This name wasn't clear to me.  I don't think I'd heard of these kind of dummy values as being called proxies before, but that's fine, as long as there's some more info about what it's a proxy for.  Maybe `mkTypeProxy` -- or `mkTypeProxyExpr` if that's not too long.

src/comp/CSyntax.hs
```diff
@@ -964,8 +965,10 @@ instance PPrint CDefn where
             (IdK i)        -> ppConId d i
             (IdKind i k)   -> ppConId d i <+> t "::" <+> pp d k
             (IdPKind i pk) -> ppConId d i <+> t "::" <+> pp d pk
-    pPrint d p (Cforeign i ty oname opnames) =
-        text "foreign" <+> ppVarId d i <+> t "::" <+> pp d ty <> (case oname of Nothing -> text ""; Just s -> text (" = " ++ show s)) <> (case opnames of Nothing -> text ""; Just (is, os) -> t"," <> pparen True (sep (map (text . show) is ++ po os)))
+    pPrint d p (Cforeign i ty oname opnames _) =
```
quark17, line 968: Why is the new argument (`cforg_is_noinline`) not being printed in the output?  I believe that it can be printed as a `noinline` pragma just above the statement.

testsuite/bsc.mcd/Misc/ClockCheckCond.bsv.bsc-vcomp-out.expected
quark17: The message is about the definition for `out` in the module (at line 10, though referring to the use of `y` at line 11).  However, what is happening under the covers is that BSC is creating an interface struct from all the definitions, and the names for the fields of that struct are taken from the interface type declaration, and `out` is declared at line 2.  So that causes an error on `out` in the module to end up pointing to `out` in the type.  It's a known issue that we should fix at some point (Bluespec Inc internal bug database issue 1238).

However, it's interesting that your PR has changed the position from the name (`out`) to the type (`Bit#(8)`).  It could be due to how you're picking positions for any ISyntax/CSyntax constructs that you're building, or maybe how you've changed the expression structures (causing `getPosition` to find a different position first).  I'll try to keep an eye out for it while looking through your changes.  (I do think it's worth understanding why positions change -- regardless of what we think of the specific test example, because it could show up as a position change in other situations that we're not testing.)

src/comp/ContextErrors.hs
quark17: Ah, an example of the error is `testsuite/bsc.verilog/noinline/NoInline_ArgNotInBits.bsv`.  Before there was one message that mentioned the type not in `Bits`; now there's the above `EBadIfcType` and a separate message for the `Bits` class.  I'd rather that `handleCtxRedWrapField` filter out the `Bits` failures and report the same single error message.

testsuite/bsc.verilog/noinline/noinline.exp
```diff
@@ -44,8 +44,7 @@ test_c_veri_bsv_modules \
 # The typedef fails because BSC doesn't expand the synonym before checking
```
quark17, line 44: It looks like this comment is no longer correct, and should be removed.  (It appears that the bug has been fixed, because the checked has moved out of GenWrap.)

src/Libraries/Base1/Prelude.bs
krame505: Correct.

src/Libraries/Base1/Prelude.bs
krame505: I'm just getting back to your comments here - I'm not sure I agree your version of `TupleSize` is what we want? 
 The version you specified would have `TupleSize () 1`, but we want `()` to have size `0` for the purposes of `checkPortNames`.  I suppose we could introduce `TupleSize'` and make `(a, ())` have size 2 but `()` have size 0 - would that address your concern?

testsuite/bsc.bugs/bluespec_inc/b1894/b1894.exp
```diff
@@ -18,8 +18,8 @@ if { $ctest == 1 } {
     # backend, and only then if the user has specified that it's OK
     # for the Verilog and Bluesim backends to diverge).
     #
-    find_regexp mkTop.cxx {2047u \& \(\(\(\(\(tUInt32\)\(\(tUInt8\)0u\)\) << 3u\) \| \(\(\(tUInt32\)\(DEF_cond__h[0-9]+\)\) << 2u\)\) \| \(tUInt32\)\(DEF_v__h172\)\);}
-    find_regexp mkTop.cxx {DEF_v__h172 = DEF_AVMeth_s_m;}
+    find_regexp mkTop.cxx {2047u \& \(\(\(\(\(tUInt32\)\(\(tUInt8\)0u\)\) << 3u\) \| \(\(\(tUInt32\)\(DEF_cond__h[0-9]+\)\) << 2u\)\) \| \(tUInt32\)\(DEF_v__h\d+\)\);}
```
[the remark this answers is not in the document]
krame505: This just seemed like a brittle test to be checking for a specific generated h number.  But I can change it to the new number if that is fine for now.

src/Libraries/Base1/Prelude.bs
krame505: Correct.

testsuite/bsc.codegen/signature/NestedIfcIntegerArg.bs
```diff
@@ -0,0 +1,14 @@
+package NestedIfcIntegerArg where
```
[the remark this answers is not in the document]
krame505: I think I just wanted an example of it being possible to include a non-synthesizable type as a method parameter in a synthesized interface, without the error that would have been previously reported.

src/Libraries/Base1/PreludeBSV.bsv
krame505: Can we consider this resolved for now?  This can be revisited when I rework output port generation.

src/comp/CSyntax.hs
krame505: Done.

src/comp/PreIds.hs
krame505: Done.

src/Libraries/Base1/SplitPorts.bs
```diff
@@ -0,0 +1,195 @@
...
+  shallowSplitPortNames' _ = portNames (_ :: a)
+
+
+-- Helper class using generics, to recursively split structs and vectors into a tuple of ports.
+class DeepSplitPorts a p | a -> p where
+  deepSplitPorts :: a -> p
+  deepUnsplitPorts :: p -> a
+  deepSplitPortNames :: a -> String -> List String
+
+instance DeepSplitPorts (UInt n) (Port (UInt n)) where
```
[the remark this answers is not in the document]
krame505: Because `Int` and `UInt` are basically newtype wrappers around `Bit`, and we don't want to recurse into them.  For example if a method `thing` returns a struct that gets split using `DeepSplitPorts`, we want field `x :: UInt 8` to yield a port named `thing_x` and not `thing_x_1`.  

I'll add a comment.

src/comp/CSyntax.hs
```diff
@@ -1120,8 +1121,11 @@ instance PPrint CDefn where
             (IdK i)        -> ppConId d i
             (IdKind i k)   -> ppConId d i <+> t "::" <+> pp d k
             (IdPKind i pk) -> ppConId d i <+> t "::" <+> pp d pk
-    pPrint d p (Cforeign i ty oname opnames) =
-        text "foreign" <+> ppVarId d i <+> t "::" <+> pp d ty <> (case oname of Nothing -> text ""; Just s -> text (" = " ++ show s)) <> (case opnames of Nothing -> text ""; Just (is, os) -> t"," <> pparen True (sep (map (text . show) is ++ po os)))
+    pPrint d p (Cforeign i ty oname opnames ni) =
+        (if ni then text "{-# noinline" <+> ppVarId d i <+> text "#-}" else text "") <+>
```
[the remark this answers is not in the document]
krame505: There may or may not be `WrapPorts` in the provisos, depending on what stage is being dumped (pre or post-ctxreduce.)  I'll change it to match `pvPrint`.

testsuite/bsc.verilog/noinline/noinline.exp
```diff
@@ -60,11 +56,11 @@ test_c_veri_bsv_modules NoInlineInSched {module_inv}
 
 if { $vtest == 1 } {
 
-compile_verilog_fail_error NoInline_ArgNotInBits.bsv T0043
+compile_verilog_fail_error NoInline_ArgNotInBits.bsv T0031
```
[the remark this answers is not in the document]
krame505: Done.

testsuite/bsc.verilog/splitports/splitports.exp
```diff
@@ -0,0 +1,61 @@
+
```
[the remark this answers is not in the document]
krame505: I agree it doesn't really belong here - especially in the follow-on PR, there are also tests for Bluesim handling of multiple output ports added here.  Let's maybe revisit in the following PR though, as moving it now would cause more annoying merge conflicts.

src/Libraries/Base1/Prelude.bs
krame505: To be specific, adding the extra fundeps on `AppendTuple` and its helpers gives
```
Error: "Prelude.bs", line 3449, column 9: (T0128)
  Overlapping typeclass instances Prelude.AppendTuple'' a b (a, b) and
  Prelude.AppendTuple'' (h, a) b (h, c) (defined at "Prelude.bs", line 3453,
  column 34 ) cannot be ordered from most-specific to least-specific, so they
  cannot support predictable instance-resolution. Please resolve this by
  changing one or both of the instances.
```

krame505:
I added responses to @quark17's comments, and add pushed some fixes.  Replying to your other items:

> * We'll eventually need documentation, but that can come with the output changes.  However, for library elements that are added here, we could document them here -- e.g. does Vector's ConcatTuple need to be in the library docs for Vector?
> * Should we add testcases for new things in the libraries?  Like SplitTuple/AppendTuple and their corner cases that were discussed in [the comments](https://github.com/B-Lang-org/bsc/pull/729#discussion_r1870330368).  But also for Vector's ConcatTuple.

Yes, there should be.  I'll try to get around to it some time this week.

> * Do we need BSV testcases for port splitting?

I can't think of anything special about BSV that would be worth testing?

> * Fix the position change in an error message in `bsc.mcd/Misc/ClockCheckCond.bsv`

See my above reply.

> * I'd like to investigate if the interface on `vMkRWire` can be removed (eliminating all the changes needed there) and then the exporting of `primMethod` won't be needed.
> * I'd like to understand why `id_to_vName` in `VModInfo.hs` needs to be changed to remove the qualifier -- you said it may have had to do with `vMkRWire`, so maybe that change goes away if we remove that.

I barely remember what was going on with `vMkRWire` now, other than that it seemed very hard to fix nicely.
What do you mean by "the interface on vMkRWire"?  If that module exists, it needs to have an interface?  
I do see that the numeric parameter on `VRWireN` isn't needed, since it is only ever 1, is that what you meant?  

> * I'd like to understand the failure when adding all the dependency directions to `AppendTuple`.  You said it led to ambiguity errors, which is surprising -- maybe they were caused by the use of `TAdd` in a base type (that I suggest changing to `TSub` in the context)

No, `AppendTuple` doesn't use `TupleSize` at all.  I replied above with the specific error message.

> * You mentioned a "PrimAction/ActionValue_ 0 disparity" somewhere -- what was that about, again?

I think that was in relation to `vMkRWire`, and why we can't just use `toWrapMethod` instead of calling `primMethod` directly - the `VRWireN` interface use `PrimAction` while `toWrapMethod` creates an `ActionValue_ 0`.

> * GenWrap could use more comments -- it was hard to review the changes there, because I couldn't immediately understand what was going on (also because the names weren't helpful).  I know this situation wasn't introduced by you, but in code that you're changing, if you have knowledge about what it's doing, it'd be good to record that in comments, if possible -- even just a few one-line comments sprinkled in long blocks of defs.  This isn't blocking, just nice to have.

I added a few more comments.

> * Better reporting of EBadIfcType in ContextErrors (see `testsuite/bsc.verilog/noinline/NoInline_ArgNotInBits.bsv` and [comments](https://github.com/B-Lang-org/bsc/pull/729/changes#r2678916368))

I'll try to get to this soon, running out of steam for today.

> * Positions of/in G0055 nameclash errors (see [comments](https://github.com/B-Lang-org/bsc/pull/729/changes#r2678967384))

Replied above - I don't really know how to obtain the desired positions.

> * Attribute `arg_names` list being too short (should that be an error? warning?) and too long (we should have a test case at least)

See my reply above - I added a check for the too-long case.  For the too-short case, it wasn't obvious what the behavior should be (after all, an empty list is the same as not specifying `arg_names`.)

src/comp/ContextErrors.hs
krame505: I updated the error to include the types lacking a `Bits` or `SplitPorts` instance as you suggested. Conveniently, there is a `Bits` pred for each of these types in the `reduced_ps`, so I don't need to use `reducePred`.  

I changed `EBadIfcType` to take a `Maybe String` and updated the error message as you suggested. It did not seem straightforward to make `handleContextReduction` take an `Id`, since it gets called in multiple places, some of which do not have a top-level id accessable.

krame505: Good point, I will add a `Maybe Id` parameter to `handleContextReduction`.

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -3540,6 +3540,114 @@ \subsubsection{Tuples}
...
+      appendTuple(t1, t2);
+
+   // Splitting a 4-tuple into a 2-tuple and a 2-tuple
+   Tuple4#(Bool, Int#(8), String, Bit#(4)) t4 =
+      tuple4(True, 42, "test", 4'hA);
+
+   Tuple2#(Tuple2#(Bool, Int#(8)), Tuple2#(String, Bit#(4))) parts =
+      splitTuple(t4);
+
+   // Appending with unit type
```
quark17: Did you mean "void" instead of "unit"?

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -3540,6 +3540,114 @@ \subsubsection{Tuples}
...
+\index{AppendTuple@\te{AppendTuple} (type class)}
+\index{appendTuple@\texttt{appendTuple} (\texttt{AppendTuple} type class function)}
+\index{splitTuple@\texttt{splitTuple} (\texttt{AppendTuple} type class function)}
+\index[function]{Prelude!appendTuple}
+\index[function]{Prelude!splitTuple}
+\index[typeclass]{AppendTuple}
+
+\begin{libverbatim}
+   typeclass AppendTuple #(type a, type b, type c)
+      dependencies (a b determines c);
```
quark17: The BSV syntax is `((a, b) determines c)`

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -3540,6 +3540,114 @@ \subsubsection{Tuples}
    foo = tuple2( !field1, field2 );
 \end{verbatim}
```
quark17: Is the new `Curry` typeclass worth documenting somewhere?

src/comp/IExpand.hs
```diff
@@ -2128,6 +2119,25 @@ evalString e = do
...
+        case a' of
+          IAps (ICon _ (ICTuple {})) _ [e_h, e_t] -> do
+            (h, _) <- evalString e_h
+            (t, _) <- evalStringList e_t
+            return (h:t, getIExprPosition e')
+          _ -> internalError ("evalStringList Cons: " ++ showTypeless a')
+      -- We get primChr for Nil, since it's a no-argument constructor
+      else if i == idPrimChr then return ([], getIExprPosition e')
+      else internalError ("evalStringList con: " ++ show i)
+    _ -> do e'' <- unheapAll e'
```
quark17: Instead of these two lines, just call `nfError`.

src/comp/TCheck.hs
```diff
@@ -2639,7 +2639,7 @@ tiExpl''' as0 i sc alts me (oqt@(oqs :=> ot), vts) = do
      -- Were any contexts without variables left unsatisfied?
      if not (null uds) then
         -- Report reduction errors
-        handleContextReduction (getPosition i) uds
+        handleContextReduction (Just i) (getPosition i) uds
```
quark17: I'm no longer sure that passing `(Just i)` is correct here.  This is for typechecking of `let i :: t = e` within a top-level def, and therefore it'd be incorrect for the message to say "could not synthesize i".  I'd be curious if we have an example of context reduction failing for WrapPorts in this situation, to know if reporting `i` is worthwhile.  Otherwise, maybe we put `Nothing` here.

testsuite/bsc.bluetcl/commands/bpackage.tcl.bluetcl-bh-out.expected
```diff
@@ -55,9 +55,9 @@ PreludeBSV vMkCRegA5
 PreludeBSV mkCReg
 PreludeBSV mkCRegU
 PreludeBSV mkCRegA
-PreludeBSV _PreludeBSV.CReg5713
-PreludeBSV _PreludeBSV.CReg5809
-PreludeBSV _PreludeBSV.CReg5904
+PreludeBSV _PreludeBSV.CReg5693
```
quark17: I'm contemplating updating these test to filter out the numbers after CReg into ####, so that we don't have to keep changing this test when PreludeBSV changes.

testsuite/bsc.lib/CReg/TestCReg_TooBig.bsv.bsc-vcomp-out.expected
```diff
@@ -1,9 +1,9 @@
 checking package dependencies
 compiling TestCReg_TooBig.bsv
 code generation for sysTestCReg_TooBig starts
-Error: "PreludeBSV.bsv", line 1002, column 37: (S0015)
+Error: "PreludeBSV.bsv", line 982, column 37: (S0015)
```
quark17: Unlike the other places where I think we can filter out numbers to avoid changing the test every time PreludeBSV changes, I don't think we can do that here, because we're testing the error message and we'll want to know if the position has changed to something else in the file.  (Maybe the infrastructure could search in PreludeBSV for the code it wants the position to be, extract the position, and use that to generate the expected file -- but that's not trivial to add.)

testsuite/bsc.names/signal_names/Method.bsv.bsc-vcomp-out.expected
```diff
@@ -17,7 +17,7 @@ arg info  [clockarg default_clock;, resetarg default_reset;]
 -- APackage resets
 [(0, { wire:  RST_N })]
 -- AP state elements
-rg :: ABSTRACT:  PreludeBSV._PreludeBSV.VRWire110 = RWire
+rg :: ABSTRACT:  PreludeBSV._PreludeBSV.VRWire103 = RWire
```
quark17: I'm contemplating updating these test to filter out the numbers after VRWire into ###, so that we don't have to keep changing this test when PreludeBSV changes.

testsuite/bsc.verilog/splitports/TooManyArgNames.bs.bsc-vcomp-out.expected
```diff
@@ -0,0 +1,10 @@
+checking package dependencies
+compiling TooManyArgNames.bs
+code generation for mkTooManyArgNamesSplitTest starts
+Error: "TooManyArgNames.bs", line 27, column 29: (S0015)
+  Bluespec evaluation-time error: 2 excess arg_names provided for method
```
quark17: BSV users won't know what "arg_names" is (since the names are just part of the code and not an attribute).  Consider replacing that with text which makes sense for either BH or BSV (unless there's some way to test for BH vs BSV and conditionally report different messages).

testsuite/bsc.lib/Prelude/AppendTuple.bs
```diff
@@ -0,0 +1,134 @@
...
+
+      "test3": when counter == 3 ==> action
+        -- Test append two 2-tuples to create 4-tuple
+        let t1 :: (Bool, Int 8)
+            t1 = (True, 5)
+        let t2 :: (UInt 4, Bit 3)
+            t2 = (12, 0b101)
+        let t4 :: (Bool, Int 8, UInt 4, Bit 3)
+            t4 = appendTuple t1 t2
+        $display "appendTuple((True, 5), (12, 0b101)) = (%b, %0d, %0d, %b)" (tpl_1 t4) (tpl_2 t4) (tpl_3 t4) (tpl_4 t4)
```
quark17: I wonder if FShow or CShow (applied to the entire tuple `t4`) would have been preferable to separate display arguments of `(tpl_1 t)`, `(tpl_2 t)`, etc, and it would have printed enums like `False` as its string instead of as `1` like we see in the expected output.  (This also applies for all the other displays.  And to the ConcatTuple test.  Not necessary, though, just mentioning.)

quark17 (commented):
I've reviewed all your recent additions, now.  Almost there! Thanks!

There are a couple previous issues remaining, but they can probably stay unresolved in this PR, or maybe I'll have a closer look when I can.  I'll post an updated list of those issues (sometime later).

doc/libraries_ref_guide/LibDoc/Prelude.tex
krame505: Ah yes, forgot about that one!

src/comp/TCheck.hs
krame505: Hmm, good catch.
We should only ever have `WrapPorts` contexts on the top-level wrapper interface definitions, so I think this situation can never arise, though.

testsuite/bsc.verilog/noinline/NoInline_ResNotInBits.bsv.bsc-vcomp-out.expected
```diff
@@ -1,11 +1,18 @@
...
-  `fnNoInline_ResNotInBits' uses type `NoInline_ResNotInBits::L' which is not
-  in the Bits class.
-  During elaboration of the interface method `fnNoInline_ResNotInBits' at
-  "NoInline_ResNotInBits.bsv", line 4, column 12.
-  During elaboration of `module_fnNoInline_ResNotInBits' at
-  "NoInline_ResNotInBits.bsv", line 4, column 12.
+  Cannot synthesize `module_fnNoInline_ResNotInBits­': The interface method
+  `fnNoInline_ResNotInBits' uses type(s) that are not in the Bits or
+  SplitPorts typeclasses: NoInline_ResNotInBits::L
+Error: "NoInline_ResNotInBits.bsv", line 4, column 12: (T0029)
```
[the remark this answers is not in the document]
krame505: We discussed this previously - I'm not entirely sure where it is coming from, but it gets fixed by the output port splitting refactor

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -6319,12 +6311,59 @@ \subsubsection{Operations on Functions}
...
+
+ // Using uncurryN
+   function Int#(32) add3(Int#(32) a, Int#(32) b, Int#(32) c);
+      return a + b + c;
+   endfunction
+
+   let add3Uncurried = uncurryN(add3);
+   Int#(32) result2 = add3Uncurried(tuple3(1, 2, 3));  // result2 = 6
  
    return(pack(map(compose(head0,toList),state)));
```
quark17: Apologies for nitpicking: These last two examples now look like they're part of the uncurryN example, because of the way that comments are un-indented above each example group.  Can you put them pack at the top (after the first example and before the new examples) and add un-indented comment(s) above them?

In fact, remove the one with `xs` since that just seems to be a duplicate of the first example (and `genList` doesn't exist).  For the `return` example, maybe just a comment saying "Using compose when mapping over a list".

testsuite/bsc.verilog/noinline/NoInline_ResNotInBits.bsv.bsc-vcomp-out.expected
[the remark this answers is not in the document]
quark17: In those cases that call `fromWrapField`, they're picking up the Bits proviso, but they should also be picking up the WrapField proviso (because of the call to `fromWrapField`).  At the point where context reduction fails, if we had both of those, we could report them as one error (just the WrapField).  But it seems like something has maybe removed the WrapField proviso, perhaps because it reduces to the other provisos and therefore has been deemed redundant.  If we can stop that removal from happening, then we could report the T0043 error in all three places (which hopefully are identical and get reported just once).

krame505:
> I think it could be ready! There are some things that I could submit as separate PRs first, and then rebase, unless that would be disruptive:
> 
> 1. I can submit the change to `vMkRWire1` as a separate PR, so it's preserved in the commit history.
> 2. I can submit a PR to make tests not sensitive to changes in PreludeBSV position (`bsc.names/signal_names/` and `bsc.bluetcl/commands/`).

If you would like to make those changes as separate PRs and rebase, that would be fine.  

> Otherwise, the remaining issues can be dealt with in the next PR or later (and by someone else):

Thank you.  I don't know if/when I will have time to look at the position changes.

> 8. The new test in `bsc.codegen/vector_interfaces/` used `-dtypecheck` to dump the wrapper in the second time the `typecheck` stage occurs.  When PR [Some miscellaneous improvements to dump flags #859](https://github.com/B-Lang-org/bsc/pull/859) is merged (improving how dumping works), the flag here may change.  (I'm not sure which will get merged first, as 859 is being re-thought.)

No preference what order those are merged, it is easy enough to change.

> Anyway, I assume we'll hold off on merging this until the follow-on PR is close to ready to merge? I'm OK with merging this ahead of time, if that's easier on you, but if so, let's do it after the pending release.
> 
> Is the follow-on PR still in progress? Should I start reviewing it?

The follow-on PR is functionally complete from my end, @nanavati hasn't had a chance to review it yet though, and I'm not sure when he will.  

If this PR is merged before reviewing the output port splitting changes, then that review can just happen here in #849.  Otherwise you will need to review it in https://github.com/krame505/bsc/pull/1, which is based on the `genwrap` branch in my fork.

PR #748, opened 2024-11-06
Allow improveIf of con/undet when only one constructor

### LINKED ISSUE #742: Evaluator regression
There appears to have been a regression in October 2020, in revision bad371b3, where @nanavati changed `improveIf` to not optimize constructors with undefined.  On the `b-lang-discuss` mailing list, an example was attached to message 712 ([here](https://groups.io/g/b-lang-discuss/message/712)), which I have boiled down to a small example:
```
import Vector::*;

`define Q_SIZE 8
typedef Bit#(2) T;

(*synthesize*)
module mkSample_code();

  Vector#(`Q_SIZE, Reg#(T))           vec_data  <- replicateM(mkRegU);
  Reg #(Maybe#(Vector#(`Q_SIZE, T)))  rg_1   <- mkRegU;
  Reg #(Maybe#(Vector#(`Q_SIZE, T)))  rg_2   <- mkRegU;
  Reg #(Vector#(`Q_SIZE, Bool))       rg_3   <- mkRegU;
  Reg #(Maybe#(Vector#(`Q_SIZE, T)))  rg_4   <- mkRegU;
  Reg #(Vector#(`Q_SIZE, Bool))       rg_5   <- mkRegU;

  rule r;
    Vector#(`Q_SIZE, T) tmp_data = readVReg (vec_data);

    if (rg_1 matches tagged Valid .d1)
      tmp_data  = d1;

    if (rg_2 matches tagged Valid .d2)
      begin
        for (int i = 0 ; i< `Q_SIZE ;i = i+1)
          begin
            if (rg_3[i] == True)
              tmp_data[i] = d2[i];
            end
      end

    if (rg_4 matches tagged Valid .d4)
      begin
        for (int i = 0 ; i< `Q_SIZE ;i = i+1)
          begin
            if (rg_5[i] == True)
              tmp_data[i] = d4[i];
          end
      end

    writeVReg(vec_data,tmp_data);
  endrule

endmodule
```

With older BSC (as far back as 2018 at least), it compiles in under a second; with newer BSC (at commit bad371b3 through to the latest release), it fails due to exceeding the max number unfolding steps.

The code is assigning a temporary variable with the values from a vector of registers, then dynamically adjusting elements of the variable's value, and then writing the variable's final values back to the registers.  The adjustments are the following:
1. If `rg_1` is valid, then replace the temp with that valid value
2. If `rg_2` is valid, then update elements from that valid value, when the corresponding value in `rg_3` is true (thus using `rg_3` as enable signals for the the data in `rg_2`)
3. The same as step 2, but using `rg_4` and `rg_5` as the data and enable signals

This should simplify.  I have not investigated why the commit broke things or what alternate optimization might help.  Ravi, is that something you could look at?

What is interesting is that removing step 1 causes the code to compile again.  Steps 2 and 3 are for-loops with vector indexing and updates; but step 1 is just a single if-statement (and no vector indexing).

If the change to `improveIf` is not vital to how BSC evaluates, perhaps a command-line flag can be added, to select the old behavior when people need it?

A workaround is to re-write the code to manually push the logic into the elements of the vector:
```
  rule r;
    Vector#(`Q_SIZE, T) tmp_data = readVReg (vec_data);

    for (int i = 0 ; i< `Q_SIZE ;i = i+1)
      begin
        if (rg_1 matches tagged Valid .d1)
          tmp_data[i]  = d1[i];

        if (rg_2 matches tagged Valid .d2)
          if (rg_3[i] == True)
            tmp_data[i] = d2[i];

        if (rg_4 matches tagged Valid .d4)
          if (rg_5[i] == True)
            tmp_data[i] = d4[i];
      end

    writeVReg(vec_data,tmp_data);
  endrule
```

quark17:
This fixes #742 by allowing `improveIf` to merge `ICCon` and `ICUndet` when the type has only one constructor, by pushing the conditional into the constructor arguments.

### DIFF
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -4398,6 +4398,42 @@ improveIf f t cnd thn@(IAps chr@(ICon _ (ICPrim _ PrimChr)) ts1 [chr_thn])
   let chrArgType = iGetType chr_thn
   (e', _) <- improveIf f chrArgType cnd chr_thn chr_els
   return (IAps chr ts1 [e'], True)
+
+-- Push if improvement inside constructors when one arm is undefined
+-- and the type has only one constructor
+--
+-- Further down, a general improveIf rule optimizes 'if c e _' to just 'e'.
+-- But that can cause poor code generation for if-else chains returning
+-- the constructors of a union type, so an earlier improveIf rule catches
+-- that situation (before the general rule can apply).
+-- However, if there is only one constructor, we do want an optimization to apply,
+-- so we put that here, prior to the blocking rule.
+--
+improveIf f t cnd thn@(IAps (ICon i1 c1@(ICCon {})) ts1 es1)
+                  els@(ICon i2 (ICUndet { iuKind = u }))
+  | numCon (conTagInfo c1) == 1
+  = do
+      when doTraceIf $ traceM ("improveIf ICCon/ICUndet triggered" ++ ppReadable (cnd,thn,els))
+      let realConType = itInst (iConType c1) ts1
+          (argTypes, _) = itGetArrows realConType
+      when (length argTypes /= length es1) $ internalError ("improveIf Con/Undet:" ++ ppReadable (argTypes, es1))
+      let mkUndet t = icUndetAt (getIdPosition i2) t u
+      (es', bs) <- mapAndUnzipM (\(t, e1) -> improveIf f t cnd e1 (mkUndet t)) (zip argTypes es1)
+      -- unambiguous improvement because the ICCon has propagated out
+      return ((IAps (ICon i1 c1) ts1 es'), True)
+improveIf f t cnd thn@(ICon i1 (ICUndet { iuKind = u }))
+                  els@(IAps (ICon i2 c2@(ICCon {})) ts2 es2)
+  | numCon (conTagInfo c2) == 1
+  = do
+      when doTraceIf $ traceM ("improveIf ICCon/ICUndet triggered" ++ ppReadable (cnd,thn,els))
+      let realConType = itInst (iConType c2) ts2
+          (argTypes, _) = itGetArrows realConType
+      when (length argTypes /= length es2) $ internalError ("improveIf Con/Undet:" ++ ppReadable (argTypes, es2))
+      let mkUndet t = icUndetAt (getIdPosition i1) t u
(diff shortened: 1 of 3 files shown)


src/comp/IExpand.hs
```diff
@@ -4398,6 +4398,42 @@ improveIf f t cnd thn@(IAps chr@(ICon _ (ICPrim _ PrimChr)) ts1 [chr_thn])
...
+                  els@(ICon i2 (ICUndet { iuKind = u }))
+  | numCon (conTagInfo c1) == 1
+  = do
+      when doTraceIf $ traceM ("improveIf ICCon/ICUndet triggered" ++ ppReadable (cnd,thn,els))
+      let realConType = itInst (iConType c1) ts1
+          (argTypes, _) = itGetArrows realConType
+      when (length argTypes /= length es1) $ internalError ("improveIf Con/Undet:" ++ ppReadable (argTypes, es1))
+      let mkUndet t = icUndetAt (getIdPosition i2) t u
+      (es', bs) <- mapAndUnzipM (\(t, e1) -> improveIf f t cnd e1 (mkUndet t)) (zip argTypes es1)
+      -- unambiguous improvement because the ICCon has propagated out
```
mieszko, line 4422: great comment, thank you for that

testsuite/bsc.evaluator/opt/ImproveIf_ConUndet_OneCon.bsv
```diff
@@ -0,0 +1,45 @@
+// This example comes from GitHub Issue #742
```
mieszko, line 1: Maybe just add that it's testing the effect of this optimization on compilation time rather than directly?

mieszko (approved):

testsuite/bsc.evaluator/opt/ImproveIf_ConUndet_OneCon.bsv
quark17: I added more comments, about what was being tested.

I also changed the test to run BSC with specific stack size and max unfolding steps.  I wasn't thinking of this as a compilation time issue -- so that part of your comment confused me -- it seemed that BSC wasn't able to compile the user's original example, even if the max steps was increased to be very large, so I was just testing the exit code of BSC (which would fail if the max unfolding steps or max stack size was reached).  However, I guess it's possible that BSC could succeed if given enough time and memory, so I have changed the test to run BSC with specific limits for the steps and stack (the current defaults), regardless of whether the BSC defaults are ever increased or whether someone runs the testsuite with higher defaults.

PR #758, opened 2025-01-17
Fix a typo in the grammar rule for shorthands in ActionValue method definitions

chiahsuantw:
Fix a typo in the grammar rule for shorthands in ActionValue method definitions (5.5.1): the semicolon after `implicitCond` is required, not optional.

image


quark17:
Thank you!

PR #759, opened 2025-01-25
Fix libraries documentation for num and str type operators

quark17:
Fixes typos in the index entries, correct and standardizes headings, correctly identifies type functions from pseudo-functions, and separately documents the `SizeOf` type function from the `sizeOf` pseudo-function.  Also expands the dependencies in the Makefile so that changes to import files and figures cause rebuild of the PDF.


krame505 (approved):
Looks good to me.

PR #765, opened 2025-03-11
Add Emacs functionality for actionvalue/endactionvalue

sheganinans:


util/emacs/bsv-mode/bsv-mode-22.el
```diff
@@ -958,13 +961,13 @@ Called by `compilation-mode-hook'.  This allows \\[next-error] to find the error
 (defconst bsv-named-block-re  "begin[ \t]*:")
 (defconst bsv-beg-block-re
   ;; "begin" "case" "casex" "fork" "casez" "table" "specify" "function" "task"
-  "\\(\\<\\(begin\\>\\|case\\(\\>\\|x\\>\\|z\\>\\)\\|generate\\|interface\\|method\\|rule\\|action\\|f\\(ork\\>\\|unction\\>\\)\\|specify\\>\\|ta\\(ble\\>\\|sk\\>\\)\\)\\)")
+  "\\(\\<\\(begin\\>\\|case\\(\\>\\|x\\>\\|z\\>\\)\\|generate\\|interface\\|method\\|rule\\|action\\|actionvalue\\|f\\(ork\\>\\|unction\\>\\)\\|specify\\>\\|ta\\(ble\\>\\|sk\\>\\)\\)\\)")
```
quark17, line 964: I notice that `fork` and `function` are written to share the initial `f`, and that `table` and `task` are written to share `ta`.  Should `action` and `actionvalue` share the `action` prefix?

However, I notice that this sharing isn't done in the next line of code.  I don't know about emacs programming to know when it should or shouldn't be used?

util/emacs/bsv-mode/bsv-mode-22.el
```diff
@@ -1464,10 +1468,13 @@ Use filename, if current buffer being edited shorten to just buffer name."
        ((match-end 10) ; endaction
 	;; Search forward for matching action
 	(setq reg "\\(\\<action\\>\\)\\|\\(\\<endaction\\>\\)" ))
-       ((match-end 11) ; endspecify
+       ((match-end 11) ; endactionvalue
+	;; Search forward for matching action
```
quark17, line 1472: I assume you should change `action` to `actionvalue` in this comment copied from above

util/emacs/bsv-mode/bsv-mode-22.el
```diff
@@ -1464,10 +1468,13 @@ Use filename, if current buffer being edited shorten to just buffer name."
...
+	(setq reg "\\(\\<actionvalue\\>\\)\\|\\(\\<endactionvalue\\>\\)" ))
+       ((match-end 12) ; endspecify
 	;; Search forward for matching task
 	(setq reg "\\(\\<task\\>\\)\\|\\(\\<endtask\\>\\)" ))
-       ((match-end 12) ; endgenerate
+       ((match-end 13) ; endgenerate
```
quark17, line 1477: Again, I don't understand how to code in this, so I don't know the meaning of the numbers that follow `match-end` throughout this file.  Do the numbers here have to correspond with numbers anywhere else?  I just want to make sure that there's not a `12` somewhere else that also needs to change to `13`.  (It seems weird that the number just needs to be different from the others in this block, but the specific value is unused?)

util/emacs/bsv-mode/bsv-mode-22.el
```diff
@@ -3063,10 +3071,13 @@ type.  Return a list of two elements: (INDENT-TYPE INDENT-LEVEL)."
 				  ((match-end 12) ; endaction
 				   ;; Search back for matching action
 				   (setq reg "\\(\\<action\\>\\)\\|\\(\\<endaction\\>\\)" ))
-				  ((match-end 13) ; endtask
+				  ((match-end 13) ; endactionvalue
+				   ;; Search back for matching action
```
quark17, line 3075: Another place where `action` should be changed to `actionvalue` in the comment

quark17 (commented):
Awesome, thank you!  This looks straightforward and should be easy to accept, but I don't know enough about coding in emacs, so I'll have to ask someone to look at it or manually try it out and see if it works.  I have commented with a few nits and questions below.

util/emacs/bsv-mode/bsv-mode-22.el
sheganinans: so looking at the docs here: https://www.gnu.org/software/emacs/manual/html_node/elisp/Simple-Match-Data.html

so the relevant part of the docs for `(match-end <count>)` is:

`[...] count specifies a subexpression in the regular expression, and the value of the function is the ending position of the match for that subexpression. `

looking at the examples, what that means is since I added `actionvalue` before `generate` in the regex, that means that everything after `actionvalue` (including `generate` and all the other values) needs to be incremented, to point to the proper match group.

util/emacs/bsv-mode/bsv-mode-22.el
sheganinans: done

util/emacs/bsv-mode/bsv-mode-22.el
sheganinans: so the commenting here is a bit awkward, that comment actually refers to the next value, and for some reason that was a mis-comment, i changed it to end-task, i think it will make sense given the larger context of this piece of code. its a little confusing but i decided to go with the pattern of the code as is

util/emacs/bsv-mode/bsv-mode-22.el
sheganinans: done

quark17:
Thank you again! Sorry it took so long to accept!

PR #772, opened 2025-04-04
Improve distribution make settings

npe9:
This pull request improves the distribution make settings.


npe9:
Brief update: I'm still trying to figure out why getting rid of the original cabal call in favor of integration with Distribution.make is causing issues. That said, I'm currently yak shaving to get the workflow down to a reasonable time so I can iterate quickly enough to fix this issue in tractable time. To do that I'm looking at caching Docker containers and creating a workflow matrix to speed up the CI/CD workflow. This looks to be getting the check time down to about 4 minutes. I will make a separate pull request with that once I clean it up.

PR #778, opened 2025-04-11
Feature/value plusargs

postoroniy:
added system function $value$plusargs same as used to be in verilog/systemverilog.
supports only 128 bit and can be used to load reg only.
as per bsc none of "let" or "Bit" can not be used  LHS.
added test for function too


postoroniy:
created different PR

PR #779, opened 2025-04-12
Silence "unused result" warning in bluesim probe module

danderson:
In modern glibc, asprintf's return value is annotated as "must use", because it can fail with -1 in OOM conditions.

This is not a correctness issue on modern OSes, since the surrounding code will at worst lead to a null pointer deref and crash in this condition. Further, most modern OSes allocate virtual memory optimistically, and crash later on failure to fault in rather than returning -1 to an allocation call. So, this change's main purpose is to declutter bluesim compilation logs.

### DIFF
--- a/src/bluesim/bs_prim_mod_probe.h
+++ b/src/bluesim/bs_prim_mod_probe.h
@@ -42,7 +42,9 @@ class MOD_Probe : public Module
   unsigned int dump_VCD_defs(unsigned int /* num */)
   {
     char* buf = NULL;
-    asprintf(&buf, "%s$PROBE", inst_name);
+    int sz = asprintf(&buf, "%s$PROBE", inst_name);
+    if (sz < 0)
+      perror("dump_VCD_defs: asprintf");
     vcd_num = vcd_reserve_ids(sim_hdl, 1);
     vcd_set_clock(sim_hdl, vcd_num, __clk_handle_0);
     vcd_write_def(sim_hdl, vcd_num, buf, bits);


danderson:
Sample of the log clutter this removes:

```
In file included from /nix/store/ksabz0s564dzcsiaj04gmkb46iq7s0kn-bluespec-2024.07/lib/Bluesim/bluesim_primitives.h:12,
                 from out/bluesim/test/DMAChannel/mkTB.cxx:7:
/nix/store/ksabz0s564dzcsiaj04gmkb46iq7s0kn-bluespec-2024.07/lib/Bluesim/bs_prim_mod_probe.h: In member function ‘unsigned int MOD_Probe<T>::dump_VCD_defs(unsigned int) [with T = unsigned char]’:
/nix/store/ksabz0s564dzcsiaj04gmkb46iq7s0kn-bluespec-2024.07/lib/Bluesim/bs_prim_mod_probe.h:45:13: warning: ignoring return value of ‘int asprintf(char**, const char*, ...)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
   45 |     asprintf(&buf, "%s$PROBE", inst_name);
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

quark17:
Sounds good, thanks!  Can I suggest, though, not using `assert`, and instead just do something like this:
```
if (sz < 0)
  perror("dump_VCD_defs: asprintf:");
```
I didn't see any place else that uses `assert` in the Bluesim code, and we don't need to ability to turn off the assertion, I wouldn't think, so might as well inline the `abort` call.  It'd be best to match whatever convention Bluesim uses for aborting, but I didn't see many such places (only a few uses of `perror` and nothing else).  I assume that `asprintf` would set `errno` if it failed, so I assume `perror` would be OK to use here.

danderson:
Works for me! I wasn't sure how to report an error from here, since as you say there's no obvious precedent. And, I was reluctant to start doing deeper surgery to plumb error results around for a condition that _really_ isn't ever going to happen :)

Updated the implementation as you suggested, `make install-src` and `make check-suite` are happy here.

PR #783, opened 2025-04-25
Find tcl8 on Fedora-based systems

danderson:
With the release of tcl 9, tcl8 commands get installed with a suffix, e.g. 'tclsh8'. platform.sh rejects tcl 9 as unsupported, and fails to find tcl 8 when installed with a suffix.

Since 'tclsh8' more specifically identifies a tcl version that's definitely supported, look for that first, and then try the generic 'tclsh' as a fallback.

---

With this change, I'm able to `make install-src` on Fedora 42. Without it, the build fails partway because platform.mk fails to find a usable tclsh.

Possible alternative: support tcl 9? The build fails with a specific "unsupported" error when I have tcl9 installed. I assumed that meant that actual work is required to port to tcl9, but if that check is just a case of "it should work, but we want to verify explicitly before allowing it", I can instead try to do that, if you'd prefer.

### DIFF
--- a/platform.sh
+++ b/platform.sh
@@ -81,7 +81,15 @@ if [ ${OSTYPE} = "Darwin" ] ; then
     # Have Makefile avoid Homebrew's install of tcl on Mac
     TCLSH=/usr/bin/tclsh
 else
-    TCLSH=`which tclsh`
+    # With the release of tcl 9, some linux distros (e.g. Fedora)
+    # install tcl 8 tools with a suffix: 'tclsh8' instead of 'tclsh'.
+    #
+    # Since 'tclsh8' more precisely identifies a compatible tcl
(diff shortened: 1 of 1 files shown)


quark17:
I believe the check you're referring to is in `src/vendor/htcl/Makefile`, although let me know if there are other places.  This is probably an error merely because we haven't tried it.  I would say, yes, see if allowing v9 works, including passing the test suite.  The changes in v9.0 that are incompatible don't seem to me to be a worry, but maybe there are some Tcl scripts in the repo that would need to be updated (for example, to not use tilde in paths).

quark17:
I was able to build with Tcl 9.0.2 without any changes, except to disable the version check.  I've open PR #801 to make that change, and I'll close this one.  Can you confirm that the other PR works for you?

PR #739, opened 2024-10-04
Fix windows build.

Febbe:
I finally managed to compile bsv.exe on Windows via MSYS2-ucrt64 (Mingw).
Most changes are related to the make system:
  - `rm -f` does not work for folders the `-r` (recursive option must be added)
  - the `/tmp` dir does not exist on Windows, instead, a local temp folder is created
  - gcc may produce `*.a`(static) and `*.dll.a`(dyn symbols) + `*.dll`(dyn impl) files, but `.dll`(both) and `.lib`(static) are also supported and are more common. `.so` files aren't created.
  - on Windows, hard links can't be updated, once created and must be removed.
  - Paths in windows may contain `:` (e.g.`C:\`), therefore a regex just for `:` will split those paths
Changes to the Haskell code:
 - All changes are regarding POSIX system-specific features that can't be ported to Windows
 - Replaced those packages with windows specific or platform-independent implementations
**DISCLAIMER:** I have zero knowledge about Haskell and utilized ChatGPT. Therefore, please briefly review or rewrite this part for me.
Changes to C++:
 - Mostly replaced POSIX-specific functions with their Windows counterpart.
 - Todo: Thread timer is sort of a hack and should be done via Windows API if possible.
 - Preallocating memory is possible on Windows but fundamentally works differently - wouldn't use it without a custom allocator.
 - Fixed warnings (treated as error) about pruning reinterpret_casts.
Now I am stuck with some errors running ghc in the build process:
```
Febbe@PC-Febbe UCRT64 /c/Users/Febbe/Desktop/ws/bsc-2024.07/bsc
$ make install-src

[...]

make[2]: Entering directory '/c/Users/Febbe/Desktop/ws/bsc-2024.07/bsc/src/comp'
Using tclsh: /ucrt64/bin/tclsh
Using tcl include flags:
Using tcl library flags: -ltcl86 -ltclstub86
Building with GHC 9.4.8
----- Normal build options -----
mkdir -p ./tmp
bsc start Fri Oct 4 23:33:16 CEST 2024
./update-build-version.sh
fatal: No names found, cannot describe anything.
BuildVersion.hs up-to-date
./update-build-system.sh
BuildSystem.hs up-to-date
[...]

PR #784, opened 2025-04-25
INSTALL.md: reorganize build requirements, add instructions for Fedora

danderson:
The reorganized requirements section opens with a full list of required and optional dependencies in plain prose. Following that, separate subsections provide handy cut-and-paste commands to run on Debian, Ubuntu and Fedora in order to install those dependencies, as well as instructions on using Cabal directly to manage GHC libraries.

The sections describing the SMT solvers and how to disable them were also restructured slightly, to heal them from the removal of interleaved package install instructions.

---

This started out as wanting to add install instructions for Fedora systems, so that I wouldn't have to hunt down the right set of packages next time I need them. But adding "Or on Fedora, run ..." at every point in the document that lists and apt-get command really cluttered things up. And so, the change evolved into this PR, which tries to first list "here's the stuff you need" in a system-agnostic way, followed by quick copy-pastable commands for specific systems.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -69,8 +69,7 @@ The following dependencies are optional, though recommended:
  - To build the STP SMT solver: a C/C++ toolchain, Perl, and the
    `flex` and `bison` parser generator tools.
  - To build the Bluespec Tcl shell (`bluetcl`): Tcl development
-   libraries. Version 8 may be required, as we have not yet tested
-   with version 9.
+   libraries (version 9.0, 8.6, or 8.5).
  - To run smoke tests: the [Icarus Verilog] simulator.
  - To run the full test suite: the Icarus Verilog simulator, Perl,
    csh, and SystemC libraries. See the [testsuite
@@ -160,9 +159,6 @@ ghcup install ghc 9.6.7
 cabal update
 cabal v1-install regex-compat syb old-time split
 ```
-
-Note that `tcl8-devel` may be needed in place of `tcl-devel`, until we have
-tested with Tcl v9.
 
 Those final four commands install the recommended GHC compiler version
 and libraries. If you would prefer to install GHC via the system's
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,17 +20,21 @@ We recommend renaming the `inst` directory to `bsc-<VERSION>` and placing
 it in a subdirectory of `/opt/`, `${HOME}/`, `/usr/share/`, or similar
 location.  For example:
 
-    $ BSC_VERSION=$(echo 'puts [lindex [Bluetcl::version] 0]' | inst/bin/bluetcl)
-    $ mkdir -p /opt/tools/bsc
-    $ mv inst /opt/tools/bsc/bsc-${BSC_VERSION}
-    $ cd /opt/tools/bsc
-    $ ln -s bsc-${BSC_VERSION} latest
+```bash
+BSC_VERSION=$(echo 'puts [lindex [Bluetcl::version] 0]' | inst/bin/bluetcl)
+mkdir -p /opt/tools/bsc
+mv inst /opt/tools/bsc/bsc-${BSC_VERSION}
+cd /opt/tools/bsc
+ln -s bsc-${BSC_VERSION} latest
+```
 
 The `inst` directory has a `bin` subdirectory, where the executables
 for the tools are found.  To use the tools, just add that directory to
 your `PATH`:
 
-    $ export PATH=/opt/tools/bsc/latest/bin:$PATH
+```bash
+export PATH=/opt/tools/bsc/latest/bin:$PATH
+```
 
 These executables will make use of other files found within the `inst`
 directory, locating them relatively from the `bin` directory.  That is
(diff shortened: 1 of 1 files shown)


quark17:
Thank you again.  I have pushed some fixes and it is now ready as far as I am concerned, but I would love to get comments from you.

Ihave fixed some typos (including the heading for Fedora being at the wrong level).  I have resolved my concerns from my earlier comments and added some clarifying text.  Perhaps the biggest change is that I replaced the use of dollar signs for example commands with blockquotes, and in particular have annotated them as being `bash` code.  I don't know if specify `bash` is helpful, as it doesn't highlight much, and occasionally highlights weird things, but I guess it could be helpful to identify any typos in our code.  Anyway, I just wanted the document to be consistent, so I matched the style that you introduced for the Fedora commands.  Again, comments on this version are welcome.

INSTALL.md
```diff
@@ -115,57 +137,98 @@ sudo apt-get install \
...
+   texlive-dingbat \
+   texlive-subfigure
+
+curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+ghcup install ghc 9.6.7
+cabal update
+cabal v1-install regex-compat syb old-time split
+```
+
+Note that `tcl8-dev` may be needed in place of `tcl-dev`, until we have
```
danderson, line 164:
```suggestion
Note that `tcl8-devel` may be needed in place of `tcl-devel`, until we have
```

INSTALL.md
```diff
@@ -115,57 +137,98 @@ sudo apt-get install \
...
-        texlive-moreverb \
-        texlive-dingbat \
-        texlive-subfigure
+```bash
+sudo dnf install \
+   @development-tools \
+   @c-development \
+   iverilog \
+   dejagnu \
+   tcl-dev \
```
danderson, line 150:
```suggestion
   tcl-devel \
```

rpm-based distros use the `-devel` suffix for development packages, rather than the `-dev` convention from deb-based distros. I got that wrong in my original PR, sorry.

danderson (commented):
Your revisions look fine to me. Thinking back, I believe a lot of the ghc tool selection text I removed was based on the assumption that all distros ship a modern-enough GHC these days, and thus that using the OS package manager would be preferable. But that's very much a personal preference, and I'm not deeply involved in Haskell these days so I could be completely wrong about how good distro versions are. If your preference for consistency and supportability is to steer towards ghcup, that makes sense to me and your revisions look good for conveying that preference and available alternatives.

Using markdown code blocks sounds reasonable to me, and should make it easier to copy-paste commands.

I ran through the Fedora installation instructions just now, and found one package that I spelled wrong in my original PR, but aside from that the instructions seem to work. Thanks!

INSTALL.md
kenta2: I will be another set of eyes to check over this PR.

kenta2:
I will be another set of eyes to check over this PR.

quark17:
I added a summary for MacOS as well.  In doing so, I noticed some things missing from the Fedora and Debian/Ubuntu commands.  I know that @kenta2 has lists of package that are used to provision the VMs for making releases, and hopefully can fill in any missing packages.  Here's what I noticed:

1. The Ubuntu/Debian and Fedora summaries do not list `systemc` or `asciidoctor`.  These are optional, but the doc says the command installs all the optional requirements.  Should those be added?

2. Also `git` might be worth mentioning, if people need to get the source code!

3. When installing GHC via GHCup, there are dependencies for GHCup, listed [here](https://www.haskell.org/ghcup/install/#system-requirements), such as `curl libffi-dev libffi libgmp-dev libgmp libncurses-dev libncurses libtinfo` for Debian/Ubuntu.

I am not sure if I captured all the packages needed for MacOS, but someone who tries it out will hopefully file an issue if it doesn't work for them.

I also considered adding a subsection for Windows, to briefly mention that it's not available but giving the current status and pointing to PR #739 that has some details.

kenta2:
> I added a summary for MacOS as well. In doing so, I noticed some things missing from the Fedora and Debian/Ubuntu commands. I know that @kenta2 has lists of package that are used to provision the VMs for making releases, and hopefully can fill in any missing packages. Here's what I noticed:
> 
>     1. The Ubuntu/Debian and Fedora summaries do not list `systemc` or `asciidoctor`.  These are optional, but the doc says the command installs all the optional requirements.  Should those be added?
> 
>     2. Also `git` might be worth mentioning, if people need to get the source code!
> 
>     3. When installing GHC via GHCup, there are dependencies for GHCup, listed [here](https://www.haskell.org/ghcup/install/#system-requirements), such as `curl libffi-dev libffi libgmp-dev libgmp libncurses-dev libncurses libtinfo` for Debian/Ubuntu.
> 
> 
> I am not sure if I captured all the packages needed for MacOS, but someone who tries it out will hopefully file an issue if it doesn't work for them.
> 
> I also considered adding a subsection for Windows, to briefly mention that it's not available but giving the current status and pointing to PR #739 that has some details.

1. As release is imminent, let's deal with `asciidoctor` and `systemc` in a separate Issue or Pull Request.  There is no need to get `INSTALL.md` perfect for all cases and all platforms in this PR.  For the record, `asciidoctor` is only needed to build a release, not for `make install-doc`.
2. Similarly, let's deal with whether to include `git` in a separate Issue or Pull Request.  I'm somewhat opposed, because the source code releases should be buildable without `git` (but I've never tested this).  Anyone acquiring the source code by `git clone` from GIThub presumably knows they need `git`.
3. For Fedora at least, the requirements to install and run  `ghcup`  are met by the `dnf` command provided by @danderson .  Let's postpone adding any `ghcup` requirements for Ubuntu into another issue or PR.  Curiously, on Fedora 41 and 42, despite `ghcup` claiming it needs `libgmp-dev` (packaged in `gmp-devel` in Fedora), it works fine without it.  It's only when you try to use `ghc` that it fails (fails to link) missing `libgmp-dev`.

PR #796, opened 2025-08-01
Add more clarity to docs about `inhigh` enable signals.

rossc719g:
It turns out that if you use "?" twice for two different action methods, you get an internal compiler error.  When the docs said the name "does not matter", I assumed it really meant that.  :-)


quark17:
In the future, you don't need to create a new PR.  You can update the branch in your repo and that will be reflected in the PR.

PR #799, opened 2025-08-11
Fix `make install-doc` to fail if any doc fails to build.

Arnavion:
Previously it would only fail if the last doc (BH_ref_guide) failed to build. So if the last doc built successfully but any of the others didn't, the build would still succeed as a whole and lead to missing docs in the final package.


quark17:
Thank you for the fix.  Sorry that you ran into this!  I'll accept the PR, although I think in the long run it might be best to not have a for-loop at all and just have separate targets, so that the build can be done in parallel.

PR #801, opened 2025-08-22
Allow Tcl 9.0

quark17:
BSC builds and runs with new Tcl 9.0 just fine.  The only thing preventing a build is a version check in a Makefile, which needs to be updated.  This replaces PR #783.

### DIFF
--- a/platform.mk
+++ b/platform.mk
@@ -66,6 +66,27 @@ $(error Unable to find tclsh)
 endif
 $(info Using tclsh: $(TCLSH))
 
+# Check that the Tcl version is supported
+# and produce a list of defines
+# (that can be preceded with -D or -optc-D when passed to CC or GHC)
+#
+TCL_VERSION = $(shell $(TOP)/platform.sh tclversion)
+ifeq ($(TCL_VERSION),8.5)
+TCL_DEFS=TCL85
+else
+ifeq ($(TCL_VERSION),8.6)
+TCL_DEFS=
+else
+ifeq ($(TCL_VERSION),9.0)
+TCL_DEFS=TCL9
+else
+$(error Unsupported Tcl version: $(TCL_VERSION)
+endif
+endif
+endif
+$(info Tcl version: $(TCL_VERSION))
+$(info Using Tcl defines: $(TCL_DEFS))
+
 TCL_INC_FLAGS = $(shell $(TOP)/platform.sh tclinc 2> /dev/null || echo fail)
 ifeq ($(TCL_INC_FLAGS), fail)
 $(error Unable to find tcl include directory)
--- a/platform.sh
+++ b/platform.sh
@@ -8,11 +8,12 @@ usage()
     echo ""
     echo "Display canonical platform information"
     echo "Valid options are:"
-    echo "   ostype   "
-    echo "   machtype "
-    echo "   tclsh "
-    echo "   tclinc "
-    echo "   tcllibs "
+    echo "   ostype"
+    echo "   machtype"
+    echo "   tclsh"
+    echo "   tclversion"
+    echo "   tclinc"
+    echo "   tcllibs"
     echo "   c++_shared_flags"
 }
 
@@ -89,12 +90,18 @@ if [ "$1" = "tclsh" ] ; then
     exit 0
 fi
 
+TCL_VERSION=$(echo 'catch { puts [info tclversion]; exit 0}; exit 1' | ${TCLSH})
+if [ "$1" = "tclversion" ] ; then
+    echo ${TCL_VERSION}
+    exit 0
+fi
+
 PKG_CONFIG=`which pkg-config`
 if [ -z "${PKG_CONFIG}" ] ; then
 	PKG_CONFIG='false'
 fi
 
-TCL_SUFFIX=$(echo 'catch { puts [info tclversion]; exit 0}; exit 1' | ${TCLSH})
+TCL_SUFFIX=${TCL_VERSION}
 TCL_ALT_SUFFIX=$(echo ${TCL_SUFFIX} | sed 's/\.//')
 
 if [ "$1" = "tclinc" ] ; then
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -59,6 +59,8 @@ YICES_LIB_FLAGS = -L../vendor/yices/lib -lyices
 HTCL_HS = ../vendor/htcl
 HTCL_INC_FLAGS = -L$(HTCL_HS)
 HTCL_LIB_FLAGS = -lhtcl
+# HTcl.hs needs macro definitions
+HTCL_INC_FLAGS += $(addprefix -D,$(TCL_DEFS))
 
 # -----
 # GHC
--- a/src/vendor/htcl/HTcl.hs
+++ b/src/vendor/htcl/HTcl.hs
@@ -109,6 +109,10 @@ import Foreign.C.String
 import Data.Time.Clock
 import Data.Time.Clock.POSIX
 
+#if MIN_VERSION_base (4,18,0)
+import Foreign.C.ConstPtr
+#endif
+
 import Foreign.ForeignPtr
 import Foreign.ForeignPtr.Unsafe(unsafeForeignPtrToPtr)
 
@@ -123,6 +127,10 @@ import qualified Control.Exception as CE
 traceCalls :: Bool
 traceCalls = False
 
+#if MIN_VERSION_base (4,18,0)
+#else
+type ConstPtr = Ptr
+#endif
 
 -- cast function
 castE :: (Enum a, Enum b) => a -> b
@@ -133,6 +141,15 @@ castI = fromIntegral
 
 castD :: (RealFloat a, RealFloat b) => a -> b
 castD = (uncurry encodeFloat) . decodeFloat
+
+--------------------------------------------------------------------------------
+-- Tcl_Size has changed from Tcl 8 to Tcl 9
+
+#ifdef TCL9
+type CTclSize = CPtrdiff
+#else
+type CTclSize = CInt
+#endif
 
 --------------------------------------------------------------------------------
 -- Common TCL function status
@@ -192,19 +209,19 @@ type PTclObjArray = Ptr PTclObj
 
 
 -- Foreign import functions --     Objects
-foreign import ccall "Tcl_NewIntObj"
+foreign import capi "tcl.h Tcl_NewIntObj"
  tcl_NewIntObj :: CInt -> IO (PTclObj)
 
-foreign import ccall "Tcl_NewWideIntObj"
+foreign import capi "tcl.h Tcl_NewWideIntObj"
  tcl_NewWideIntObj :: CLLong -> IO (PTclObj) -- tcl uses long long
 
-foreign import ccall "Tcl_NewDoubleObj"
+foreign import capi "tcl.h Tcl_NewDoubleObj"
  tcl_NewDoubleObj :: CDouble -> IO PTclObj -- tcl double
 
-foreign import ccall "Tcl_NewStringObj"
+foreign import capi "tcl.h Tcl_NewStringObj"
  tcl_NewStringObj :: Ptr CChar -> CInt -> IO PTclObj
 
-foreign import ccall "Tcl_NewListObj"
+foreign import capi "tcl.h Tcl_NewListObj"
  tcl_NewListObj :: CInt -> Ptr PTclObj ->  IO PTclObj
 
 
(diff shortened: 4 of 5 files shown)


danderson (approved):
With this change, `make full_clean && make release` builds correctly on my Fedora system. I verified that in `tclsh`, `info tclversion` reports `9.0`, and that tcl 8 is not installed. So, looks good to me, thank you!

danderson:
I optimistically posted the previous comment when the build was almost, but not quite complete. And of course, the final linking of bluetcl failed:

```
[149 of 149] Linking bluetcl
/usr/bin/ld: /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp/HTcl.o: in function `HTcl_htclObjToBool_info':
(.text+0x1318): undefined reference to `Tcl_AddObjErrorInfo'
/usr/bin/ld: /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp/HTcl.o: in function `HTcl_zdwzdctoTclObj2_info':
(.text+0x1a39): undefined reference to `Tcl_NewIntObj'
/usr/bin/ld: /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp/HTcl.o: in function `HTcl_zdwzdctoTclObj7_info':
(.text+0x1de4): undefined reference to `Tcl_NewIntObj'
/usr/bin/ld: /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp/HTcl.o: in function `HTcl_zdwzdctoTclObj4_info':
(.text+0xa94c): undefined reference to `Tcl_NewIntObj'
collect2: error: ld returned 1 exit status
ghc-9.6.6: `gcc' failed in phase `Linker'. (Exit code: 1)
```

It's quite possible this is something broken in my build environment, I'm digging into that now.

danderson:
For reference, the ghc linking command, which executes from working directory `src/comp`:

```
ghc -Wtabs -fmax-pmcheck-models=800 -hidir /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp -odir /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp -stubdir /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp -O2 -hide-all-packages -fasm -Wall -fno-warn-orphans -fno-warn-name-shadowing -fno-warn-unused-matches -package base -package containers -package array -package mtl -package unix -package regex-compat -package bytestring -package directory -package process -package filepath -package time -package old-time -package old-locale -package split -package syb -package integer-gmp -package text  -iGHC/posix -iLibs -i../Parsec -i../vendor/stp/include_hs -i../vendor/yices/include_hs -i../vendor/htcl '-tmpdir /tmp'  -I../vendor/stp/include -I../vendor/yices/include -L../vendor/htcl   -lpthread --make bluetcl -j1 +RTS -M4G -A128m -RTS -L../vendor/stp/lib -lstp -L../vendor/yices/lib -lyices  -ltcl9.0 -ltclstub -lhtcl -o bluetcl -no-hs-main -x c bluetcl_Main.hsc
```

libtcl9.0 and libtclstub are both available in /usr/lib64, so at a glance at least the linker flags look plausible.

danderson:
Aha! Looks like in my installation, that function is a C preprocessor define, not a function in its own right:

```
$ grep -R Tcl_NewInt /usr/include/*.h
tclDecls.h:#define Tcl_NewIntObj(value) Tcl_NewWideIntObj((int)(value))
```

I'm not well versed in Haskell FFI, but I do see HTcl.hs does a foreign import of Tcl_NewIntObj, and I presume that import assumes that a corresponding symbol is defined - and thus is defeated by the #define.

Given that #define, it looks like the two functions are compatible enough that we could just redefine tcl_NewIntObj in HTcl in terms of tcl_NewWideIntObj, but I'm not sure how to square that with Haskell's type system, since one function takes a CInt whereas the other takes a CLLong.

danderson:
I quickly hacked up https://github.com/B-Lang-org/bsc/commit/ec552baed0c560a98b5ea136124fc1f0bebc5016 , which makes HTcl use the adapted-to functions, effectively replicating the effect of the preprocessor defines in Haskell.

With that patch, bluetcl builds successfully and `make check-smoke` passes. I have `check-suite` running now, but it'll take a while. So far the systemC tests fail, but I think that's because I'm missing systemC toolchain bits and pieces. Everything else is passing so far, fingers crossed.

CAUTION: this is the first Haskell I've written in fifteen years, so I'm sure it's rubbish. Aside from obvious things like not using monadic operators to make the code read nicely, I'm not sure if htcl_AddObjErrorInfo needs to do more work to refcount/clean up the Tcl string that it constructs and passes to tcl_AppendObjToErrorInfo. As the code stands, that might be a memory leak? Or perhaps Tcl convention is that passing the string to AppendObjToErrorInfo makes it that function's problem to clean up? I'm unsure.

If the HTcl changes look directionally right, I can probably also scrounge up a tcl8 install and check that things still work with the changes there. I believe all the functions involved already existed in tcl8, so it feels like it should all be fine?...

quark17:
I am trying to recreate my build of BSC with Tcl9 and I am getting the same errors that you reported.  I foolishly deleted the directory where I thought I had things working before, so I can't see what I had done there.

The GHC documentation on FFI does have a section ["Importing C functions that turn out to be CPP macros"](https://wiki.haskell.org/GHC/Using_the_FFI#Importing_C_functions_that_turn_out_to_be_CPP_macros).  It presents us with several possible workarounds:

(1) If we use the GHC flag `-fvia-C` to cause GHC to generate C code for the backend, then the error does not come up.  I'm not sure if this can be done just for the HTcl module, or whether there would be a performance impact.

(2) If we stay with the native code generation, then a workaround is to import wrappers that call the macro functions, rather than importing the macro functions directly.  This would be easy to do, by putting the wrappers in the `src/vendor/htcl/libhtcl.a` library.

(3) We could avoid the wrappers by just calling other functions (inlining the macros), as you have demonstrated in your patch.  This works for Tcl9 and appears to also work for Tcl8.  However, for Tcl8, it would not be calling the same function as before, and I don't know if there is a performance penalty for that (on any of the machines that we care about)?  Also, this would hardcode the macro definition, where a wrapper would not.  (For example, macOS doesn't use the macro, so maybe has different implementations for NewInt and NewWideInt.)

(4) An alternative to (3) is to use ifdefs to choose between the Tcl8 and Tcl9 FFI imports.  The only downside is that this would still hardcode the Linux macro -- although maybe the OS could be part of the ifdef condition?

I'm starting to come around to the use of a wrapper to resolve this.  But I'm interested in what others think.

quark17:
Oops, I don't think `-fvia-C` is an option any more.  This section from the GHC manual on [using header files with FFI](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/ffi.html#using-header-files) also explains again that header files are not consulted, so the FFI calls are not subject to macro expansion.

quark17:
Despite everything I said, there does appear to be a [`CApiFFI`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/ffi.html#extension-CApiFFI) extension to GHC that allows you import function with `capi` rather than `ccall`; this allows specifying the C header and can therefore support calls that are macros.  This extension has been around since GHC 7.6.1, so it's available for all the GHC versions we support.  However, the code generation didn't handle `const` types until GHC 9.6 when [`ConstPtr`](https://hackage.haskell.org/package/base/docs/Foreign-C-ConstPtr.html) was added.  I don't know if earlier versions would just have a warning or would produce an error?  Here is a [discussion post](https://discourse.haskell.org/t/constptr-compatibility/8649) of people posting the ifdefs they use to handle it for older GHC versions, that they are not fully happy with.

I think we need to handle the possibility of macros, even for Tcl 8.  See, for example, PR #739 that attempts to get BSC compiling for Windows, where the FFI macro issue was encountered (and the suggestions about `CApiFFI` and `ConstPtr` were made).

I was therefore going to write C wrappers for all of the functions that we import.  But it appears that `CApiFFI` automates those wrappers.  So maybe we should use `CApiFFI` instead.  If this doesn't work with GHC before 9.6, then we could ifdef to choose between `CApiFFI` and wrappers. (Or choose not to support GHC before 9.6?)  That's assuming that `CApiFFI` is preferable to handwritten wrappers; if there is no benefit, then it'd be simpler to just use handwritten wrappers.

I'll start by hand-writing wrappers and pushing a commit for that, to this PR.  If we later want to replace that with `CApiFFI` in some way, we can.

danderson:
The build fails for me:

```
[  5 of 148] Compiling HTcl             ( ../vendor/htcl/HTcl.hs, /var/home/dave/hack/sentinel65x/bsc/src/comp/../../build/comp/HTcl.o ) [Missing object file]
/tmp/ghc15633_0/ghc_117.c: In function ‘ghczuwrapperZC5ZCmainZCHTclZCTclzuListObjGetElements’:

/tmp/ghc15633_0/ghc_117.c:41:154: error:
     error: passing argument 3 of ‘Tcl_ListObjGetElements’ from incompatible pointer type [-Wincompatible-pointer-types]
       41 | HsInt32 ghczuwrapperZC5ZCmainZCHTclZCTclzuListObjGetElements(void* a1, Tcl_Obj* a2, long long* a3, Tcl_Obj*** a4) {return Tcl_ListObjGetElements(a1, a2, a3, a4);}
          |                                                                                                                                                          ^~
          |                                                                                                                                                          |
          |                                                                                                                                                          long long int *
   |
41 | HsInt32 ghczuwrapperZC5ZCmainZCHTclZCTclzuListObjGetElements(void* a1, Tcl_Obj* a2, long long* a3, Tcl_Obj*** a4) {return Tcl_ListObjGetElements(a1, a2, a3, a4);}
   |                                                                                                                                                          ^
In file included from /usr/include/tcl.h:2439,

                 from /tmp/ghc15633_0/ghc_117.c:7:0: error: 

/usr/include/tclDecls.h:1785:61: error:
     note: expected ‘Tcl_Size *’ {aka ‘long int *’} but argument is of type ‘long long int *’
     1785 |                                 Tcl_Obj *listPtr, Tcl_Size *objcPtr,
          |                                                   ~~~~~~~~~~^~~~~~~
     |
1785 |                                 Tcl_Obj *listPtr, Tcl_Size *objcPtr,
     |                                                             ^
/tmp/ghc15633_0/ghc_117.c: In function ‘ghczuwrapperZC6ZCmainZCHTclZCTclzuGetStringFromObj’:

/tmp/ghc15633_0/ghc_117.c:43:119: error:
     error: passing argument 2 of ‘Tcl_GetStringFromObj’ from incompatible pointer type [-Wincompatible-pointer-types]
       43 | char* ghczuwrapperZC6ZCmainZCHTclZCTclzuGetStringFromObj(Tcl_Obj* a1, long long* a2) {return Tcl_GetStringFromObj(a1, a2);}
          |                                                                                                                       ^~
          |                                                                                                                       |
          |                                                                                                                       long long int *
   |
43 | char* ghczuwrapperZC6ZCmainZCHTclZCTclzuGetStringFromObj(Tcl_Obj* a1, long long* a2) {return Tcl_GetStringFromObj(a1, a2);}
   |                                                                                                                       ^

/usr/include/tclDecls.h:1754:43: error:
     note: expected ‘Tcl_Size *’ {aka ‘long int *’} but argument is of type ‘long long int *’
     1754 |                                 Tcl_Size *lengthPtr);
          |                                 ~~~~~~~~~~^~~~~~~~~
     |
1754 |                                 Tcl_Size *lengthPtr);
     |                                           ^

<no location info>: error:
    `gcc' failed in phase `C Compiler'. (Exit code: 1)
```

This is with `gcc (GCC) 15.2.1 20250808 (Red Hat 15.2.1-1)`, ghc 9.6.6 installed from Fedora's packages, and tcl 9.0.0. I set up a separate build sandbox that uses ghcup instead of Fedora's packaged GHC, and it fails in the same way.

Incidentally, I have a hypervisor with spare resources kicking around. I could provision you a Fedora install with ssh access, if it'd help work this out without having to round-trip through me?

quark17:
> The build fails for me:

Ok, this could be my fault for using `CLLong` instead of `CLong` in defining `CTclSize` (line 152 of `HTcl`).  If you change that, does it then work?

quark17:
> > The build fails for me:
> 
> Ok, this could be my fault for using `CLLong` instead of `CLong` in defining `CTclSize` (line 152 of `HTcl`). If you change that, does it then work?

Actually, since `Tcl_Size` seems to be defined as `ptrdiff_t` in Tcl 9, we should use `CPtrdiff`.  That works on my machine.  If you make that change, does it work?  That is, in `src/vendor/htcl/HTcl.hs`
```
#ifdef TCL9
type CTclSize = CPtrDiff
```

quark17:
Awesome.  FYI, the testsuite README has a section on parallelizing.  You should be able to run `make -j 8 checkparallel`, say.  It runs each directory in parallel; all of the output will be jumbled, but I think there's a process at the end that collects up all the individual directory results into one report.

Issue #803, opened 2025-08-25
iverilog VPI error for user defined task with no args


timksf:
Hi,
VPI code for a user-defined system task that has no args errors on `vpi_free_object`  in the generated `*_calltf` function.

This can be observed by removing the parameter of the imported function `my_time` in `testsuite/bsc.options/GCD.bsv` and then running the tests:

```
vvp: vpi_priv.cc:261: PLI_INT32 vpi_free_object(vpiHandle): Assertion `ref' failed.
```

The cause is that `vpi_iterate` returns `NULL` for an empty args list, leading to the failing assertion.

PR #804, opened 2025-08-25
fix vpi for user defined system task with no args

timksf:
see #803

I just noticed  that this fix collides with #802, let me know how you want to handle this, I could also just create a new imported function or test case.


quark17:
Oh, wow!  Thank you so much for this!  Because earlier versions of iverilog didn't fully support VPI, I just assumed that this failure was still something unsupported, and didn't look into whether BSC's code generation might be wrong.  This is fantastic.

Assuming that the test suite passes (which I have just kicked off), I am happy to accept this PR and cancel #802.  The "options" testcase doesn't care what the design does, it's just testing the messages that BSC prints, so removing the argument is just as fine as fully adding the argument.

Because we assumed this was a problem with iverilog, there are various tests which are disabled when the vsim is iverilog.  Those can be re-enabled now.  I am searching through the testsuite manually to find them, but also hopefully the testsuite run will report unexpected passes (XPASS) for those now, and should automatically alert us to them.

PR #811, opened 2025-09-11
Switch from Hyper to NFData

gkaracha:
Supersedes #807.

Since I'm not familiar enough with the codebase to be sure that forcing more things won't give rise to cycles/non-termination, this PR preserves the existing evaluation semantics. 

I tried to derive the `NFData` instances as much as was possible, and wrote manual ones for the remaining cases (e.g., to avoid the class between `DeriveAnyClass` and `GeneralizedNewtypeDeriving`, or because the existing `Hyper` instance only forced parts of the input).

I also added a few `-- XXX` comments, following the local coding style, because some fields I thought would be evaluated actually weren't, and as far as I could tell that was most likely by accident rather than a deliberate choice—perhaps the desired behavior is to force them after all.

### DIFF
--- a/src/comp/APaths.hs
+++ b/src/comp/APaths.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveAnyClass #-}
+
 module APaths(
               PathGraphInfo,
               PathUrgencyPairs, PathNode,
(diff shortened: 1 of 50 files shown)


blackgnezdo:
@quark17 do you have any thoughts about his change?

blackgnezdo:
@gkaracha if you get a chance to address the comments hopefully this can land promptly for us to get this via upstream?

PR #828, opened 2025-10-23
Change RSchedule to use maps instead of lists for lookups.

nanavati:
Bad performance cases for this may not be common, but I have encountered one.

### DIFF
--- a/src/comp/RSchedule.hs
+++ b/src/comp/RSchedule.hs
@@ -94,7 +94,7 @@ rSchedule :: Id -> ResourceFlag -> M.Map MethodId Integer -> MethodUsesMap ->
              (RuleId -> RuleId -> Bool) -> ErrorMonad (RAT,[RRM])
 rSchedule moduleId rFlag rMaxs rMap areSimult =
     -- We don't need to worry about keys conflicting when combining xxs
-    -- because those keys came from the keyd of rMap.
+    -- because those keys came from the keys of rMap.
     let concatTuple (xxs,yys) = (M.unions xxs, concat yys)
         f = rSchedule' moduleId rFlag rMaxs areSimult
     in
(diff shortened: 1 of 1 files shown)


quark17:
Cool, sounds reasonable.  Code looks fine, but I have one question.

The expected Verilog changed in one example, and a signal name changed from `MUX_accumulator$write_1__SEL_1` to `MUX_multiplier$write_1__SEL_1`.  These are both submodules (`accumulator` and `multiplier`) and I would guess that the conditions under which they are being written are the same, so CSE has unified the mux selectors for both into one definition, under one name.  I would guess that the choice of name depends on the order they appear in a list, and that order has changed.

All good, but I worry that the order may not be stable, and perhaps we care about that?  There are two places where the map is flattened: one is for printing, where `sortRatForPrint` is applied, and the other is in AState, where no sorting is performed.  It's probably inefficient to sort the entire output of `M.toList` there, but I wonder if we could one of the following: sort the list somewhere later (for example, sort inside `mkBlob` that is being immediately mapped to the list) or preserve information as map for longer (and sort when it ultimately stops being a map).  (I haven't taken time to remind myself what `mkBlob` does, to know if there's an obvious answer here.)

src/comp/RSchedule.hs
```diff
@@ -267,11 +282,11 @@ colorFw rMax dropEdges st g
...
 
 -- backward pass: pick up vertices and color them
-colorBk :: [Integer] -> [(UniqueUse,Integer)] -> [RRM] -> StkL ->
-           ErrorMonad ([(UniqueUse, Integer)], [RRM])
+colorBk :: [Integer] -> M.Map UniqueUse Integer -> [RRM] -> StkL ->
+           ErrorMonad (M.Map UniqueUse Integer, [RRM])
 colorBk _  cs es [] = return (cs,es)
 colorBk rMaxL cs es (Vertex vns@(v,_) : vs) =
-    colorBk rMaxL ((v, pickColor rMaxL cs vns):cs) es vs
+    colorBk rMaxL (M.insert v (pickColor rMaxL cs vns) cs) es vs
```
quark17: This silently replaces an existing key.  In keeping with your use of internal errors elsewhere, did you want to use an insert function reports an internal error when the key exists?

quark17: I don't feel strongly.  I can go with whatever you want.

A comment wouldn't hurt.  I assume there's no performance impact to using a function with an error.

nanavati: I ended up taking out the internalError for rSchedule and switching to a comment because I thought the code looked nicer and closer to what was there before.

PR #833, opened 2025-10-27
Implement LiberalTypeSynonyms as GHC does

### LINKED ISSUE #221: bsc Internal Compiler error
The following 

```
package ZipCrash where

import List

class Functor f where
  map :: (a -> b) -> f a -> f b

class Zip f where
  toListX   :: f a -> List a
  fromListX :: List a -> f a

  zipXWith :: (a -> b -> c) -> f a -> f b -> f c
  zipXWith f xs ys = fromListX (List.zipWith f (toListX xs) (toListX ys))

interface Bar at bt =
  a :: at
  b :: bt

type Foo t = Bar t t

instance Functor Foo where
  map f x = Foo (f x.a) (f x.b)

instance Zip Foo where
  zipXWith f x y = Foo (f x.a y.a) (f x.b y.b)
```

causes `bsc ZipCrash.bs` to crash with

```
Internal Bluespec Compiler Error:
Please report this failure to your Bluespec technical contact.
If you do not know your contact, you may email support@bluespec.com.
The following internal error message should be included in your
correspondence along with any other relevant details:
Bluespec Compiler (build 02ff76b)
expandSyn,truncType
(1, 0, ZipCrash.Functor ZipCrash.Foo, ZipCrash.Bar (TGen 0) (TGen 0))
```

nanavati:
Fixes #221 and several other more complex examples (included in the pull request).

### DIFF
--- a/testsuite/bsc.bugs/github/gh221/gh221.exp
+++ b/testsuite/bsc.bugs/github/gh221/gh221.exp
@@ -35,3 +35,6 @@ compile_pass ShouldCompile.bs
 # Now it fails with an accurate partially-applied type synonym error.
 compile_fail_error ZipCrash2.bs T0013
 compare_file ZipCrash2.bs.bsc-out
+
+# If the type synonym is not partially applied, the example compiles.
+compile_pass ZipNoCrash3.bs
--- a/testsuite/bsc.bugs/github/gh221/ZipNoCrash3.bs
+++ b/testsuite/bsc.bugs/github/gh221/ZipNoCrash3.bs
@@ -1,0 +1,24 @@
+package ZipNoCrash3 where
+
+import List
+
+class Functor f where
+  map :: (a -> b) -> f a -> f b
+
+class Zip f where
+  toListX   :: f a -> List a
+  fromListX :: List a -> f a
+
+  zipXWith :: (a -> b -> c) -> f a -> f b -> f c
+  zipXWith f xs ys = fromListX (List.zipWith f (toListX xs) (toListX ys))
+
+interface Bar at =
+  a :: at
+
+type Foo = Bar
+
+instance Functor Foo where
+  map f x = Foo { a = f x.a }
+
+instance Zip Foo where
+  zipXWith f x y = Foo { a = f x.a y.a }


quark17:
Did you want me to preserve the separate commits when merging this? And, if so, did you plan to clean up the commit history before I do that?  I think separate commits is useful here, but some cleanup commits should be merged into prior commits they are cleaning up (like the updating of Bug1263 and the adding of comments on the new Error constructor).

Also, can you provide an explanation of what this feature (LiberalTypeSynonyms) means in BSC?  I'm having trouble understanding what is allowed or not allowed.  I see the GHC docs for [Liberalised type synonyms](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/liberal_type_synonyms.html#liberalised-type-synonyms), but BSC doesn't have features like forall, so a lot doesn't apply.  And I'm having trouble reconciling what ought to apply.

For example, why does `Test.bs` fail?  The GHC docs say "You can apply a type synonym to a partially applied type synonym", so why does BSC report a partial application of `Identity` in the definition of the synonym `X`?

testsuite/bsc.bugs/github/gh221/gh221.exp
```diff
@@ -0,0 +1,19 @@
...
+compile_fail_error ZipCrash.bs T0013
+compare_file ZipCrash.bs.bsc-out
+
+compile_pass Example1.bs
+
+# The type synonyms are fine, but there is another error
+compile_fail_error Example2.bs T0032
+compare_file Example2.bs.bsc-out
+
+compile_pass Example3.bs
```
quark17: If I understand, Example3 is the same as Example2, but the "other error" has been corrected (missing deriving of Bits), and thus Example3 is proving that Example2 compiles fine.  Could we put a comment about that here?

(Example2 might also be testing that the type synonym is fine and the compiler has gotten to the point of context errors, and further testing what the context error looks like -- type synonyms have been expanded and don't appear.)

quark17 (commented):
A few requests for additional comments in the testsuite, below.

And I think I understand a little more.  The reason that `ZipCrash.bs` has an error (about partially applied type synonyms) is because the context doesn't allow it: typeclass instance arguments can't have partial application.

And the reason that `Test.bs` crashes is because the context doesn't allow it: struct type arguments can't be partially applied.

(I wonder if we could include this context explanation in the error message?  The position helpful, though perhaps subtle -- you updated the position to point to `Identity` in `type X = W Identity` but, if I understand correctly, the problem is not that `Identify` can't be partial in the definition of `X`, only that `Identity` can't be partial as an argument to `W`.)

And can we include a variant of `ZipCrash` that defines `type Foo t = Bar t`, without the double `t`?  Does BSC still complain about partial application in that case?  Or is BSC able to notice when a typesynonym has the same type arguments as its expansion and thus allow expanding the partial application?

quark17:
Awesome, thank you!  One question: do you care enough to make the error message more sophisticated to drop the `s` on `1 arguments`?

Thank you for the explanation.  Yes, that helps.  We can separately think about whether that needs to go in the documentation somewhere.  I notice, for BSV, the language guide has section 7.1 on "Type synonyms" that says, "The new type and the original type can be used interchangeably anywhere", which I guess isn't quite true -- as ZipCrash2 shows.

Now I'm curious if we should try a ZipCrash3 where the type synonym is `type Foo = Bar` -- that should compile without an error, yeah? because the synonym itself has no arguments to be partially applied?

nanavati:
ZipCrash3 surprisingly does not compile. Instead it fails with an unbound constructor error. I'm guessing that means the synonym isn't being expanded somewhere. I left the example in as an expected failure, but did not feel it was a significant enough issue to be worth filing separately.

I looked at Error.hs and none of the other places that count arguments are careful about the plural, so I don't want to be the first. Presumably we need a utility that count arguments and generates the right version for you (and then we could use it everywhere).

nanavati:
It turns out "Unbound constructor" was the correct, but misleading error. If you change to struct / interface syntax, the example compiles as we expect. I've updated the tests accordingly. 

There probably should be a better error message for "Unbound constructor" when there is a struct of the same name in scope, but that's not an issue for this PR.

quark17:
I'm confused why `Foo {}` is accepted.  My understanding is that `Foo` there is not a type, so I would not expect synonym expansion to be performed there; it's a name, which I would assume is either the name of a type or the name of a constructor (which is identical to the type name).

I would guess that it's because `TIMonad.disambiguateStruct` is using `findTyCon` to lookup the type constructor, and that has an arm for type synonyms that allows it to keep going if the synonym has a left constructor to follow.

I'm not sure it's the correct thing to do, but it doesn't seem harmful, as I don't think it can really be used except in the case of synonyms like `type A = B`, so I don't feel strongly about cleaning it up.  Any thoughts?

PR #837, opened 2025-10-31
Make zero-size wires carry zero values (like zero-size regs), not undefined

### LINKED ISSUE #836: Values of size 0 are not consistently treated as 0
The behavior of a zero-size BypassWire is inconsistent with that of a zero-size register, as illustrated by the attached program.

When compiled, the `$display` controlled by a `Reg (Bit 0)` prints, but the one for a `BypassWire (Bit 0)` does not, because the `BypassWire` has an undefined value instead of 0.

This behavior creates problems with polymorphic code instantiated with zero-size types.

[BypassWire0.bs.txt](https://github.com/user-attachments/files/23247598/BypassWire0.bs.txt)

nanavati:
Fixes #836

### DIFF
--- a/src/comp/ITransform.hs
+++ b/src/comp/ITransform.hs
@@ -493,9 +493,10 @@ iTrAp ctx (ICon _ (ICPrim _ p)) _ [e, e'] | isLT p, e == e' = (iFalse, True)
 
 -- XXX: preserve the _ "kind"? Or treat different _ differently?
 -- _ == e --> _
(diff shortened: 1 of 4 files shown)


nanavati:
Good point. When I looked, I found a buggy optimization rule in ITransform, which I fixed. Fixing that would have fixed the wire examples (though I definitely think a belt-and-suspenders approach is warranted with them), but the broken rule was causing other problems (as my new testcase demonstrates).

I looked a bit more and that was the only issue I could find. ITransform is where this sort of thing happens in ISyntax and while zero-sized values can survive into ASyntax we don't do much with them there. The only two things I found were:
1. Explicitly injecting a literal 0 of size 0 in AConv for method return values
2. Turning 0-bit expressions into 1-bit expressions with value 0 if they survive all the way to Verilog generation.

PR #838, opened 2025-11-04
Teach bluetcl Bluespec Haskell syntax.

nanavati:
The syntax MutableVar is exposed to bluetcl via new commands. Every place where bluetcl used a pv printing function it now uses a pf function. Output syntax is directly adjusted in a few places.

When bluetcl parses types or constructors, it now supports Bluespec Haskell and Bluespec SystemVerilog syntax. Parses using the current mode are preferred, but if parsing fails in one syntax the parser is permissive and tries to parse with the other syntax.

### DIFF
--- a/testsuite/config/bluetcl.tcl
+++ b/testsuite/config/bluetcl.tcl
@@ -43,7 +43,7 @@ proc bluetcl_exec { cmdline outname } {
     set output [make_bluetcl_output_name $outname]
     set cmd "$bluetcl $cmdline >& $output"
     verbose "Executing: $cmd" 4
-    set status [exec_with_log "run_bluetcl" $cmd 2]
+    set status [exec_with_log "bluetcl_exec" $cmd 2]
     cd $here
     return [expr $status == 0]
 }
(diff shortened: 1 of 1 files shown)


quark17:
Cool! Thanks!

There is one failure in the BSC testsuite because `bsc.bluesim/interactive/mkTest_help.out.expected` needs to be updated.

There are also failures in the BDW testsuite, because its uses of `bluetcl_run_compare_pass` etc will need to be updated.  Unfortunately, we don't have a good way to sync PRs to the two repos.  My suggestion is to make a PR on the BDW testsuite, so that I can see your proposed changes there, and if you confirm that those changes pass with the updated BSC, then I'll accept the BSC PR and re-run the CI for the BDW PR.  Is that reasonable?

nanavati:
Added some guard code around `Bluetcl::syntax set` because it does not work reliably if the syntax has already been read. Even `NOINLINE`, `-fno-cse` and `-fno-full-laziness` didn't change anything.

Also added basic documentation for `Bluetcl::syntax` to the user guide.

quark17:
Thank you!

PR #840, opened 2025-11-04
Fix the zero-size boundary case in List.oneHotSelect

### LINKED ISSUE #839: List.oneHotSelect does not work when selecting among 0-bit things
The attached example fails with the following error when attempting to compile to Verilog (or Bluesim, presumably, but I didn't check that):

[OneHotSelectZero.bs.txt](https://github.com/user-attachments/files/23328234/OneHotSelectZero.bs.txt)

```
Error: "List.bs", line 436, column 26: (S0015)
  Bluespec evaluation-time error: index -1 out-of-range (bit extraction - high
  index)
  During elaboration of the body of rule `' at "OneHotSelectZero.bs", line 24,
  column 8.
  During elaboration of `mkOneHotSelectZero' at "OneHotSelectZero.bs", line
  11, column 0.
```

nanavati:
Fixes #839

### DIFF
--- a/src/Libraries/Base1/List.bs
+++ b/src/Libraries/Base1/List.bs
@@ -433,7 +433,8 @@ oneHotSelect bs xs =
            widen b =
                letseq wb :: Bit sa1
                       wb = signExtend b
-               in     wb[(valueOf sa - 1) : 0]
+                      ix = valueOf sa - 1
+               in if ix >= 0 then wb[ix : 0] else 0
     in  unpack (fold (|) (zipWith (\ b x -> widen (pack b) & pack x) bs xs))
 
 --@ Update an element in a list.
--- a/testsuite/bsc.bugs/github/gh839/Makefile
+++ b/testsuite/bsc.bugs/github/gh839/Makefile
@@ -1,0 +1,5 @@
+# for "make clean" to work everywhere
+
+CONFDIR = $(realpath ../../..)
+
+include $(CONFDIR)/clean.mk
--- a/testsuite/bsc.bugs/github/gh839/OneHotSelectZero.bs
+++ b/testsuite/bsc.bugs/github/gh839/OneHotSelectZero.bs
@@ -1,0 +1,26 @@
+package OneHotSelectZero where
+
+import qualified List
+import Vector
+
+type Fin n = UInt (TLog n)
+
+type NUM_TAGS = 8
+
+{-# verilog mkOneHotSelectZero #-}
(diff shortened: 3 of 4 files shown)


nanavati:
The point of `oneHotSelect` is specifically to not get any muxes like you have with `r0`. We want to do the selection with a sea of ANDs and ORs only. I think your version would be a step back in hardware quality that we do not want.

quark17:
Sorry, the generated code that I showed was still bound up with your specific example that was reintroducing a check of the index.

What is currently generated is this (for 4-bit elements):
```
result = r0 | r1 | r2 | ... ;

r0 = {4{idx[0]} & elem0;
r1 = {4{idx[1]}} & elem1;
r2 = {4{idx[2]}} & elem2;
```
although technically, it doesn't generate that, it extends and then truncates, and BSC doesn't seem to have optimized that away, so what we actually get is this:
```
r0 = s0[3:0] & elem0;
s0 = {5{idx[0]}};

r1 = s1[3:0] & elem1;
s1 = {5{idx[1]}};
```

and my code generates this (currently):
```
result = r0 | r1 | r2 | ... ;

r0 = idx[0] ? elem0 : 0 ;
r1 = idx[1] ? elem1 : 0 ;
r2 = idx[2] ? elem2 : 0 ;
```

Both compute a result independently for each bit of the index and combine with (`|`) rather than a mux -- this is an optimization that requires knowing that the bits of the index are one-hot.

You are pointing out that, in my code, there is still a ternary operator (mux) that could be represented instead using AND (`&`).

That, however, is an optimization that does not depend on any knowledge about the one-hot-ness of the index bits -- but correct me if I'm wrong?  I believe it is always legal to replace `(c ? e : 0)` with `{N{c}} & e`.  Downstream tools should be able to apply this optimization.  If they are not, then BSC's Verilog backend could perform this optimization (and then both my function and the existing function would generate the same code -- or rather, mine would be slightly cleaner, because it wouldn't be extending and truncating).

BSC has code for doing optimizations in the Verilog backend, but nearly all of it is turned off! because we've observed that is has no effect on QoR: anything that BSC can do at the Verilog level, the downstream tools can also do, and probably better.  Where we have outside knowledge that's not visible, we should use that -- and the OR is an example of that.  But I don't believe the AND is an example of that.  Do you believe that downstream tools would be inefficient when given `(c ? e : 0)` and would do better with `{N{c}} & e`?   If you think so, should we not turn on that optimization in BSC in all cases, not just this one?

PR #845, opened 2025-11-13
Add lookupAll utility

krame505:
Add a simple utility function for looking up all values in an association list.

### DIFF
--- a/doc/libraries_ref_guide/LibDoc/List.tex
+++ b/doc/libraries_ref_guide/LibDoc/List.tex
@@ -385,6 +385,21 @@ function Maybe#(b_type)
 \hline
 \end{tabular}
 
+\index[function]{List!lookupAll}
+\index{lookupAll@\te{lookupAll} (\te{List} function)}
+\begin{tabular}{|p{.7 in}|p{4.9 in}|}
+\hline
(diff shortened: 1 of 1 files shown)


quark17:
FYI, documentation is not created from the literate comments; we'd need to update `doc/libraries_ref_guide/LibDoc/List.tex`.  I can push a commit for that, unless you want to?

Anyway, cool.  Sounds like a reasonable addition, although, for judging whether something belongs in the libraries and by what name, my usual metric isn't working here.  I look to see what Haskell/GHC do, and the name `lookupAll` isn't anywhere in GHC, just in a few contributed Hackage packages, and I don't see anything with that type either (or more general, that would still apply to List).  It seems reasonable, though -- but did you have particular thinking that led you to the name and submitting a PR?

krame505:
Huh, I totally thought it was in the Haskell standard lib, for some reason.  It's a function in the [Silver programming language](https://github.com/melt-umn/silver) that I find myself using frequently.

I can update the docs

rsnikhil:
Both the name and code are ok by me.

Re. the name, I did a quick scan of OCaml, Racket and Python libraries and didn't find any common name for this function.  Even the single-value retrieval functions are variously called 'find', 'assoc', 'lookup', 'match', ...) in various languages and by various authors.

PR #847, opened 2025-11-20
Unused import warning

nanavati:


src/comp/BinUtil.hs
```diff
@@ -53,7 +53,7 @@ type BinMap a = M.Map String (BinFile a)
 readImports :: ErrorHandle -> Flags -> BinMap a -> HashMap -> CPackage ->
                IO (CPackage, BinMap a, HashMap)
 readImports errh flags binmap0 hashmap0
-            (CPackage pkgId exps imps fixs ds includes) = do
+            (CPackage pkgId exps imps _old_impsigs fixs ds includes) = do
```
quark17: Is `_old_impsigs` always going to be empty?  You removed an internal error for `CImpSign` below, which I think would be equivalent to having an internal error here if `_old_impsigs` is not `[]`.  Do you think that's worth adding back?

src/comp/Deriving.hs
```diff
@@ -956,12 +956,12 @@ addAutoDeriv :: Flags -> SymTab -> Id -> [CType] -> Id -> [CTypeclass]
                  -> [CTypeclass]
 addAutoDeriv flags r i tvs clsId derivs
                          -- incoherent matches are resolved *after* reducePred
-    | Right True <- fst (runTI flags False r check) = derivs
+    | (Right True, _, _) <- runTI flags False r check = derivs
```
quark17: You added an import of `fst3` but then didn't use it here, where presumably you meant to? (Ironically adding a GHC warning of the unused import.)

src/comp/GenSign.hs
```diff
@@ -35,22 +35,27 @@ import TypeCheck(qualifyClassDefaults)
 
 -- only exports what the user asked to export
 -- (auto-exports everything if the user said nothing)
-genUserSign :: ErrorHandle -> SymTab -> CPackage -> IO CSignature
-genUserSign errh symtab cpkg =
+genUserSign :: ErrorHandle -> SymTab -> CPackage -> IO (CSignature, S.Set Id)
+genUserSign errh symtab cpkg@(CPackage pkgName _ _ _ _ _ _) =
```
quark17: Ah, this has to be positional because `CPackage` doesn't have named fields.  Do you think this is a good excuse to add named fields?  Although I notice that most of `CSyntax` doesn't have named fields.  I guess if only a few places are using individual fields like this, then maybe it's fine to leave it unnamed and positional.

src/comp/GenSign.hs
```diff
@@ -681,3 +694,39 @@ classToIClass i k (Class { csig=tvs, super=ps, funDeps2=bss2,
         CIclass incoh ps' (IdKind i k) tvis fds poss
 
 -- ---------------
+-- Package usage tracking for unused import warnings
+
+-- | Extract packages used by exports (for Phase 3 of unused import detection).
```
quark17: What is the vertical bar at the start of this comment?

quark17: The mention of "Phases" was confusing, because this is the only place in the code that it appears (and doesn't say what Phase 1 is) and the PR had no description text.  But I see that the commit message has the three phases -- I should have looked there.  Although I dunno if maybe there's a place in the code to also include that info.

src/comp/Pred.hs
```diff
@@ -206,17 +208,17 @@ instance Ord Class where
 
 -- someone should comment what all these
 -- things are that go into an Inst.
-data Inst = Inst CExpr [TyVar] (Qual Pred)
+data Inst = Inst CExpr [TyVar] (Qual Pred) (Maybe Id)
```
quark17: As the comment says, we should add comments about what the fields are -- I won't make you do the existing ones (if you don't know them), but can we at least add a comment for the new field?

src/Libraries/Base1/SShow.bs
```diff
@@ -1,6 +1,5 @@
 package SShow where
 
-import ListN
```
quark17: Is this the right resolution, or does it indicate that `SShow` is missing an instance for `ListN`?

src/Libraries/Base3-Misc/Makefile
```diff
@@ -7,7 +7,7 @@ include ../common.mk
 
 .PHONY: build
 build: $(BUILDDIR)
-	$(BSC) -u $(BSCFLAGS) Misc.bsv
+	$(BSC) -u $(BSCFLAGS) -suppress-warnings T0129:S0080 Misc.bsv
```
quark17: Ah, this suppression is needed because the import of `DefaultValue` is not removed, because it's needed for compilation.

src/comp/bsc.hs
```diff
@@ -367,10 +367,10 @@ compilePackage
 
     start flags DFimports
     -- Read imported signatures
-    (mimp@(CPackage _ _ imps _ _ _), binmap, hashmap)
+    (mimp@(CPackage _ _ imps impsigs _ _ _), binmap, hashmap)
         <- readImports errh flags binmap0 hashmap0 min
     when (hasDump flags DFimports) $
-      let imps' = [ppReadable s |  (CImpSign _ _ s) <- imps]
+      let imps' = [ppReadable s |  (CImpSign _ _ s) <- impsigs]
```
quark17: Just a minor question of whether you want to also change `imps'` to `impsigs'`.

quark17 (commented):
This is awesome, thank you!

I probably should have reviewed this commit by commit, instead of looking at the whole final diff.  So apologies if that affected the quality of my comments.  (It's also been a lot to get through.)

I still have not reviewed the testsuite changes, though -- I'll do that when I can -- so this is just reviewing the compiler and libraries, so far.

I see that you resolved warnings about `ListN` and `DefaultValue` by just removing the imports. That's probably a good approach -- leaving those libraries unchanged means that users will also get a warning about their imports of those libraries, which will help deprecate those files.  But another possibility would be, inside the library files, to re-export what used to be in them -- any opinion on whether to resolve that way instead?

To compile each of the Base3 library directories, we define a top-level file that imports everything in the directory and then the Makefile just needs to run BSC once on that, with the `-u` flag.  To avoid warnings you've inserted re-exports to those top files or added warning suppression on the BSC command line (or both).  I don't think anyone should be importing these top files -- my suggestion might even be to get rid of them!  Instead, we could build those directories in the same way that we build the Base1 and Base2 directories.

I haven't looked closely at the testsuite yey, but I think I saw that you resolved many failures by adding the warning to expected output -- any reason why you chose that rather than removing the unneeded imports?  I could see how leaving the imports causes these to be additional tests for the warning, but it also means lots of files that would have to change if the message ever changed.

I have not thought deeply about the implementation, but nothing jumped out at me.  By the way, do you know GHC does it, and whether there's anything to be learned from their experience?

I'm glad to see that typeclass instances are also tracked, so that files imported only for their instances don't lead to false warnings.  I look forward to having a close look at the tests.

Thank you again!

testsuite/bsc.driver/unused_imports/LocalExportWithImportedType.bsv
```diff
@@ -0,0 +1,15 @@
+// Test: Local export using imported type (Phase 2)
+// Expected: NO warning - Helper's Byte type is used in exported function signature
```
quark17: I'm not sure this is testing the `export`.  The type `Byte` is used in the local definition, so the import is justified.  Whether the def is exported or not won't change that -- removing the export should still see no warning?

Is a future step to check for unused definitions within a package?

testsuite/bsc.driver/unused_imports/unused_imports.exp
```diff
@@ -0,0 +1,73 @@
+# Tests for unused import warnings
+
+# Compile helper packages first
+compile_pass Helper.bsv
+compile_pass Helper2.bsv
+compile_pass Helper3.bsv
+compile_pass HelperAlias.bsv
+compile_pass_warning OrphanInstance.bsv T0127 1
```
quark17: Orphan instances are already checked in `bsc.typechecker/instances/orphan` (and the warning is expected also in tests in `bsc.typechecker/class_defaults/` and `bsc.typechecker/instances`).  I was worried there might be something redundant here, but I have convinced myself that's not the case.  Testing the import of a file for its orphan instance requires having a file with an orphan instance; and you check for an expected orphan-instance warning, to confirm that the test is set up correctly.  Maybe a comment explaining why you're checking the warning would help.

testsuite/bsc.driver/unused_imports/OrphanInstance.bs.bsc-out.expected
```diff
@@ -0,0 +1,3 @@
+Warning: "OrphanInstance.bs", line 9, column 10: (T0127)
```
quark17: This file is unused and should be deleted.

The message doesn't even match what BSC outputs -- maybe it was a potential change that was considered or a remnant from older BSC?

testsuite/bsc.driver/unused_imports/run_tests.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
quark17: Can we delete this file?  As I mentioned on another PR, you can run the tests in just this directory with `make localcheck` (or `make check` when you want to recurse into subdirs).

testsuite/bsc.driver/unused_imports/TypeclassConstraint.bsv
```diff
@@ -0,0 +1,13 @@
+// Test: Typeclass in constraint (Phase 1)
```
quark17: I was just a little surprised by "constraint".  We already have "context" and "proviso" (that tend to be used along BH/BSV lines unfortunately), I wasn't sure that we needed to add a further word?  No super strong objection, though.

testsuite/bsc.driver/unused_imports/TypeSynonymExpansion.bsv
```diff
@@ -0,0 +1,13 @@
+// Test: Type synonym expansion (Phase 1)
```
quark17: Just a minor thought: The "expansion" in the name and description was confusing (it implied more to me than is actually going on here), so could maybe be dropped -- the name `Byte` appears in the source, so it needs to be imported.

quark17 (commented):
Ok, I finished having a look over the testsuite.  There are a few deeper questions about the behavior, mixed in with many minor comments.

Under `bsc.interra/` and `bsc.lib/BRAM/`, you chose to update expected output to include the warning rather than remove the imports from the source (as you did in many other test directories) -- was there a reason for that?

src/Libraries/Base1/CShow.bs
```diff
@@ -1,6 +1,5 @@
 package CShow where
 
-import ListN
```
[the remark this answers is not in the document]
nanavati: CShow is implemented generically, so ListN will be handled via its generic representation as Vector.

src/Libraries/Base1/SShow.bs
nanavati: Like CShow, SShow is implemented generically, so ListN will be handled via its generic representation as Vector.

src/comp/GenSign.hs
nanavati: I'm not sure what I was thinking there. Deleted.

src/comp/ISyntaxCheck.hs
```diff
@@ -91,8 +91,8 @@ eqTypeFinal flags symt e t1 t2
           addExplPreds eqs
           vp <- mkVPredFromPred [] (IsIn numEqCls [t1', t2'])
           satisfy eqs [vp]
-    in  case (fst $ runTI flags False symt satisfyEq) of
-          Right ([],_) -> True
+    in  case runTI flags False symt satisfyEq of
```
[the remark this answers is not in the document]
nanavati: Fixed

src/Libraries/Base3-Contexts/Makefile
```diff
@@ -7,7 +7,7 @@ include ../common.mk
 
 .PHONY: build
 build: $(BUILDDIR)
-	$(BSC) -u $(BSCFLAGS) Contexts.bsv
+	$(BSC) -u $(BSCFLAGS) -suppress-warnings T0129:S0080 Contexts.bsv
```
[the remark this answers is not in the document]
quark17: Ah, I see that `Contexts` is actually different.  For `Misc`, the imported packages are empty and have been obsoleted.  But for `Contexts`, the imported files have contents but are not exported -- and this is because (presumably) the author thought that users might want to import `Contexts` in their code and would want that only to import `ModuleContext` and `ModuleCollect` but not the others.  I'm not sure people should be doing that -- shouldn't they just import the one they're using?  I kind of think we should (eventually) eliminate `Misc.bsv`, `Math.bsv`, and `Contexts.bsv` and use a makefile of some sort.

src/Libraries/Base3-Misc/Misc.bsv
```diff
@@ -8,7 +8,9 @@ import BuildVector::*;
 import BUtils::*;
 import BypassReg::*;
 import Cntrs::*;
-import DefaultValue ::*;
+// imported so that the .bo files are generated
+import DefaultValue::*;
+import FShow::*;
```
quark17: Why did you add an import of `FShow`?  This is in `Base1` and is being built there.

quark17:
I've reviewed your latest commits -- they look good, thank you!  If I understand, those were in response to my first review, and any commits in response to my second review (of the testsuite) are still to come.

src/comp/bsc.hs
nanavati: Done

src/comp/TypeCheck.hs
```diff
@@ -28,25 +28,31 @@ import SymTab
...
-  let checkDef d = (defErr, snd defTI)
-        where defTI  = runTI flags ai s $ tiOneDef d
-              defErr = case (fst defTI) of
+  let checkDef d = (defErr, warns, usedPkgs)
+        where (result, warns, usedPkgs) = runTI flags ai s $ tiOneDef d
+              defErr = case result of
                           (Left emsgs)  -> Left emsgs
                           (Right cdefn) -> rmFreeTypeVars cdefn
-  let (checks, wss) = unzip (map checkDef ds)
+  let checked = map checkDef ds
```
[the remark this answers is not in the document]
nanavati: No, it can be inlined.

src/Libraries/Base3-Misc/Misc.bsv
nanavati: We need something to force the building of the empty DefaultValue and FShow packages.

testsuite/bsc.driver/gensign/gensign.exp
```diff
@@ -163,9 +163,16 @@ compile_pass ImportReExportTestDataConFull.bsv
...
 # ---------------
 # Test that rexport of packages which reexport the same package are OK
+# ReExportSame_Top imports both P and Sub, exports both P::* and Sub::*,
+# where Sub only re-exports from P. Both imports are used via non-empty re-exports,
+# so neither warns, even though either P or Sub could be deleted individually.
 
 compile_pass ReExportSame_Top.bsv
 
+# Test non-empty package re-export (should not warn)
+compile_pass ReExportSame_TopGood.bsv
```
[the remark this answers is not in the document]
nanavati: I updated the comments to explain the difference.

testsuite/bsc.driver/gensign/gensign.exp
```diff
@@ -163,9 +163,16 @@ compile_pass ImportReExportTestDataConFull.bsv
 
 # ---------------
 # Test that rexport of packages which reexport the same package are OK
+# ReExportSame_Top imports both P and Sub, exports both P::* and Sub::*,
```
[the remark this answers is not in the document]
nanavati: Yes, arguably there's something to warn about because there is redundancy, but I'm not sure what to call the particular warning and I haven't thought through detecting it, so I'm leaving it alone for now.

testsuite/bsc.driver/unused_imports/LocalExportWithImportedType.bsv
nanavati: You're right it is the local use that ensures that there is no warning. But that local use might not be found if the synonym is expanded away too soon. The other thing I was after was making sure that exporting with an imported synonym works in general (I have a vague recollection of issues there and thought there might be some buried synonym expansion that makes things a little odd).

testsuite/bsc.driver/unused_imports/unused_imports.exp
nanavati: Added a comment about the test setup.

src/Libraries/Base3-Misc/Misc.bsv
quark17: I understand for `DefaultValue`, which is a source file in this directory.  My question is about `FShow`, which is not a source file in this directory; it is in `../Base1/` and is being built there, and should not need building or rebuilding here?

testsuite/bsc.driver/unused_imports/unused_imports.exp
```diff
@@ -0,0 +1,73 @@
...
+compile_pass_no_warning FieldAccess.bsv
+compile_pass_no_warning FieldAccessBS.bs
+
+# Test 6: Instance use - should NOT warn
+compile_pass_no_warning InstanceUse.bsv
+compile_pass_no_warning InstanceUseBS.bs
+
+# Test 6b: Orphan instance use - should NOT warn about OrphanInstance
+compile_pass_no_warning OrphanInstanceUse.bsv
+compile_pass_warning OrphanInstanceBS.bs T0127 1
```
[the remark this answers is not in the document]
nanavati: Yeah, the compare_file might be overkill but it feels harmless.

I've moved orphan instance compilation as you suggested.

testsuite/bsc.driver/unused_imports/OrphanInstance.bs.bsc-out.expected
nanavati: Deleted. Probably some in-development thing I forgot to get rid of.

testsuite/bsc.driver/unused_imports/run_tests.sh
nanavati: Right. Deleted.

testsuite/bsc.driver/unused_imports/test_where.bs
```diff
@@ -0,0 +1,8 @@
+package TestWhere where
```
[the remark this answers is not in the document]
nanavati: Deleted.

testsuite/bsc.driver/unused_imports/TransitiveSynonym.bsv
```diff
@@ -0,0 +1,14 @@
+// Test: Transitive synonym expansion (Phase 1)
+// Expected: NO warning for HelperAlias or Helper - both used via MyByte synonym chain
+
+package TransitiveSynonym;
+
+import HelperAlias::*;  // MyByte = Byte (from Helper)
+import Helper::*;       // Byte = Bit#(8)
```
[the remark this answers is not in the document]
nanavati: Yeah, there's a warning we miss here and now I'm correctly calling it out as a bug.

testsuite/bsc.driver/unused_imports/TypeclassConstraint.bsv
nanavati: Since you don't have a strong objection, I'll be lazy and leave it alone.

testsuite/bsc.driver/unused_imports/TypeSynonymExpansion.bsv
nanavati: I changed the description to be bit more specific about what is happening.

testsuite/bsc.driver/unused_imports/HelperAlias.bsv
```diff
@@ -0,0 +1,13 @@
...
+// This creates a synonym chain for testing
+
+package HelperAlias;
+
+import Helper::*;
+
+// Synonym that uses Helper's Byte synonym
+typedef Byte MyByte;
+
+export MyByte;
```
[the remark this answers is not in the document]
nanavati: I think it is an unnecessary import, so there is a warning we miss in the Transitive cases (which I now document).

testsuite/bsc.misc/deprecate/deprecate.exp
```diff
@@ -5,9 +5,11 @@ compile_pass DeprecatedLibrary.bs
 # Test that a use produces a message
 compile_pass_warning TestDeprecateMsg.bsv P0072
 
-# Test that no use (while imported) produces no warning
+# Test that no use (while imported) produces no deprecation warning
+# It does produce an unused import warning.
 compile_pass TestDeprecateNoMsg.bsv
-no_warnings [make_bsc_output_name TestDeprecateNoMsg.bsv]
+find_n_warning [make_bsc_output_name TestDeprecateNoMsg.bsv] P0072 0
```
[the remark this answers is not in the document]
nanavati: Now testing that the warning shows up and all warnings go away if we suppress it.

testsuite/bsc.driver/unused_imports/LocalExportWithImportedType.bsv
nanavati: And, yes, unused definitions are a possible future direction, but that's less interesting because most of the modules we write are "export everything".

testsuite/bsc.driver/unused_imports/UnusedImport.bs.bsc-out.expected
```diff
@@ -0,0 +1,2 @@
+Warning: "UnusedImport.bs", line 5, column 8: (T0129)
```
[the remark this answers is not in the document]
nanavati: Done

testsuite/bsc.driver/gensign/gensign.exp
```diff
@@ -163,9 +163,16 @@ compile_pass ImportReExportTestDataConFull.bsv
...
+# where Sub only re-exports from P. Both imports are used via non-empty re-exports,
```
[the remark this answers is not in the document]
nanavati: I just meant re-exports that actually contribute something to the final signature (rather than a re-export of an empty module, which is different). I clarified the language.

src/Libraries/Base3-Misc/Misc.bsv
nanavati: We need something to force a build of FShow somewhere? This seemed like a plausible location. Or are you suggesting actually mentioning FShow.bo explicitly in a Makefile?

nanavati:
Aside from the detail about how to make FShow build I *think* I've responded to everything. If there's anything I missed, please let me know.

src/Libraries/Base3-Misc/Misc.bsv
quark17: `FShow.bsv` is not in this directory!  It is already being built in the directory that it's in.  What makes you think it's not?

quark17: The directory `bsc/src/Libraries/Base1/` contains the following files with empty packages, that are all built when executing `make build` in that directory: FShow, Environment, RWire.  (And LevelFIFO merely re-exports FIFOLevel.)

src/comp/Error.hs
```diff
@@ -2809,6 +2810,10 @@ getErrorText (WOrphanInst inst) =
             " are defined in other packages. This can lead to confusing and inconsistent " ++
             " instance resolution if the orphan instance is not imported everywhere it could be used."))
 
+getErrorText (WUnusedImport pkg) =
```
quark17: Can we move this to be in numerical order (at the end of the Type errors)?

src/comp/TCheck.hs
```diff
@@ -3157,7 +3157,7 @@ writeableIfc :: Flags -> SymTab -> CType -> Bool
 writeableIfc flags r t
     | Just _ <- isWriteType r t = True
     -- incoherent matches are resolved *after* reducePred
-    | Right (Just _) <- fst $ runTI flags False r checkWriteable = True
+    | (Right (Just _), _, _) <- runTI flags False r checkWriteable = True
```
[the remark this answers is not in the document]
quark17: In the three places where I commented that fst3 was imported but not used, I assumed you would use it, but instead you resolved by removing the import.  Which I guess is OK.

quark17 (commented):
Looks good.  I'd just like the error moved to be in sorted order in `Error.hs` and I'd like `FShow` removed from `Misc.bsv` (and I think the comment about the `.bo` could go too, but won't push).  If those a fixed, I'll merge.  I could do them, but wanted to check with you.

quark17:
I've rebased and added a commit that resolves my comments.  Following your suggestion, I added tests for FShow and DefaultValue, that you can import the empty package (for backwards compatibility) but get a warning that it is unnecessary.  (Someday we can decide to remove that support?  Also, I did not add tests for the older packages.)  When the checks pass, I'll merge it -- then will rebase and merge other PRs, although that may be a serial process, since several of them change the .bo format, so will have to rebase each and wait for them to pass the checks, before merging and moving to the next one.

quark17 (approved):

PR #848, opened 2025-11-21
Don't use letrec where it isn't necessary

nanavati:
I'm currently working on code to lift monomorphic dictionary bindings to the top level. That lifting is easier to do when the dictionaries are created with non-recursive bindings. To that end, I've taught the typechecker to track recursive and non-recursive dictionary bindings separately so that it can do better code generation (and only generate letrec when there is true recursion).

Ironically, the answer to "when there is true recursion" appears to be never (unless I've missed something). The trick is that ctxreduce runs before typecheck and when ctxreduce encounters mutually recursive sets of instances, it strength-reduces them to their necessary external inputs (e.g. Just Eq a for Eq (List a), not Eq (List a) too). Any recursion is resolved by using the top-level instance / instance-producing definitions (which are already mutually recursive because they're at the top level). So with my new code, ctxreduce runs and sees dictionary recursion (both self-recursion and mutual recursion). But since all ctxreduce is doing is reducing contexts, not generating bindings, those recursive bindings are thrown away. When typecheck runs, it sees apparently non-recursive dictionaries and generates code according.

I've also replaced letrec with letseq in some random places where we weren't actually using recursion.

### DIFF
--- a/src/comp/CSyntax.hs
+++ b/src/comp/CSyntax.hs
@@ -689,6 +689,7 @@ cTApply f ts = CTApply f ts
 
 cLetSeq :: [CDefl] -> CExpr -> CExpr
 cLetSeq [] e = e
+cLetSeq ds1 (Cletseq ds2 e) = Cletseq (ds1 ++ ds2) e
 cLetSeq ds e = Cletseq ds e
 
 cLetRec :: [CDefl] -> CExpr -> CExpr
--- a/src/comp/Parser/BSV/CVParserImperative.lhs
+++ b/src/comp/Parser/BSV/CVParserImperative.lhs
@@ -14,7 +14,7 @@
 > import Parser.BSV.CVParserCommon
 > import Parser.BSV.CVParserAssertion
 > import Parser.BSV.CVParserUtil
-> import CSyntax hiding (cLetSeq, cLetRec)
+> import CSyntax
 > import CSyntaxUtil
 
 > import Id
@@ -3038,15 +3038,6 @@ Ensures that variables are declared, etc.
 
 ==============
 END ASSERTIONS
-
-> cLetRec :: LetFn CExpr
-> cLetRec [] body = body
-> cLetRec defs body = Cletrec defs body
-
-> cLetSeq :: LetFn CExpr
-> cLetSeq [] body = body
-> cLetSeq defs (Cletseq defs' body) = Cletseq (defs ++ defs') body
-> cLetSeq defs body = Cletseq defs body
 
 detect whether assignment is to the fake "uninitialized" value
 


src/comp/TCMisc.hs
```diff
@@ -477,27 +603,27 @@ reducePredsAggressive' dvs es bs1 s1 vps1 = do
...
     vps2' <- concatMapM (expTConPred . expandSynVPred) vps2
-    reducePredsAggressive' dvs es (bs1 ++ bs2) s2 vps2'
+    reducePredsAggressive' dvs es (sbs2 <++ sbs1) s2 vps2'
    else do
     -- satMany is inside a loop, so it doesn't consistently apply
     -- its accumulated substitution to the reduced predicates.
     -- Apply the substitution here to clean that up before returning
     -- to the external caller.
-    return (apSub s2 vps2, map mkDefl (bs1 ++ bs2), s2)
+    return (apSub s2 vps2, apSub s2 (sbs2 <++ sbs1), s2)
```
quark17: Is this adding an `apSub` that didn't used to be there?  If so, is there a test case for what that fixes?

src/comp/TCMisc.hs
```diff
@@ -48,6 +57,116 @@ import PreStrings(sAcute)
 import IOUtil(progArgs)
 import Debug.Trace
 
+-------
```
quark17: Any thoughts on whether to put this section in its own file?  I don't have strong feelings.

nanavati:
> Sounds good. Thank you!
> 
> Instead of promoting transitively, could you in theory have three levels -- seq on top of rec on top of seq?

I think it is possible, but not simple. Cases of mutual recursion (A depends on B, B depends on A) first show up as bindings that were assumed to be non-recursive actually depending on a recursive binding. So I don't think you can conclude that it is safe to put a promoted binding in an inner binding group unless you do more analysis.

The analysis that occurs to me is tsorting the recursive binding group to figure out the true cycles vs things that depend on the cycles. Then you could tsort the only-dependent "recursive" bindings and put them in a letseq. I could see that helping minimize letrec more, but since recursive binding groups are rare (only @krame505 seems to be able to create them) I don't think the improvement is worth the effort (especially the effort of finding a testcase where that makes a difference since constructing testcases to generate recursive binding groups is not simple to begin with).

src/comp/TCMisc.hs
nanavati: Oh, that's a good point. We really don't want TCMisc breaking the abstraction. I'll carve it out.

nanavati: Carved out in my latest update.

src/comp/CSyntax.hs
```diff
@@ -685,6 +687,14 @@ cTApply :: CExpr -> [CType] -> CExpr
 cTApply f [] = f
 cTApply f ts = CTApply f ts
 
+cLetSeq :: [CDefl] -> CExpr -> CExpr
+cLetSeq [] e = e
+cLetSeq ds e = Cletseq ds e
```
quark17: Is it worth adding an arm here for combining nested `letseq`?
```
cLetSeq ds1 (Cletseq ds2 e) = Cletseq (ds1 ++ ds2) e
```
I notice that `CVParserImperative.lhs` defines its local version of `cLetSeq` this way.

src/comp/Parser/BSV/CVParserImperative.lhs
```diff
@@ -14,7 +14,7 @@
 > import Parser.BSV.CVParserCommon
 > import Parser.BSV.CVParserAssertion
 > import Parser.BSV.CVParserUtil
-> import CSyntax
+> import CSyntax hiding (cLetSeq, cLetRec)
```
quark17: This file defines its own `cLetSeq` and `cLetRec` so you've hidden the ones from `CSyntax`.  However, the local definitions are nearly identical and could probably be removed (if you update the `CSyntax` version to merge nested `letseq`).  The local versions are defined using the type alias `LetFn`, but I believe that it elaborates to the same thing.  Did you have an opinion?  I'm also OK not holding things up over a minor cleanup (particularly if it's debatable whether it's cleaner).

src/comp/SolvedBinds.hs
```diff
@@ -0,0 +1,134 @@
+-- | SolvedBinds are sets of solved dictionary bindings.
```
quark17: My only thought, looking at it now, was whether the name of the file should be something like `TCSolvedBinds.hs` or `CSolvedBinds.hs`, indicating its connection to typecheck or CSyntax.  But we don't do that with files like `Scheme` and `Pred` and `Subst`, so I think this is fine.  And anyway, perhaps eventually, hierarchy can be used to indicate which subsystem a file is part of, rather than its name.

quark17 (commented):
Almost there, thanks!  The only question blocking merging is whether you want to add an arm to `cLetSeq` for merging nested letseq.  Otherwise, just a couple minor thoughts.

src/comp/CSyntax.hs
nanavati: Yes, letting us remove the definitions in CVParserImperative is worth the change. Done.

PR #849, opened 2025-11-25
Port splitting part 2: Support methods with multiple output ports

krame505:
This is nowhere close to finished, but I've started working on adding support for methods with multiple output ports, as an extension of my previous work on input port splitting in #729[^1].  The high-level idea is to make the port-splitting type classes allow the creation of wrapper methods that return tuples of `Bit` values, make `AConv` understand `Bit` tuples as synthesizable, and plumb this all the way through the rest of the compiler.  `AIDef` and `AIActionValue` are changed to contain multiple `ADef`s, and `AMethCall` and `AMethValue` now contain an index of which output port's value is being accessed from the method.

I've gotten quite a ways into implementing this - `bsc` mostly compiles except for `AState.hs`, which I am still trying to make sense of. The Bluesim, lambda calculus and SAL (what is that?) back ends are currently stubbed out and give an error when encountering `AIDef`/`AIActionValue` with multiple `ADef`s.  I think we only really care about supporting multiple output ports for the Verilog and Bluesim backends, the others may be left as future work.  

One question I have for @quark17 is what is the most straightforward path to make Bluesim support methods with multiple output ports - should we generate structs for these methods' sim functions to return, containing the output port values?  Or just return all the port values as a big bit vector and do the appropriate slicing?  

This is still a draft with much debugging to do, but I would appreciate any early feedback.  

[^1]: I'm not sure how to make that branch show up as the base of this pull request, since the port splitting branch is on my fork of the bsc repo, so all those changes show up here as well.  You can see the diff with only my new output port splitting changes [here](https://github.com/krame505/bsc/compare/genwrap...MatX-inc:bsc:split-output-ports).


krame505:
Note that once the type functions PR (#888) gets merged, we should look at how type functions can help clean things up here.  `TupleSize` and `AppendTuple` should at least have type functions (`TupleSizeOf` and `TAppendTuple`?)

krame505:
I have rebased this PR on upstream, and updated the expected results accordingly.  

Note that when this pull request is stacked on top of https://github.com/B-Lang-org/bsc/pull/939, the changes to internal signal names that appear in the lambda calculus tests no longer appear, meaning https://github.com/B-Lang-org/bsc/pull/871 is no longer needed.  

I intend to do another round of cleanup here to try to avoid excessive use of tuple append, and there might also be some nice use cases for type functions.

src/comp/AExpr2STP.hs
```diff
@@ -583,15 +584,35 @@ convAExpr2SExpr mty (AMethCall ty@(ATBit width) modId methId args) = do
     -- get the actual port name, so that methods which share the same output port
     -- will appear logically equivalent
     smap <- gets stateMap
-    let portId = getMethodOutputPort smap modId methId
-        e = (AMethCall ty modId portId args)
+    let e = case getMethodOutputPorts smap modId methId of
+              [portId] -> AMethCall ty modId portId args
```
nanavati: Is this a soft requirement that methods have only one output? If not, how is that avoided?

src/comp/ASyntax.hs
```diff
@@ -838,25 +837,18 @@ aIfaceProps (AIAction      { aif_props = p }) = p
...
-aIfaceResSize (AIActionValue {aif_value = (ADef _ (ATBit n) _ _) }) = n
-aIfaceResSize x = internalError ("aIfaceResSize: " ++ show x)
-
-aIfaceResType :: AIFace -> AType
+aIfaceResTypes :: AIFace -> [AType]
 -- XXX should be ATAction?
-aIfaceResType (AIAction { }) = ATBit 0
-aIfaceResType (AIDef { aif_value = (ADef _ t _ _)}) = t
-aIfaceResType (AIActionValue { aif_value = (ADef _ t _ _)}) = t
+aIfaceResTypes (AIAction { }) = [ATBit 0]
```
quark17: Does this need to return a zero-size element?  Can it return an empty list?

src/comp/ASyntax.hs
```diff
@@ -1560,9 +1573,14 @@ instance PPrint AExpr where
     pPrint d p (ATaskValue _ i _ _ n) = pparen (p>0) $ pPrint d 1 i <> text ("#" ++ itos(n))
     pPrint d p (AMethCall _ i m es) =
         pparen (p>0 && not (null es)) $
-        pPrint d 1 i <> sep (text "." <> ppMethId d m : map (pPrint d 1) es)
+        pPrint d 1 i <>
+        sep (text "." <> ppMethId d m : map (pPrint d 1) es)
```
quark17: Looks like this is just a whitespace change -- that's ok, or feel free to remove it

src/comp/ASyntax.hs
```diff
@@ -1560,9 +1573,14 @@ instance PPrint AExpr where
...
     pPrint d p (AMethValue _ i m) =
         pparen (p>0) $ pPrint d 1 i <> text "." <> ppMethId d m
+    pPrint d p (ATuple _ es) =
+        pparen (p>0) $ parens (commaSep (map (pPrint d 0) es))
```
quark17: Is there a reason for the outer `pparen`?  I think it would result is extra parens around the tuple.

src/comp/ASyntax.hs
```diff
@@ -1560,9 +1573,14 @@ instance PPrint AExpr where
...
+    pPrint d p (ATupleSel _ e idx) =
+        pparen (p>0) $ pPrint d 0 e <> text "[" <> pPrint d 0 idx <> text "]"
```
quark17: The printing of `e` should have parens around it, if it's a complex expression (which is done by passing a non-zero value for the second argument, I think).  And the outer `pparen` is unnecessary (if we consider the bracket indexing to bind tightly).

src/comp/ASyntax.hs
```diff
@@ -1854,7 +1874,12 @@ instance PPrintExpand AExpr where
                <> if (null es) then empty else (parens (hsep ( punctuate comma docArgs )) )
                    where
                    docArgs = map (pPrintExpand m d defContext) es
-    pPrintExpand m d ec (AMethValue _ i meth) = pPrint d 1 i <> text "." <> ppMethId d meth
+    pPrintExpand m d ec (AMethValue _ i meth) =
```
quark17: The `AMethValue` change is whitespace only, but that's OK if it's intentional.

The `pPrintExpand` for `ATuple` and `ATupleSel` have the same issues as mentioned above for `pPrint`.

src/comp/AVerilogUtil.hs
```diff
@@ -886,9 +886,10 @@ vState  flags rewire_map avinst =
         inps =  [ (mkVId (portid s ino),
                    mkArgId m k ino,
                    vSize argType)
-                  | (meth@(Method m _ _ mult ps mo me),
+                  | (meth@(Method m _ _ mult ps outs me),
                      (argTypes,_,_))
                         <- zip (vFields vi) mts,
+                        -- (VName s, vps)
```
quark17: What's this line? Should it be deleted?

src/comp/BinData.hs
```diff
@@ -999,6 +999,7 @@ instance Bin AType where
     writeBytes (ATReal)           = do putI 2;
     writeBytes (ATArray sz t)     = do putI 3; toBin sz; toBin t
     writeBytes (ATAbstract i szs) = do putI 4; toBin i; toBin szs
+    writeBytes (ATTuple ts)       = do putI 5; toBin ts
```
quark17, line 1002: These changes will require bumping the header in GenBin and GenABin, I assume?

testsuite/bsc.bluetcl/commands/module.tcl.bluetcl-bh-out.expected
```diff
@@ -115,19 +115,19 @@ ios1_1 {Prelude.Inout Test.Bar}
 ----------
 module ports mkT
 Command: module ports mkT
-interface {{interface r {{method _write r__write {clock default_clock} {reset no_reset} {args {{{name r__write_1} {port r__write_1} {size 1}}}} {enable EN_r__write} {ready RDY_r__write}} {method _read r__read {clock no_clock} {reset no_reset} {args {}} {result r__read} {ready RDY_r__read}}}} {inout b b {port b} {clock default_clock} {reset default_reset}}}
+interface {{interface r {{method _write r__write {clock default_clock} {reset no_reset} {args {{{name r__write_1} {port r__write_1} {size 1}}}} {results {}} {enable EN_r__write} {ready RDY_r__write}} {method _read r__read {clock no_clock} {reset no_reset} {args {}} {results {{{port r__read} {size 1}}}} {ready RDY_r__read}}}} {inout b b {port b} {clock default_clock} {reset default_reset}}}
```
quark17, line 118: Hm, the BDW tests pass, so I guess maybe BDW doesn't use this command? or else the tests are not sufficient.

But it does appear to be a backwards incompatible change in Bluetcl output that will need to be prominently mentioned in the release notes.

Also, the User Guide doc section for Bluetcl commands will need to be updated.

src/comp/ASyntax.hs
```diff
@@ -904,7 +896,7 @@ addRdyToARule rdyId r0@(ARule { arule_id = ri, arule_pred = e }) = (d, r)
 aIfaceSchedNames :: AIFace -> [ARuleId]
 aIfaceSchedNames (AIAction { aif_body = rs}) = map arule_id rs
 aIfaceSchedNames (AIActionValue { aif_body = rs}) = map arule_id rs
-aIfaceSchedNames (AIDef { aif_value = d }) = [adef_objid d]
+aIfaceSchedNames (AIDef { aif_name = i }) = [i]
```
quark17, line 901: I'm a little concerned about this change.  I wouldn't be surprised if there were differences in the Ids (such as properties or position).  But maybe there are changes elsewhere that fix or account for this -- I'll keep an eye out.

quark17:
I notice that there are failures in the testsuite.  You mention the ones in SAL and LambdaCalc, but there are also others?
```
./bsc.interra/messages/EResources/testrun.sum: Running ./EResources.exp ...
Error FAIL: `EResources.bs.bsc-vcomp-out.filtered' differs from `EResources.bs.bsc-vcomp-out.expected.filtered'

./bsc.misc/lambda_calculus/testrun.sum: Running ./lambda_calculus.exp ...
Error FAIL: `lc-sysTb.out.filtered' differs from `lc-sysTb.out.expected.filtered'

./bsc.misc/sal/testrun.sum: Running ./sal.exp ...
Error FAIL: `CTX_sysTb.sal.filtered' differs from `CTX_sysTb.sal.expected.filtered'

./bsc.scheduler/avmeth/testrun.sum: Running ./avmeth.exp ...
Error FAIL: `AVArgUse_C.bsv.bsc-sched-out.filtered' differs from `AVArgUse_C.bsv.bsc-sched-out.expected.filtered'
Error FAIL: `AVArgUse_SBR.bsv.bsc-sched-out.filtered' differs from `AVArgUse_SBR.bsv.bsc-sched-out.expected.filtered'

./bsc.scheduler/testrun.sum: Running ./scheduler.exp ...
Error FAIL: `SplitIf.bsv.bsc-sched-out.filtered' differs from `SplitIf.bsv.bsc-sched-out.expected.filtered'
Error FAIL: `SplitIf2.bsv.bsc-sched-out.filtered' differs from `SplitIf2.bsv.bsc-sched-out.expected.filtered'

File: src/comp/AExpr2STP.hs (hunk 1 above)
Comment (quark17) [reply src/comp/AExpr2STP.hs]: I had this question, too, but see the other arms below:  This arm is only encountered is there's no `ATupleSel` applied to the method, and so we expect it to only have one output.

File: src/comp/ASchedule.hs [hunk 11]
```diff
@@ -4137,13 +4137,13 @@ cvtIfc (AIActionValue _ _ ifPred ifId ifRs (ADef dId t _ _) _) =
     -- call will be in the same Rule structure as the action part)
     -- (note that, if the method body is not split into multiple
     -- rule, dId and rId will be the same)
-    [(Rule rId rOrig [ifPred, rPred] [ifPred, rPred, dExpr] rActs)
+    [(Rule rId rOrig [ifPred, rPred] ([ifPred, rPred, dExpr]) rActs)
```
quark17: Why are parens added around a list?

src/comp/ACheck.hs
```diff
@@ -98,7 +98,6 @@ chkAIface aa@(AIAction { aif_body = rs }) =
 
 chkAIface aa@(AIActionValue { aif_value = d, aif_body = rs }) =
     tracePP "chkAIface AIActionValue" aa $ (all chkARule rs) && (chkADef d)
-
```
quark17, line 101: This blank line should be reinstated

src/comp/AExpand.hs
```diff
@@ -537,6 +539,10 @@ getExprSize (APrim _ _ _ es) = (nub $ concat vars, sum terms, 1 + maximum depths
 
 getExprSize (AMethCall t i mid args) = ([mid],1,1)
 getExprSize (AMethValue t i mid)     = ([mid],1,1)
+getExprSize (ATupleSel t e i)        = (vars, terms + 1, depth + 1)
```
quark17: I'm not convinced that these two arms should increment the terms/depth, since they don't really have any logic.  But also, this is only used to decide what to inline vs lift, and it may not matter here, since I assume they are only applied at port boundaries.

src/comp/CType.hs
```diff
@@ -457,14 +457,20 @@ getActionValueArg t = internalError ("getActionValueArg: " ++ ppReadable t)
 -- These are used during foreign function processing to determine if arguments
 -- and return values are polymorphic or of a known size.
 isTypePolyBit :: Type -> Bool
+isTypePolyBit (TAp (TCon (TyCon i _ _)) (TAp (TCon (TyCon i' _ _)) arg))
```
quark17, line 460: This is a new case arm -- was it not needed before, or did you find a bug?  is there a testcase that's associated with this change?

src/comp/IExpandUtils.hs
```diff
@@ -362,19 +362,24 @@ isPrimType (ITCon i _ _) = i == idPrimAction ||
...
 -- Primitive arrays
 isPrimType (ITAp (ITCon i _ _) elem_ty) | i == idPrimArray = isPrimType elem_ty
+-- Tuples of bits
+isPrimType t | isBitTupleType t = True
 isPrimType _ = False
 
 -- Primitive type applications
 isPrimTAp :: IType -> Bool
 isPrimTAp (ITCon _ _ (TIstruct SInterface{} _)) = True
-isPrimTAp (ITCon i _ _) = i == idActionValue_ ||
```
quark17, line 376: Ah, because `isPrimTAp` is only called inside `isPrimType`, the removal of `idActionValue_` is OK because it gets added to `isPrimType`.

src/comp/SignalNaming.hs
```diff
@@ -67,6 +68,12 @@ signalNameFromAExpr' (expr@AMethCall { }) =
     connectWith "_" (map signalNameFromAExpr' (ae_args expr))
 signalNameFromAExpr' (expr@AMethValue { }) =
     ppString (ae_objid expr) ++ "_" ++ ppString (unQualId (ameth_id expr))
+signalNameFromAExpr' (expr@ATuple { }) =
+    "TUPLE_" ++
+    connectWith "_" (map signalNameFromAExpr' (ae_elems expr))
+signalNameFromAExpr' (expr@ATupleSel { }) =
+    signalNameFromAExpr' (ae_exp expr) ++
+    "_" ++ itos (ae_index expr)
```
quark17: I'd be tempted to insert `SEL_` (or `TPL_`) before the number, otherwise it might just look like part of the expression's name.

src/comp/VIOProps.hs
```diff
@@ -151,9 +151,12 @@ getIOProps flags ppp@(ASPackage _ _ _ os is ios vs _ ds io_ds fs _ _ _) =
                          -- create the method name map
                          let nmap = M.fromList $
                                     createVerilogNameMapForAVInst flags v,
-                         -- for each method that has an output port
-                         vfi@(Method { vf_output = Just (vname,pprops) })
-                             <- vFields (avi_vmi v),
+                         -- for each method (not clocks or resets)
+                         vfi@(Method {}) <- vFields (avi_vmi v),
+                         -- for each method output port
+                         (methpart, (vname, pprops))
```
quark17: Might want to use a different name than `methpart`, since elsewhere we talk about EN, RDY, args, and result as being the parts of methods -- use a name that indicates it's an output?

testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
```diff
@@ -109,43 +110,39 @@ module mkDesign(clk,
...
-  assign _0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_result_a_BITS_ETC___d57 =
+  assign _0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_result_a_BITS_ETC___d27 =
 	     x__h336 + 6'd1 ;
-  assign _0_CONCAT_IF_0_CONCAT_result_a_BITS_8_TO_4_PLUS_ETC___d58 =
+  assign _0_CONCAT_IF_0_CONCAT_result_a_BITS_8_TO_4_PLUS_ETC___d18 =
 	     x__h432 + 6'd1 ;
-  assign _0_CONCAT_result_a_BITS_8_TO_4_PLUS_0_CONCAT_IN_ETC___d59 =
+  assign _0_CONCAT_result_a_BITS_8_TO_4_PLUS_0_CONCAT_IN_ETC___d9 =
 	     x__h528 + 6'd1 ;
-  assign _theResult_____1_snd__h939 =
```
quark17: The only real change in this file is that this def got inlined.  Do we know what caused that difference?

testsuite/bsc.lib/Prelude/TupleSize.bs
```diff
@@ -3,8 +3,8 @@ package TupleSize where
 -- Test the TupleSize type class by using it in proviso constraints
 
 -- Helper function that requires TupleSize constraint
-tupleSize :: (TupleSize t n) => t -> Integer
```
quark17, line 6: This still needs to have the proviso on it -- the fact that BSC doesn't complain was the subject of another PR that's pending, I believe

testsuite/bsc.misc/lambda_calculus/lc-sysDynArrSelWithImplCond.out.expected
```diff
@@ -31,21 +31,13 @@ dim_sysDynArrSelWithImplCond =
...
-     let { (def__element_first__h710 :: Bit #8) =
-	       meth_first_FIFO2 (inst_fs_3__sysDynArrSelWithImplCond state0)
-	 ; (def__element_first__h704 :: Bit #8) =
-	       meth_first_FIFO2 (inst_fs_2__sysDynArrSelWithImplCond state0)
-	 ; (def__element_first__h698 :: Bit #8) =
-	       meth_first_FIFO2 (inst_fs_1__sysDynArrSelWithImplCond state0)
-	 ; (def__element_first__h692 :: Bit #8) =
-	       meth_first_FIFO2 (inst_fs_0__sysDynArrSelWithImplCond state0)
-	 ; (def_x__h558 :: Bit #2) = meth_read_RegN (inst_idx__sysDynArrSelWithImplCond state0)
+     let { (def_x__h558 :: Bit #2) = meth_read_RegN (inst_idx__sysDynArrSelWithImplCond state0)
```
quark17: You mention that the LC/SAL changes go away when this PR is stacked on 939 -- but those commits are not currently part of the PR?

testsuite/bsc.scheduler/SplitIf.bsv.bsc-sched-out.expected
```diff
@@ -70,7 +70,7 @@ order: [RL_r, RL_toggleDir, RL_done, RL_incr]
  (dir.read, [(dir.read, 1)]),
  (dir.write, [(dir.write NOT_dir___d2, 1)]),
  (test.RDY_xfer, [(test.RDY_xfer, 1)]),
- (test.xfer, [(test.xfer dir__h205, 1)])]
+ (test.xfer, [(test.xfer dir___d1, 1)])]
```
quark17, line 73: It's interesting that this changed from `__h` to `___d` -- do we know why that is?

testsuite/bsc.verilog/inline/inline.exp
```diff
@@ -4,8 +4,8 @@
 # there is only one use.  (Inlining for optimization is OK.)
 if {$vtest == 1} {
     compile_verilog_pass RWireOneUse.bsv "" -keep-inlined-boundaries
-    find_n_strings sysRWireOneUse.v {assign rw$wget = } 1
-    find_n_strings sysRWireOneUse.v {assign rw$whas = } 1
+    find_n_strings sysRWireOneUse.v {assign rw$wget_RES_1 = } 1
```
quark17: Is it possible for non-split outputs to not have `RES_1` added to their def names?

src/comp/AConv.hs
```diff
@@ -684,17 +730,25 @@ aTypeConvE a (ITAp (ITCon r _ _) elem_ty) | r == idPrimArray =
...
     (ICon _ (ICString _ s)) -> ATString (Just (genericLength s))
     otherwise               -> ATString Nothing
+aTypeConvE a (ITCon i _ _) | i == idPrimUnit = ATBit 0
 aTypeConvE a t = abs t []
   where abs (ITCon i _ _) ns = ATAbstract i (reverse ns)
         abs (ITAp t _) ns = abs t ns
         abs _ _ = -- ATAbstract idBit []        -- XXX what's this
                   internalError ("aTypeConvE|" ++ show t)
 
+aTypesConv :: Id -> IType -> [AType]
```
quark17: I wonder if this should be renamed to reflect that it's converting/extraction a tuple type

src/comp/APaths.hs
```diff
@@ -1219,9 +1241,31 @@ findEdges env (AMethCall t i qmi exprs) =
         (edges, ms) = concatUnzip (map f (zip [1..] exprs))
         meth_arg_mux = PNStateMethodArgMux i mi
         muxes = if null exprs then ms else meth_arg_mux:ms
-    in  ([PNStateMethodRes i mi], edges, muxes)
+    in  ([PNStateMethodRes i mi 1], edges, muxes)
 findEdges env (AMethValue t i qmi) =
-    ([PNStateMethodRes i (unQualId qmi)], [], [])
+    ([PNStateMethodRes i (unQualId qmi) 1], [], [])
+findEdges env (ATupleSel _ (AMethCall t i qmi exprs) oi) =
```
quark17, line 1247: Do we need new test cases for path analysis through split inputs/output?

src/comp/IExpand.hs
```diff
@@ -302,7 +302,10 @@ iExpand errh flags symt alldefs is_noinlined_func pps def@(IDef mi _ _ _) = do
               t = iGetTypeNorm norm e
               -- the name for the new IDef being created
               i = case expr_name of
-                    Just name ->
+                    -- Heuristic: treat names starting with _ as bad names,
+                    -- as these may come from the arguments of functions like
+                    -- id _x = _x
+                    Just name | not (isEmptyId name) && head (getIdBaseString name) /= '_' ->
```
quark17: This is PR 871 -- I thought you said that's no longer needed?

src/comp/IExpand.hs
```diff
@@ -4880,7 +4897,7 @@ isCanon (ICon _ (ICModParam { })) = True
 isCanon (ICon _ (ICClock { })) = True
 --isCanon (IAps (ICon _ (ICPrim _ PrimBlock)) _ _) = True                -- XXX is this the best way?
 isCanon (IAps (ICon _ (ICSel { })) _ [_]) = True
-isCanon (IAps (ICon _ (ICOut { })) _ [_]) = True
+--isCanon (IAps (ICon _ (ICOut { })) _ [_]) = True
```
quark17: What happened here?  Is it because we don't want to treat tuple creation as canon?  (Does this check need to be more precise, then, or does it not matter?)

src/comp/IfcBetterInfo.hs
```diff
@@ -93,10 +90,9 @@ fieldInfoToBetterInfo :: Flags -> SymTab -> (Id,Maybe FieldInfo) -> BetterInfo
...
 fieldInfoToBetterInfo flags symTab (fieldId, Just fi) =
     BetterMethodInfo {mi_id = fieldId,
-                      mi_result = maybe (id_to_vPort fieldId) (str_to_vPort) mres,
                       mi_ready  = maybe (id_to_vPort $ mkRdyId fieldId) str_to_vPort mrdy,
                       mi_enable = maybe (id_to_vPort $ mkEnableId fieldId) str_to_vPort  men,
                       mi_prefix = maybe fieldId (setIdBaseString fieldId) mprefix
                }
     where prags   = fi_pragmas fi
-          (mprefix,mres,mrdy,men,_,_,_) = getMethodPragmaInfo prags
+          (mprefix,_,mrdy,men,_,_,_) = getMethodPragmaInfo prags
```
quark17, line 98: I assume that interface pragmas for renaming the output port are still supported (since I don't see a change to the testsuite for that) -- but where does that happen, then, if not here?

src/comp/ISplitIf.hs
```diff
@@ -445,7 +445,7 @@ iSplitIface flags ieface@(IEFace i xargs (Just (e,t)) Nothing wp fi)
             in (smap, IEFace i xargs Nothing (Just irules_opt) wp fi)
       else case e of
             (IAps (ICon av (ICTuple {fieldIds = [_val_id,_act_id]}))
-                      [ITNum _] [val_,act_])
```
quark17, line 448: Do we need tests for if-splitting of interfaces with multiple output ports?  (though I don't know if anyone uses that feature)

src/comp/ISyntaxUtil.hs
```diff
@@ -176,18 +174,34 @@ itList, itMaybe :: IType -> IType
...
+isBitTupleType (ITAp (ITAp (ITCon i _ _) t1) t2) | i == idPrimPair =
+  isBitType t1 && isBitTupleType t2
+isBitTupleType t = isBitType t
+
 -- extension point for ActionValue methods
 isActionType :: IType -> Bool
 isActionType x = (x == itAction) || (isitActionValue_ x) || (isitAction x)
 
 -- extension point for ActionValue methods
 isValueType :: IType -> Bool
```
quark17: Nothing calls `isValueType` now -- you can remove it.

(Otherwise, do you need to add something to check for zero-size output?)

src/comp/TCheck.hs
```diff
@@ -710,43 +710,42 @@ tiExpr as td exp@(CmoduleVerilog name ui clks rsts args fields sch ps) = do
...
                         let final_port = lastOrErr "chkResType" ps
                         -- XXX kill PrimAction once imports in Prelude are converted over
                         if (isActionWithoutValue t) || (isPrimAction t)
-                         then return (inputs, Just final_port, Nothing)
-                         else if (isBit t)
-                               then return (inputs, Nothing, Just final_port)
-                               else errBadResType t
+                        then return (inputs, Just final_port, [])
+                        else if (isBitTuple t)
+                        -- XXX should have multiple output ports for bit tuples here?
```
quark17: What is this comment about?

src/comp/SimMakeCBlocks.hs
```diff
@@ -584,9 +595,13 @@ cvtIFace modId pps def_map meth_map method_order_map reset_list m =
                    -- ready is off (the user lied about it being always_en'd),
                    -- but, at that point, all bets are off anyway
                    check_rdy ss' ++ ret_stmts
-             Nothing -> check_rdy $
+             [] -> check_rdy $
                         cvtActions modId name
                             def_map method_order_map S.empty body rst_ids
+             -- TODO: Support methods with multiple return values
```
quark17: I'm confused by the internalError here and above -- Bluesim seems to work for split output ports, so where does that happen, if not here?

quark17 (commented):
I've had a look over all the src changes and most of the testsuite.  It seems like you left in the IExpand heuristic for underscore names, and I haven't carefully audited all the name changes due to that -- I wasn't sure if that's intentional.

Can you say a little something about the change to `ActionValue_`?  The argument used to be a numeric type, but now it's either `Bit n` or `Tuple`.  Can you saw what was involved in that change?

Can you also say a little about the change, for AIFace, to use the `aif_name` ID rather than the ID inside the `aif_value` Def.  (I think this is for both `AIDef` and `AIActionValue`.)

If I understand, Bluesim works with split output ports ... by having Bluesim just convert the tuple back to one concatenated bit vector.  So Bluesim isn't going to look any different than before -- it won't show the separate names in a VCD dump, for example.  (So there's potentially room in the future for better support in Bluesim.)

I noted a few places where I think we have more tests.

src/comp/IExpand.hs
krame505: Whoops, that should not have been included.

krame505:
I had forgotten that the heuristic from #871 was already included in this branch, I removed it.  

The testsuite failures were because I had not re-generated the expected files after reverting the `_` names in https://github.com/B-Lang-org/bsc/pull/849/changes/d08529873950bfb20cda7779104b244a71497a0a.  I'm going to go ahead and rebase this on https://github.com/B-Lang-org/bsc/pull/939 - can that PR be merged now?

src/comp/AConv.hs
```diff
@@ -382,7 +379,7 @@ aIface flags iface@(IEFace i its maybe_e maybe_rs wp fi) = do
                                    ae <- aExpr val_
                                    --trace ("exit av " ++ ppReadable i) $ return ()
                                    return (AIActionValue its' wp g i arule_list
-                                           (ADef i (aTypeConv i t) ae []) fi )
+                                            (ADef i (aTypeConv i t) ae []) fi )
```
nanavati: Skip the whitespace change.

src/comp/AConv.hs
```diff
@@ -638,11 +588,105 @@ aExpr e@(ICon _ (ICInout { iConType = it, iInout = i})) | (isitInout_ it) = do
...
+-- port selected from value part of ActionValue method
+aSelExpr ((sel, atype) : sels) base@(ICon i (ICStateVar { }) : es)
+    | (sel == idPrimFst || sel == idPrimSnd)
+    , [(iav, atypeTup), (m, _)] <- dropWhile ((== idPrimSnd) . fst) sels = do
+  i' <- transId i
+  -- arguments should have been dropped in IExpand
+  when (not (null es)) $
+      internalError ("AConv.aExpr actionvalue value with args " ++
+                     ppReadable sels ++ "\n" ++ ppReadable base)
+  let idx = toInteger $ (if sel == idPrimSnd then 1 else 0) + length sels - 2
```
nanavati: The selector counting and tuple gymnastics seem very similar between this and the value method case. Can we unify the code?

src/comp/AConv.hs
nanavati: I think that would make the code clearer.

src/comp/AExpand.hs
nanavati: I'd also like to see a bit more justification for the calculations here.

src/comp/AExpr2STP.hs
nanavati: Ah, then a comment about ATupleSel (or maybe even a clarifying pattern match) is in order here.

testsuite/bsc.bugs/bluespec_inc/b1490/b1490.exp
```diff
@@ -9,7 +9,7 @@ proc rts_flags { heapsize } {
 # -----
 
 compile_verilog_pass Bug1490Bool.bsv {} $rtsflags
-compile_verilog_pass Bug1490MyBool.bsv {} $rtsflags
+compile_verilog_pass Bug1490MyBool.bsv {} [rts_flags 300]
```
quark17: Do we know why more heap is needed here?

src/comp/AExpr2Yices.hs
```diff
@@ -587,16 +588,37 @@ convAExpr2YExpr mty (AMethCall ty@(ATBit width) modId methId args) = do
...
     -- will appear logically equivalent
     smap <- gets stateMap
-    let portId = getMethodOutputPort smap modId methId
+    let e = case getMethodOutputPorts smap modId methId of
+              [portId] -> AMethValue ty modId portId
+              ports -> internalError ("convAExpr2YExpr: unexpected output ports: "
+                               ++ ppReadable (modId, methId, ports))
+    -- XXX This could be an unevaluated function, applied to converted arguments
+    addUnknownExpr mty e width
+convAExpr2YExpr mty (ATupleSel ty@(ATBit width) (AMethCall _ modId methId args) selIdx) = do
```
nanavati, line 602: Similar comment about ATupleSel here (possibly worse because the mismatch case leads to an internalError?).

testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
```diff
@@ -59,15 +59,15 @@ module mkDesign(clk,
 	       IF_0_CONCAT_IF_0_CONCAT_result_a_BITS_8_TO_4_P_ETC__q2,
 	       IF_0_CONCAT_result_a_BITS_8_TO_4_PLUS_0_CONCAT_ETC___d15,
 	       IF_0_CONCAT_result_a_BITS_8_TO_4_PLUS_0_CONCAT_ETC__q1,
-	       _theResult_____1_snd__h755,
-	       notB__h260,
-	       quotient__h275;
+	       _theResult_____1_snd__h843,
```
quark17: I don't think you need to change this file now.  Check that the test passes without this change and, if so, remove it.

The test filters both the output and the expected file to replace unique suffix numbers with XXX, so that they are not involved in the comparison.  Unfortunately, we're not storing the filtered expected file, but the pre-filter expected file.  That means that there may be differences from BSC's new output, while not actually causing the test to fail (so this update is benign).  You previously had a change to more than just the IDs, which is why you had to change the file; now, I don't think you need to.

testsuite/bsc.misc/lambda_calculus/lc-sysTb.out.expected
```diff
@@ -31,16 +31,16 @@ dim_sysTb =
 rule_RL_r0_sysTb :: MOD_sysTb -> (Bool, MOD_sysTb, ());
 rule_RL_r0_sysTb =
     (\ (state0 :: MOD_sysTb) ->
-     let { (def_c1__h532 :: Bit #51) = meth_read_RegN (inst_c1__sysTb state0)
```
quark17, line 34: I would guess that the filter for unique IDs doesn't replace the letters, so changes from `__h#` to `__d#` require an update here.  (Doubly so because of the extra underscore in `__d` names.)

testsuite/bsc.misc/sal/CTX_sysTb.sal.expected
```diff
@@ -19,17 +19,17 @@ BEGIN
      #) ;
   
   rule_RL_r0 (state0 : STATE) : [ BOOLEAN, STATE ] =
-    LET def_c1__h532 : Bit{51}!T = CTX_RegN{51}!meth_read(state0.inst_c1)
-    IN LET def_c2__h543 : Bit{51}!T = CTX_RegN{51}!meth_read(state0.inst_c2)
+    LET def_c2___d6 : Bit{51}!T = CTX_RegN{51}!meth_read(state0.inst_c2)
```
quark17, line 22: I would guess that the filter for unique IDs doesn't replace the letters, so changes from `__h#` to `__d#` require an update here.  (Doubly so because of the extra underscore in `__d` names.)

testsuite/bsc.scheduler/SplitIf.bsv.bsc-sched-out.expected
quark17: In several update expected files (under `bsc.interra` and `bsc.scheduler`), this is the only difference.

The `__d` suffix is in AConv, while `__h` is heap references in the evaluator.  So it may be because method outputs are not resolved until AConv now?  (Although none of these are for split outputs, just methods with single outputs.)

What's also concerning is the extra underscore.  The suffix only has one underscore (`__d`); the fact that we see three (`dir___d1`) suggests that an underscore is being inserted somewhere.  Oh, this could be an old bug in`AConv.newAIdFromAExpr`:
```
new_name = signalNameFromAExpr expr ++ "_" ++ aconvPref ++ itos n
```
That underscore is unneeded, because `aconvPref` has two underscores.  I don't know if removing it would cause a lot of changes in expected files elsehwre, in which case that can be a later a PR.

src/comp/ASyntax.hs
krame505: Copy/paste error I think.

testsuite/bsc.misc/lambda_calculus/lc-sysDynArrSelWithImplCond.out.expected
krame505: Fixed.  There are still some changes to _h names becoming _d names, and some defs changing order.

testsuite/bsc.scheduler/SplitIf.bsv.bsc-sched-out.expected
krame505: > It's interesting that this changed from __h to ___d -- do we know why that is?

Yeah, it has to do with method outputs being resolved in AConv now.

krame505: Yeah, it looks like the extra underscore is coming from `newAIdFromAExpr`.  I'd prefer to keep that for a future PR, just to reduce the churn in expecteds here.

krame505 (commented):
First pass addressing more minor comments.

src/comp/IfcBetterInfo.hs
krame505: The result-rename pragma is still honored, just earlier: GenWrap reads it via `lookupResultIfcPragma` and threads the chosen name through `saveFieldPortTypes` (GenWrap ~L2210), so by the time we build `BetterMethodInfo` the output port name is already set and `mi_result` is no longer needed here.

src/comp/AConv.hs
krame505: Done.

src/comp/AExpand.hs
krame505: It was totally arbitrary, I can make it just the size/max depth of the sub-exprs.

src/comp/SignalNaming.hs
krame505: Done.

src/comp/VIOProps.hs
krame505: Done.

src/comp/TCheck.hs
krame505: I think I misunderstood what was going on here, this is specifically for handling `module verilog` where bit tuples/multiple output ports should not appear anyway.

testsuite/bsc.bluetcl/commands/module.tcl.bluetcl-bh-out.expected
krame505: I updated the bluetcl section of the user guide.

src/comp/ASyntax.hs
krame505: Previously output ports and methods were interchangable, and `AIDef` contained a single `aif_value` with an `adef_objid` that matched the `aif_name`.  Now one is the method name and one is the port name. There are a number of places where this needed to be updated.

src/comp/IExpand.hs
krame505: I don't have any recollection of why I changed this, but reverting it doesn't seem to make a difference, so I'll revert it.  It's possible that I was testing something and never changed it back.

src/comp/BinData.hs
krame505: Done.

src/comp/ASyntax.hs
krame505: Yeah, in fact it should to match the `vf_outputs` created by `chkVeriIfc` in TCheck.

testsuite/bsc.bugs/bluespec_inc/b1490/b1490.exp
krame505: I don't know, but it no longer seems to be needed.

src/comp/CType.hs
krame505: This is due to the change to `ActionValue_` now taking a tuple of `Bit` types, instead of a numeric type parameter.

src/comp/SimMakeCBlocks.hs
krame505: Hmm.  I discovered that this actually is the incomplete remains of an earlier version before I decided that we should add tuples in ASyntax - currently for the sim backend, this was always either an empty list or a single element containing an `ATuple`.  It looks like a substantial portion of the sim changes can actually be reverted now.

testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
krame505: This is no longer the case.  I think it may have had to do with the `getExprSize` change for `ATuple`/`ATupleSel`?

src/comp/APaths.hs
krame505: I added some more tests.

krame505:
I believe I have now addressed all comments.

> I've had a look over all the src changes and most of the testsuite. It seems like you left in the IExpand heuristic for underscore names, and I haven't carefully audited all the name changes due to that -- I wasn't sure if that's intentional.

That was an accident.  I have removed that and rebased on main, the expecteds are all up to date (and much less changed.)

> Can you say a little something about the change to ActionValue_? The argument used to be a numeric type, but now it's either Bit n or Tuple. Can you saw what was involved in that change?

Yeah, it used to have a single numeric param that was just the size of the single output.  Now there are potentially multiple outputs, so it needs to be able to contain either a single `Bit` value or a tuple of `Bit` values.

> Can you also say a little about the change, for AIFace, to use the aif_name ID rather than the ID inside the aif_value Def. (I think this is for both AIDef and AIActionValue.)

I don't 100% remember what the reason that this was needed.  But I think it had something to do with that the method name was showing up in one place and the port name in the other, and these were previously interchangeable, but now there are some cases where one is expected or the other.  

> If I understand, Bluesim works with split output ports ... by having Bluesim just convert the tuple back to one concatenated bit vector. So Bluesim isn't going to look any different than before -- it won't show the separate names in a VCD dump, for example. (So there's potentially room in the future for better support in Bluesim.)

Yeah, this could probably be improved.  For example we might want to generate structs to hold the return values for methods with multiple output ports, instead of just turning them into bit vectors.

> I noted a few places where I think we have more tests.

Addressed.

testsuite/bsc.verilog/splitports/DeepSplit.bs
```diff
@@ -74,4 +89,13 @@ sysDeepSplit =
...
+      when i == 6 ==> $display "getFoo: " (cshow s.getFoo)
+      when i == 7 ==> $display "getBar: " (cshow $ s.getBar s.getFoo)
+      when i == 8 ==> $display "getZug: " (cshow s.getZug)
+      when i == 9 ==> do
+        res <- s.update (DeepSplit $ Foo { x = 77; y = 88; })
+        $display "update: " (cshow res)
+      when i == 10 ==> do
+        res <- s.update (DeepSplit $ Foo { x = 33; y = 44; })
+        $display "update: " (cshow res)
+      when i == 11 ==> $finish
```
quark17: This is missing a final newline

quark17 (commented):
Aside from a missing newline, I think this is ready!

src/comp/AConv.hs
```diff
@@ -507,69 +504,17 @@ aExpr (IAps (ICon i (ICSel { })) ts (e:es))
...
-        t = aTypeConvE e (iGetType e)
-    in
-        -- the value side carries no arguments
-        -- the cookie "n" will connect it back up to the action side
-        return (ATaskValue t i name isC n)
+      unfoldICSel :: IExpr a -> ([(Id, AType)], [IExpr a])
+      unfoldICSel e@(IAps (ICon i (ICSel {})) _ [e']) =
+          let (sels, a) = unfoldICSel e'
+          in  ((i, aTypeConvE e $ iGetType e) : sels, a)
+      unfoldICSel e@(IAps (ICon i (ICSel {})) _ a) = ([(i, aTypeConvE e $ iGetType e)], a)
```
nanavati: I think this is mainly covering the methods with arguments case, but it also matches an empty list. Is the empty list case meaningful?

src/comp/AConv.hs
```diff
@@ -638,11 +588,113 @@ aExpr e@(ICon _ (ICInout { iConType = it, iInout = i})) | (isitInout_ it) = do
   ai <- aInout i
   return (ASInout at ai)
 
+aExpr (ICon i _) | i == idPrimUnit = return $ ASInt i (ATBit 0) (ilDec 0)
```
nanavati, line 596: Do you understand where the PrimUnit is coming from?

I think it's probably fine to filter out zero-width things in A syntax, but I wonder if this is a behavior change worth understanding.

src/comp/AConv.hs
```diff
@@ -638,11 +588,113 @@ aExpr e@(ICon _ (ICInout { iConType = it, iInout = i})) | (isitInout_ it) = do
...
+-- This covers both value methods (which still carry their arguments) and the
+-- value part of an ActionValue method (whose arguments were dropped in
+-- IExpand, leaving an extra avValue_ selector before the method).  The leading
+-- PrimSnd selectors are skipped to reach the method; the number of skipped
+-- selectors (length sels - length rest) plus the head selector gives the
+-- 0-based output port index.
+aSelExpr ((sel, atype) : sels) base@(ICon i (ICStateVar { }) : es)
+    | (sel == idPrimFst || sel == idPrimSnd)
+    , let rest = dropWhile ((== idPrimSnd) . fst) sels
+    , length rest == 1 || length rest == 2 = do
```
nanavati: The rest == 1 case is a value methods and rest == 2 is an ActionValue?

If so, I think it would be better to separate them cleanly and handle them directly. Especially if that opens the door to unifying the case of selectors applied to an ActionValue output with the bare ActionValue output itself (the next clause of aSelExpr).

src/comp/AConv.hs
```diff
@@ -662,12 +714,14 @@ aTypeConv _ (ITAp (ITCon i _ _) (ITNum n)) | i == idInout_ = ATAbstract idInout_
...
+  ATTuple (aTupleTypesConv a t)
 aTypeConv _ t | t == itReal = ATReal
 aTypeConv _ t | t == itString = ATString Nothing
 -- Deal with AVs
-aTypeConv _ (ITAp (ITCon i t (TIstruct SStruct fs@(val:_))) (ITNum n)) =
-    ATBit n
-    -- internalError ("Yes\n\n" ++ (show a) ++"\n\n" ++ (show n))
+aTypeConv a (ITAp (ITCon i _ _) t) | i == idActionValue_ =
+    aTypeConv a t
+aTypeConv a (ITCon i _ _) | i == idPrimUnit = ATBit 0
```
nanavati, line 726: So the Bit 0 thing is a general issue? And seems new? Then I definitely want to understand it (and fix it if possible).

src/comp/AExpr2Yices.hs
```diff
@@ -583,20 +584,46 @@ convAExpr2YExpr mty (APrim i (ATBit width) p args) = do
 -- Method calls create independent variables, with given width
 -- XXX Passing the current context is just a heuristic
 -- XXX TODO: some methods calls may be mutex, such as FIFO.full and FIFO.empty
+-- A bare AMethCall/AMethValue (with no ATupleSel applied) refers to a method
```
nanavati, line 586: Not strictly necessary, but noticing the repeated code if some of it could be moved to AExpr2Util, that might be an improvement.

src/comp/bluetcl.hs
nanavati, line 1: I it is probably worth some tests to make sure port types come through cleanly all the way to `bluetcl` in the case of multiple output ports, if you haven't written that already.

src/comp/IExpand.hs
```diff
@@ -318,8 +318,9 @@ iExpand errh flags symt alldefs is_noinlined_func pps def@(IDef mi _ _ _) = do
           in
               -- return the expression that should replace the heap pointer,
               -- and maybe a Def, if the expression is a def reference
-              if simple e || isActionType t then
+              if simple e || isActionType t || isPairType t then
```
nanavati, line 321: Isn't the proper check not just that it is a pair type but that the eventual leaves of the pair type are all simple? And presumably some similar check on the payload of the ActionType (since it is no longer guaranteed to be Bit n).

src/comp/IExpand.hs
```diff
@@ -318,8 +318,9 @@ iExpand errh flags symt alldefs is_noinlined_func pps def@(IDef mi _ _ _) = do
...
                   -- inline the expression, no def is created for this heap ptr
+                  --trace ("not inlining " ++ show i ++ " " ++ ppReadable t) $
```
nanavati: Do you want to keep this trace?

src/comp/IExpand.hs
```diff
@@ -1158,8 +1160,8 @@ iExpandMethod' implicitCond curClk (i, bi, e0) p0 = do
             rdyId      = mkRdyId i
         let enablePort :: Maybe VPort
             enablePort = toMaybe (isActionType methType) (BetterInfo.mi_enable bi)
-        let outputPort :: Maybe VPort
-            outputPort = toMaybe (isValueType  methType) (BetterInfo.mi_result bi)
+        let outputPorts :: [VPort]
+            outputPorts = map (id_to_vPort . mkId (getPosition i) . mkFString) outs
```
nanavati, line 1163: Has the tuple type / number of output ports consistency check happened somewhere else?

src/comp/IExpand.hs
```diff
@@ -2562,20 +2564,16 @@ walkNF e =
...
-
-                        -- the inner selector can wind up on the heap
-                        -- because of "move" in evalHeap
-                        [e_ref@(IRefT t ptr ref)] | (isitActionValue_ t) || (isitAction t)
-                            ->  do (P p' e', ws) <- walkNF e_ref
-                                   upd (pConj p0 p') (IAps f ts [e']) ws
+                        -- We can be selecting the avValue or avAction from an ActionValue method,
+                        -- or a tuple member out of the result of calling a method with multiple outputs,
+                        -- and need to recurse.
+                        [e] | (i_sel == idAVValue_ ||
```
nanavati, line 2569: This could be a bit stricter (no action recursing inside an ActionValue value, no recursing inside the action at all, value recursion also can't contain actions). I'm not sure if that extra strictness is worthwhile, but I thought it was worth considering.

src/comp/IExpand.hs
```diff
@@ -3878,7 +3887,7 @@ conAp' _ (ICPrim _ op) fe@(ICon prim_id _) as | strictPrim op = do
               when doTrans $ traceM ("conAp: iTransform fallthrough: " ++ ppReadable (op, mkAp fe as'))
               errh <- getErrHandle
               case (iTransExpr errh (mkAp fe as')) of
-                  (e', True) -> do
+                  (e', True) | isBitType (iGetType e') -> do
```
nanavati, line 3889: I think this is probably the right choice (iTransExpr doesn't really handle non-Bit types), but please be careful that none of the `pack . unpack` optimizations regress.

src/comp/IfcBetterInfo.hs
nanavati, line 1: Are we reaching the point where IfcBetterInfo can be sunset? If not, what is missing (since I thought all the renaming pragmas are handled in GenWrap  via PrimMethod)?

src/comp/ISplitIf.hs
nanavati: Do we if-split top-level methods at all?

src/comp/ISyntaxUtil.hs
```diff
@@ -121,18 +121,16 @@ isSimpleType t = t == itInteger ||
...
-    | (i == idActionValue_) = (x == 0)
-isitAction (ITAp (ITCon i (IKFun IKStar IKStar) _) t)
-    | (i == idActionValue)  = t == itPrimUnit
+isitAction (ITAp (ITCon i (IKFun IKStar IKStar) _ ) t)
+    | (i == idActionValue_) || (i == idActionValue) = isEmptyType t
 isitAction x = (x == itAction)
 
--- note this returns false for x == - because ActionValue_ 0 is really an Action
+-- note this returns false for x == () because ActionValue_ () is really an Action
+-- Also handle ActionValue_ (Bit 0), which can be introduced by foreign functions.
```
nanavati, line 129: I continue to be somewhat skeptical of being more permissive about `Bit 0`.

src/comp/ITransform.hs
```diff
@@ -249,8 +249,8 @@ runCSE e@(IAps _ _ _) = do
   -- because "runCSE" is called from iTrExpr, which already recurses on the
   -- arguments, so runCSE will already have been called on the arguments.
   let t = iGetType e
-  -- Only CSE applications that are not actions
-  if not (isActionType t) then
+  -- Only CSE applications that are not actions or tuple method values
+  if not (isActionType t || isPairType t) then
```
nanavati, line 253: This is because we want the stack of ICSel to be directly visible for AConv?

src/comp/SimPackage.hs
```diff
@@ -385,9 +385,14 @@ getPortInfo pps aif =
         ps   = map fst (vf_inputs vfi)
         ins  = [ (t,i,vn) | ((i,t),vn) <- zip args ps ]
         rt   = aIfaceResType aif
-        ret  = case (vf_output vfi) of
-                 (Just (vn,_)) -> Just (rt,vn)
-                 Nothing       -> Nothing
+        -- A value method has at most one result.  In Bluesim that whole value
```
nanavati, line 388: Does this complicate making VCD output for methods with multiple output ports work?

src/Libraries/Base1/Prelude.bs
```diff
@@ -3536,6 +3548,7 @@ tuple8 :: a -> b -> c -> d -> e -> f -> g -> h   -> Tuple8 a b c d e f g h
 tuple8 a b c d e f g h = (a,b,c,d,e,f,g,h)
 
 class AppendTuple a b c | a b -> c where
```
nanavati, line 3550: I am wondering how some of these tuple type classes avoid being incoherent because when applied to a polymorphic tuple because when the right parameter is substituted in for a tuple parameter, it starts looking like a bigger tuple. 

Can you explain why this works out?

src/Libraries/Base1/Prelude.bs
```diff
@@ -4691,94 +4710,135 @@ instance (SplitPorts a p, TupleSize p n, WrapPorts p pb, WrapMethod b v, Curry (
...
 
-  inputPortNames _ (Cons h t) = checkPortNames (_ :: a) h `listPrimAppend` inputPortNames (_ :: b) t
+  inputPortNames _ (Cons h t) =
+    filterZeroWidthPorts (_ :: PortsOf a) (checkPortNames (_ :: a) h)
+    `listPrimAppend` inputPortNames (_ :: b) t
   inputPortNames _ Nil = error "inputPortNames: empty arg names list"
 
+  outputPortNames _ = outputPortNames (_ :: b)
+
   saveMethodPortTypes _ modName (Cons h t) result = do
```
nanavati, line 4720: Is this now saving types for both input and output ports? It's entirely possible I've missed it, but where are the two lists being joined?

testsuite/bsc.scheduler/paths/paths.exp
nanavati, line 1: I suspect more test cases would be useful here since there are a lot of scenarios possible.

nanavati (dismissed):
Lots of small comments to clean up, but otherwise looks really good!

src/Libraries/Base1/Prelude.bs
krame505: Yeah - `savePortTypes` here saves it for this input, and the recursive `saveMethodPortType` saves the rest of the ports, eventually hitting the base case that saves the outputs in one of the two below instances.

src/comp/AConv.hs
krame505: Yeah, a type that splits into 0 non-empty ports yields an empty tuple aka `PrimUnit`.

src/comp/AConv.hs
krame505: This is just handling `PrimUnit` here where it didn't used to be (because an input that split into no ports was handled by `Curry`.)

src/Libraries/Base1/Prelude.bs
krame505: I'm not sure I completely follow the concern.  Adding `incoherent` seemingly breaks things (the `AppendTuple` test in the testsuite fail to compile), and for the purposes of the `SplitPorts` type class the tuple types on which this is called should be fully concrete.  I didn't change the type class at all in this PR, bgesides adding the `TAppendTuple` type function.  Can we deal with this concern later?

src/comp/SimPackage.hs
krame505: Maybe.  I haven't really looked into it that deeply.  I do recall that initially I tried pushing the seperate ports all the way through Bluesim and determined that wasn't really feasible, and ASyntax tuples were needed.

src/comp/AConv.hs
krame505: Do you mean `ICSel` applied to an empty list?  No, that shouldn't ever occur.

src/comp/ISyntaxUtil.hs
krame505: I don't totally remember anymore why this change was needed, but I think this may have been a case where we previously had a `PrimAction`.

src/comp/IfcBetterInfo.hs
krame505: Getting there.  Ready/enable signal naming is still done here, but those could probably also be moved to `PrimMethod`.  Not sure about the prefix, since that is also used in interface flattening.

src/comp/AConv.hs
krame505: I think you previously suggested that I combine them, unless I misunderstood your previous comment.  I can make that change.

src/comp/AExpr2Yices.hs
krame505: Done.

src/comp/IExpand.hs
krame505: I don't think so?  This is an exception to the `simple` check to also make sure that tuples and actions get flattened without introducing extra defs.

src/comp/IExpand.hs
krame505: This gets checked in `checkPortNames` in the Prelude called by the `WrapField` utilities.

src/comp/ITransform.hs
krame505: Yes.

src/comp/IExpand.hs
krame505: I don't think I follow.  There can still be an arbitrary expression inside the thing being selected from until we do `walkNF` on it?

src/comp/bluetcl.hs
krame505: Done.

testsuite/bsc.scheduler/paths/paths.exp
krame505: I added a large pile of additional tests.

krame505:
I addressed/responded to Ravi's comments and added a lot more tests for path analysis - I think this is ready to merge now.

quark17 (approved):

PR #852, opened 2025-12-08
Replace MVarStrict Control.Concurrent.MVar.Strict

nanavati:
This is built on top of #851 

It is separate because it involves adding a dependency on a new package from Hackage: strict-concurrency.


nanavati:
> Seems reasonable. Note that the CI is failing, because the `build-and-test-ubuntu.yml` and `build-and-test-macos.yml` files in `.github/workflows/` need to be updated to add `strict-concurrency` to the `cabal` install command.

Added. Now let us hope CI works.

PR #854, opened 2025-12-08
Improve the performance of expression and type substitution in ISyntax

nanavati:
This is performance work building on top of #851 

This includes the following improvements:
- switching to the set versions of ISyntax free variable operations almost everywhere
- batching expression and type substitutions in the evaluator to save traversals
- specializing the batch traversal operations so that we don't lose much performance (if any) in the single-substitution case
- only rebuilding a substituted expression when something changes which doesn't just improve performance and reduce allocation, it also helps preserve sharing between expressions and types where possible

### DIFF
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -2868,8 +2868,10 @@ evalApAccum tag exprCtx typeCtx (ILam i t body) (E a : as) = do
   evalApAccum "ILam-accum" (M.insert i a' exprCtx) typeCtx body as
 
 -- Continue accumulating for ILAM with type argument
-evalApAccum tag exprCtx typeCtx (ILAM i k body) (T t : as) =
-  evalApAccum "ILAM-accum" exprCtx (M.insert i t typeCtx) body as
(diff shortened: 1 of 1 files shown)


src/comp/IExpand.hs
```diff
@@ -2849,6 +2855,35 @@ mkApUH :: HExpr -> [Arg] -> G HExpr
 mkApUH f es = do es' <- mapM evalArgUH es
                  return (mkAp f es')
 
+-- Accumulate substitutions when applying to a chain of ILam/ILAM
+-- This batches substitutions to avoid repeated hyper calls
+
+-- Continue accumulating for ILam with expression argument
```
quark17: This comment should go below the type signature, on the `ILam` arm.

quark17 (commented):
I've rebased onto `main`, removing the commits that were from 851.  Looks good!  But note that I didn't look too deep at every line in `ISyntaxSubst`.  The testsuite is passing, so I'll merge it, but I want to correct the position of a comment first.

Do you want to preserve the commit history or should I squash merge?

PR #856, opened 2025-12-12
Use a difference list in BinData to speed up the writeBin phase.

nanavati:


nanavati:
> Seems reasonable. I don't have a sense of what's efficient or not, so can't comment on that. For example, it seems like `Out` is only ever appending one element at a time, so would it be efficient to build the list in reverse, always just cons'ing an element on the top?

Building the list in reverse might also work, but, where possible, I like to avoid solutions that force you to think about when things are and aren't in reverse order. The difference list is a simple solution to the slowdown in appending as bo files get bigger that doesn't grow stack usage (as Data.Sequence did) which is why I went with it. I was happy that I didn't have to look at the details of how things were being appended (beyond knowing that we repeatedly appended small things to a bigger one).

PR #857, opened 2025-12-12
Avoid constructing an intermediate map in fixupBack.

nanavati:


nanavati:
It's possible GHC was doing some deforestation. The win from this wasn't as big as I'd hoped, but there were still improvements in time and allocation so I think it is still worth doing.

PR #859, opened 2025-12-14
Some miscellaneous improvements to dump flags

nanavati:
There are two headline improvements here:

1. Allowing the `%s` digraph in dump file names to get the compiler pass included in the output file name.
2. Allowing  `-dall` to take a file argument as well.

To get `-dall=file` (with no digraphs) to work, I had the compiler start appending to dump files after the first write.

That turned out to be tricky because of how we were using dump flags in the testsuite. In one case I redirected dumps with `%m` so that tests didn't see the dumps of multiple modules mixed together. The other issue was how our double-parsing (once for dependencies and once for compilation) interacted with the dump code. I decided the least-surprising behavior that worked with our tests was to have the parsing stages be an exception to the append after first write rule (when they are being explicitly dumped).


quark17 (commented):
Awesome.  Just a couple comments.

src/comp/TopUtils.hs
```diff
@@ -94,19 +97,38 @@ sdump errh flags t d names a =
...
     case (dumpInfo flags d) of
         Just (Just file) -> do
-            writeFileCatch errh (substNames names file) a
+            let dumpPath = substNames names' file
+            currentDumped <- readVar dumpedFiles
+            let hasBeenDumped = dumpPath `S.member` currentDumped
+            let isSpecificDump = hasDumpStrict flags d
+            -- Don't allow explicit parsed dumps to accumulate to avoid a double dump
+            -- from the double parse (first for dependencies and then for code).
+            let doubleParsedPasses = [DFvpp, DFbsvlex, DFparsed]
```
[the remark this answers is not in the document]
nanavati: Part of the strangeness of the parser passes is because of their use in the testsuite. The other unusual thing about them is that, unlike other repeated passes, they're running the same code on the same inputs - just capturing dependencies once and the whole package the second time.

I wasn't worried about parser stages overwriting anything other passes dump because the parser stages are first. I guess that does leave open the case of parser pass dumps with -u, which I haven't thought about.

src/comp/TopUtils.hs
```diff
@@ -94,19 +97,38 @@ sdump errh flags t d names a =
...
+            let isDoubleParseDump = isSpecificDump && d `elem` doubleParsedPasses
+            let writeFileFn = if hasBeenDumped && not isDoubleParseDump
+                              then appendFileCatch
+                              else writeFileCatch
+            -- Add a header and footer to -dall dumps in case they end up in the same file.
+            let a' = if hasDumpStrict flags d then a else header ++ a ++ footer
```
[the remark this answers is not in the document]
nanavati: Added headers and footers unconditionally and taught the testsuite to remove them.

src/comp/Depend.hs
```diff
@@ -349,33 +342,72 @@ flags in the CC variable, for example CC="cc -g", then it will work.
...
+-- If False, it's just a bsWarning
+parseFile :: ErrorHandle -> Flags -> Bool -> FilePath -> IO (CPackage, TimeInfo, [WMsg])
+parseFile errh flags fatal_name_mismatch fname = do
+    let isClassic = hasDotSuf bscSrcSuffix fname
+
+    t <- getNow
+    let dumpnames = (Just (baseName (dropSuf fname)), Nothing, Nothing)
+
+    start flags DFcpp
+    file <- doCPP errh flags fname
```
quark17: I think this is the source of the testsuite error.  Before, `doCPP` was called with the encoded name -- but now, that's not defined until a few lines below, as `fname_encoded`.  If I move that definition earlier and pass it to `doCPP` (in place of `fname`), then the test passes.

src/comp/bsc.hs
```diff
@@ -298,45 +301,21 @@ compile_no_deps errh flags name = do
...
     -- ===== the break point between file manipulation and compilation
 
-    -- We don't start and dump this stage because that is handled inside
-    -- the "parseSrc" function (since BSV parsing has multiple stages)
-    (pkg@(CPackage i _ _ _ _ _), t)
-        <- parseSrc (syntax == CLASSIC) errh flags True name file
-    when (getIdString i /= baseName (dropSuf name)) $
-         bsWarning errh
-             [(noPosition, WFilePackageNameMismatch name_rel (pfpString i))]
+    (pkg, t, parse_warns) <- parseFile errh flags False name
```
quark17, line 298: I note that `parseFile` constructs the encoded name, and down below inside `compilePackage`, it also constructs the encoded name -- I don't know if it's cleaner to just create the name once and stick with it.  (But it's a bit unclear to me when an unencoded name is needed.)

(Also, it's not clear when a function expects an encoded name or not, and when a name passed to it is encoded or not.  I don't know if it's worth using variables like `encoded_name` and `unencoded_name` everywhere; or if the typecheck can keep track for us, by making them separate types.)

What you've got is ok, though -- it's fine to worry about minor cleanups later.  Just thinking outloud.

src/comp/bsc.hs
```diff
@@ -298,45 +301,21 @@ compile_no_deps errh flags name = do
...
-    setSyntax syntax
-
-    t <- getNow
-    let dumpnames = (baseName (dropSuf name), "", "")
-
-    start flags DFcpp
-    file <- doCPP errh flags name
-    _ <- dumpStr errh flags t DFcpp dumpnames file
-
+compileFile errh flags binmap hashmap name = do
```
quark17: Consider inlining `compileFile` into `compile_no_deps` -- `compile_no_deps` does nothing but call `compileFile` and is the only caller of it.

src/comp/bsc.hs
```diff
@@ -298,45 +301,21 @@ compile_no_deps errh flags name = do
...
     -- ===== the break point between file manipulation and compilation
```
quark17: BTW, I'm not sure this comment is meaningful anymore?

testsuite/bsc.bluetcl/packages/makedepend/error1.bluetcl-out.expected
```diff
@@ -1,3 +1,3 @@
-Error: "./Foo.bsv", line 7, column 13: (P0005)
+Error: "Foo.bsv", line 7, column 13: (P0005)
```
quark17, line 1: This seems like an improvement, but do you know why it happened?

testsuite/bsc.driver/depend/WFilePackageNameMismatch_Top.bs.bsc-out.expected
```diff
@@ -1,4 +1,4 @@
 checking package dependencies
-Error: "./WFilePackageNameMismatch_Sub.bs", line 1, column 8: (S0001)
-  File name `WFilePackageNameMismatch_Sub' does not match package name
+Error: "WFilePackageNameMismatch_Sub.bs", line 1, column 8: (S0001)
+  File name `./WFilePackageNameMismatch_Sub.bs' does not match package name
```
quark17, line 3: I'm not sure if this change in the body is for the best, but I have no strong objection.

testsuite/bsc.interra/preprocessorTestcases/define/Define_BITVECTOR.bsv.bsc-out.expected
```diff
@@ -1,5 +1,5 @@
 checking package dependencies
-compiling Define_BITVECTOR.bsv
-Warning: Unknown position: (S0001)
+Warning: "Define_BITVECTOR.bsv", line 5, column 9: (S0001)
```
quark17, line 2: Cool to see the position improve (though I guess the name was in the body of the message anyway).  Also cool to see that the message comes out during checking of the dependencies, before the "compiling" message.

testsuite/bsc.syntax/bsv05/import-foreign/ImportForeignModule_DefaultClock_NoName_InputClock_NoName.bsv.bsc-out.expected
```diff
@@ -20,24 +20,4 @@ endpackage: ImportForeignModule_DefaultClock_NoName_InputClock_NoName
 -----
 
 compiling ImportForeignModule_DefaultClock_NoName_InputClock_NoName.bsv
-=== parsed:
```
quark17, line 23: Cool, evidence of no duplicate parsing now.

testsuite/bsc.preprocessor/include/DupInclude.bsv.bsc-out.expected
```diff
@@ -9,13 +9,4 @@ Warning: "DupInclude.bsv", line 1, column 1: (S0089)
     libdir1/DupInclude.defines
     libdir2/DupInclude.defines
 compiling DupInclude.bsv
-Warning: "DupInclude.bsv", line 1, column 1: (S0088)
```
quark17, line 12: Ah, these two messages are deleted, because they are duplicates of the two messages above -- they occur during parsing, and we've eliminated the unnecessary second parsing, so we've eliminated the second batch of warning messages.  Cool!

quark17 (commented):
Looks good!  Almost ready to merge!

I confirmed that passing `fname_encoded` to `doCPP` doesn't affect anything else, so I'll push that change and the testsuite should go green -- but feel free to adjust it, if you don't like my exact edits.

I just have some minor questions below.

Did you want to add any new tests, for new behaviors?  For example, the appending to previously opened files?  Or anything to do with `-dall` changes?  (Fine for that to go in a later PR, if necessary.)

If you hadn't wanted to change all the expected files for tests that were dumping stages, we could have added a filter in the Tcl functions, to remove the header/footer.  But since you've updated everything, that's good too.  (As long as we don't change our mind about the formatting!)

Anyway, great to see this PR finished!

quark17:
Oh, and did you want to preserve the commit history?  If so, feel free to merge my fixup commit into wherever it should go.

quark17:
Ok.  I've rebased.  And I've edited the commit that put `strip_dump_wrapper` in `unix.exp` to put it in a better place.  And I've pushed a commit that fixes the Mac (by passing an encoded absolute path to CPP) -- feel free to edit that or merge into an earlier commit, or not.  I expect the testsuite will go green and we'll be able to merge (pending anything you want to change in response to my minor comments).

quark17:
The rebase introduced some failures, which I've fixed by adding testsuite changes to the original commits that triggered them.

An introduced use of `-dtypecheck=` needed to be changed to `-dwrapper_typecheck`.

And for noinline tests, Bits instance errors are reported in a different order ... which I thought was because of eliminating double parsing, but now I realize that doesn't make sense: both messages come out in typecheck.  That _is_ the commit where the difference was introduced, though.  But here's what I observe: the order of the messages changes when `-u` is provided!  If I compile manually, without `-u`, the output is unchanged; but because the testsuite uses `-u`, the expected file needs to change.

The testsuite should go green, but I'd like to understand this behavior change.  Do you have an understanding why `-u` would affect the order?  What changed in commit 0b6e746b that make the flag affect the order of typecheck messages?

src/comp/bsc.hs
nanavati: Yeah, it seems obsolete now that the real parse is happening earlier. Delete it, I guess?

src/comp/bsc.hs
nanavati: I think it is a tradeoff between constructing the encoded name twice vs passing arguably redundant information between the two functions. As long as constructing the encoded name isn't too painful, I'd rather use the simpler interface which can't pass through inconsistent filenames.

src/comp/bsc.hs
nanavati: Sounds fine as a future cleanup

testsuite/bsc.bluetcl/packages/makedepend/error1.bluetcl-out.expected
nanavati: I assume these changes about the details of when filenames are and aren't encoded. As long as none of the messages are inaccurate (e.g. pointing to modules that don't exist or something) I don;t have a strong preference.

testsuite/bsc.driver/depend/WFilePackageNameMismatch_Top.bs.bsc-out.expected
nanavati: Ditto

quark17: Oh!  I think this is merely because `parseFile` is passing `fname` to the error instead of `fname_encoded`, and I suspect that doing that will restore the name.

quark17: No, my mistake, it's not that.  It's that the file is a dependency that is being compiled, and the filename is constructed from prepending the directory (from the search path) onto the package name.

So this is providing useful information about exactly what file has been opened.  In theory, we could strip the path and just report the basename -- the user would need to turn on verbosity to see exactly which of the search-path directories the file was found in.  Or we could look for `./` specifically and strip that, but that seems fragile.  Seems reasonable to leave things as they are.

quark17:
When I double-checked the noinline tests, I found that _prior_ BSC was giving different order with `-u` or not, but new BSC (with your changes) is stable.  I guess I was just mistaken when I commented before?  Anyway, this PR is an improvement!

I've pushed a commit to remove that old comment and inline `compileFile`.  I think that's all that's left to be done.  When this goes green (it passes on my computer), I'll merge it.  Thanks!

PR #871, opened 2026-01-07
Treat names starting with _ as bad names in IExpand

krame505:
This is needed to prevent a slight regression in signal naming in my port-splitting branch, as the `WrapPorts` type class must apply functions such as `id` to `Bit` values corresponding to output ports in synthesized modules, causing them to become "infected" with the function parameter id when creating defs in `IExpand`.  

After some discussion with @nanavati, the best fix we can think of is to add a heuristic to treat names starting with `_` as bad names in `IExpand`, and change `id` (and also `appendTuple` in my branch) to prefix parameters with underscores.  This in itself causes some limited changes to signal names, which seem to be about the same quality or a slight improvement now.  But we thought it best to make this change in a separate pull request, to clarify what differences in signal names are due to this change vs. other port-splitting-related changes.

### DIFF
--- a/src/Libraries/Base1/Prelude.bs
+++ b/src/Libraries/Base1/Prelude.bs
@@ -2573,7 +2573,7 @@ messageM s = do
 --@ endfunction: id
 --@ \end{libverbatim}
 id :: a -> a
-id x = x
+id _x = _x
 
 --@ Make a function curried
 --@ \index{curry@\te{curry} (Prelude function)}
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -285,7 +285,10 @@ iExpand errh flags symt alldefs is_noinlined_func pps def@(IDef mi _ _ _) = do
               t = iGetType e
               -- the name for the new IDef being created
               i = case expr_name of
-                    Just name ->
+                    -- Heuristic: treat names starting with _ as bad names,
+                    -- as these may come from the arguments of functions like
+                    -- id _x = _x
+                    Just name | not (isEmptyId name) && head (getIdBaseString name) /= '_' ->
                         setKeepId $
                         mkIdPost name (mkFString (iExpandPref ++ show p))
                     _ ->
--- a/testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
+++ b/testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
@@ -46,108 +46,105 @@ module mkDesign(clk,
   wire [10 : 0] result;
 
   // remaining internal signals
-  wire [5 : 0] _0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_I_ETC___d55,
-	       _0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_r_ETC___d56,
-	       _0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_result_a_BITS_ETC___d57,
-	       _0_CONCAT_IF_0_CONCAT_result_a_BITS_8_TO_4_PLUS_ETC___d58,
-	       _0_CONCAT_result_a_BITS_8_TO_4_PLUS_0_CONCAT_IN_ETC___d59,
-	       x__h144,
-	       x__h146,
-	       x__h240,
-	       x__h242,
-	       x__h336,
-	       x__h338,
-	       x__h432,
-	       x__h434,
-	       x__h528,
-	       x__h530,
-	       y__h147;
-  wire [4 : 0] IF_0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_IF_0_CONCA_ETC___d54,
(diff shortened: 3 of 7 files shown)


quark17:
Did you consider looking for two underscores, rather than one?  That is a heuristic that BSC already uses for instantiation names.  And did you consider an attribute -- in the same way that we have "hide" and "keep", we could put an attribute on function arguments.  (Although maybe underscores is simpler.)  A big hammer would be automatically putting the attribute on arguments of functions inside the Prelude.  I'm not sure I have an opinion, just want to understand what you considered.  I think that two underscores would eliminate the testsuite differences -- it's only because of names like `_read` (register method) and `_theResult` (created by the BSV parser, for struct creation and in do-blocks), which are now no longer marked "keep", that we see differences.  But these are possibly names that shouldn't be kept anyway.  (Although I'm curious how we get just `_read` and not `foo_read`?)  (I was also wondering if there's an issue with created names like `name1_CONCAT_name2`, where the first name starts with an underscore; but I believe that we only construct names from expressions in ASyntax, so that's not a concern at the evaluator.)

I notice that there are two kinds of "changes": (1) unique numbers change (`__d` numbers decrease, `__h` numbers increase) and (2) the base of the def name changes (`_read` becomes the name of the register).

It's worth noting that the first category are not real "changes".  When the testsuite compares files, it filters out the unique numbers; so those unique numbers have been changing in the output due to past commits to BSC, but no change has been needed in the expected file.  For files with category 2 changes, though, we now need to update the expected file, and so all the changes to the unique numbers become visible.  If we pre-filtered the expected output to not have unique numbers in it, then we wouldn't see these differences when need to make changes. (But I don't know if there's anything to be done about it now.)  Anyway, just a note for reviewers that only the category 2 changes are real.

Generating reasonable names in the output is important, so it's important to understand the impact on users (for base name changes).  It's interesting that we're seeing so few differences.  And most of those are in the SAL/LambdaCalc dumps, which occur before the backends, when inlining happens -- so maybe most other tests aren't see this because changed names have been inlined anyway.  Or maybe we don't have sufficient testing for def names?!  I don't recall -- I'll have to look in `bsc.names`, to see what we have.

The name change in the SAL/LambdaCalc tests are due to `_read` becoming the name of the register.  For example, we have this code in `bsc.misc/lambda_calculus/MultiArityConcat.bsv`:
```
module sysMultiArityConcat();
   Reg#(Bit#(3)) src1 <- mkRegU;
   Reg#(Bit#(4)) src2 <- mkRegU;
   Reg#(Bit#(5)) src3 <- mkRegU;

   Reg#(Bit#(12)) snk <- mkRegU;

   rule d;
      snk <= { src1, src2, src3 };
   endrule
endmodule
```
And previously BSC generated this in `expanded` stage:
```
_read__h37 :: Prelude.Bit 3
_read__h37 = .Prelude.read ·3 src1
-- IdProp: _read__h37[IdP_keep]
```
Which has a "keep" property.  I'm surprised that a name `_read` was created!  Possibly it's because of constructing the fields of an interface, so the names of the fields are picked up.  Anyway, with your change, the output is this:
```
__h37 :: Prelude.Bit 3
__h37 = .Prelude.read ·3 src1
-- IdProp: __h37[IdP_bad_name]
```
And that ultimately leads to a def named `src1` in the ASyntax instead of `_read`.  (This is an improvement, but if the def is called `_read` because it's the expression of a struct/interface field, then this "improvement" won't apply to other field names, maybe arguing for a better mechanism here, although maybe registers are different from other kinds of module boundaries, but again that could maybe be handled by putting a "bad" prop on the interface type).

I guess I'm OK with this change, but still wondering whether two underscores might be the better answer for now (if it avoids testsuite differences).  But if this improves names for registers, that's good, so I don't know.

nanavati:
I agree that the "bad" name heuristic should be the same between instantiations and these function names we're talking about. I just thought that the heuristic for instantiations was a single _, not two.

I think that logic is here: https://github.com/B-Lang-org/bsc/blob/5dccc4bad4dcfdaa67b9f74fc3b1096e07af4211/src/comp/IStateLoc.hs#L395

I guess it is a bit more complicated than I thought, since it is two underscores if it is a "keep" Id and one _ if it is not. In any case, I think we can match it, but I'm not sure if we can get a keep attribute on a function parameter name or not (to test the behavior). Do you know if we can?

nanavati:
Did you try putting the (* keep *) attribute in the source code. I think you can do that in BSV (at least in certain contexts), I'm just not sure whether or not function arguments are one of those contexts.

quark17:
The BSV parser only allows the "keep" attribute on variable declarations in an imperative block.  But those attributes are meaningless, because the parser keeps all declared variables, regardless.  (There's a variable `isGlobalKeep` in `CVParser.lhs` which is ORd with any "keep" attribute that's found, and it is currently defined as True.)  There's no reason why we couldn't add support for "keep" attribute on arguments/parameters.  But if names are being kept globally, then we could also just extend that so the BSV parser is keeping argument/parameters names too.

krame505:
Closing in favor of https://github.com/B-Lang-org/bsc/pull/939.

quark17:
This feature could still be useful.  Your original motivation for adding it may now solved by 939, but it's doing something orthogonal to 939.

PR #873, opened 2026-01-19
Fix off-by-one column tracking for $-prefixed identifiers

sequencer:
## Summary

  Fix incorrect column number tracking when lexing $-prefixed identifiers (Verilog system tasks).

  The pattern `('$':x:cs)` consumes two characters (the `$` and the first identifier character `x`), so the remaining input `cs` starts at column `c+2`, not `c+1`.

  ## Problem

  This bug caused incorrect column positions to be reported for Verilog system tasks like `$display`, `$finish`, `$time`, etc. All subsequent tokens on the same line after a `$`-identifier would have their column numbers off by one.

### DIFF
--- a/src/comp/Lex.hs
+++ b/src/comp/Lex.hs
@@ -278,7 +278,7 @@ lx lf f l c (ch:cs) | isDigit ch =
   in
     case span isDigit cs of
     (s', cs') -> lxFloat (ch:s') cs'
-lx lf f l c ('$':x:cs) | isIdChar x = spanId [] (c+1) cs
+lx lf f l c ('$':x:cs) | isIdChar x = spanId [] (c+2) cs
     where spanId r cn (y:cs) | (isIdChar y || y == '$') = spanId (y:r) (cn+1) cs
           spanId r cn cs' =
              let fs = mkFString ('$':x:reverse r)


quark17:
Hi, thank you for contributing!

In the future, it would be immensely helpful if you can provide a working example that demonstrates the problem.  We also need to add a test case to the testsuite, to demonstrate that the change works and test for regression.

I was able to create the following example (after wasting some time trying to make a BSV example until I remembered that `Lex` is only used for BH):
```
package Test where

f :: Bool -> Action
f x = $fwrite(x)
```
The message from BSC is currently this:
```
Error: "Test.bs", line 4, column 13: (T0020)
  Type error at:
    x

  Expected type:
    Prelude.File

  Inferred type:
    Prelude.Bool
```
With your change, the column number for `x` is 14.  That is the correct position, if we consider the first column to be 0 -- offhand I couldn't remember if BH counts columns this way, but I confirm that it does with another example.  So I believe your change is correct.

We can also confirm that the position matches that for functions without the dollar sign:
```
package Test2 where

f :: Bool -> Action
f x = $fwrite(x)

g :: Bool -> Action
g x = dfwrite(x)

dfwrite :: File -> Action
dfwrite x = $fwrite(x)
```
Currently BSC reports column 13 for `f` and 14 for `g`.

I can add a testsuite example, unless you're able to do that?  I think it should go in `testsuite/bsc.syntax/bh/` and there is already a test `LexPos.hs` that is tested in `bh.exp`.  That was different lexer position bug, and unfortunately the file name doesn't indicate it more specifically.  We don't have to fix that now, but can start adding files with better names.  For example, `LexPos_TaskArg.bs`.  I would use `Test2` (above) for the contents, and check for both positions in the `bh.exp` file.

Finally, was that Copilot review initiated by you? or is that something new in the repo settings that was caused on our end?  If it's something you initiated, can you not do that in the future?  If you want to use to Copilot to review your code before submitting a PR, that's fine, but I only need to see the final PR that you submit.  That would save me time and reduce clutter on the PR.

PR #874, opened 2026-01-20
Update git repo check logic in Makefile

Vekhir:
An issue occurs in the case of Arch Linux, if the release archive is downloaded into the same directory as the build instructions. As the latter are versioned with Git, this causes the logic to assume that it was cloned from the bsc git repo.

Checking whether any submodules are registered is slightly more precise than checking for raw git.

If we aren't in a git repository or that repo has no registered submodules, then the build is definitely created from the archive. Running `git submodule update` won't help in that case.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -30,8 +30,9 @@ checkghcpkg=$(if $(shell ghc-pkg list --simple-output $(1)),,$(error GHC package
 $(foreach pkg,$(GHC_PKGS),$(call checkghcpkg,$(pkg)))
 
 # Check all submodules are initialized
-in_git_repo=$(shell git rev-parse --is-inside-work-tree 2> /dev/null )
-ifeq ($(in_git_repo),true)
+submodexists=$(shell git submodule status $(1) 2> /dev/null )
+in_git_repo:=$(foreach submod,$(SUBMODS),$(call submodexists,$(submod)))
+ifneq ($(in_git_repo),)
 define SUBMOD_MSG
 Submodule $(1) missing. Initialize with
     git submodule update --init --recursive


quark17:
Thanks!

Instead of `git submodule`, I suggest `git submodule status $(1)`.  That command will generate output only if the specific submodule exists, regardless of whether any other submodules exist in the repo.  (In your case, there's no submodules, but this way we support someone else who might have put the BSC source into a repo that does have submodules.)

Also, I think you got the conditions backwards?  When the output is `neq` to empty, the submodule exists.  Anyway, I suggest this code (which is only a change of the first two lines, from what's in the repo now):
``` make
submod_status=$(shell git submodule status $(1) 2> /dev/null )
ifneq ($(submod_status),)
define SUBMOD_MSG
Submodule $(1) missing. Initialize with
    git submodule update --init --recursive
endef
checksubmod=$(if $(wildcard $(1)/.git),,$(error $(SUBMOD_MSG)))
else
define SUBMOD_MSG
Submodule $(1) missing.
    This archive was exported from Git without the source files for
    submodules. You will need to acquire those files or, preferably,
    build instead from a clone of the BSC Git repository
endef
checksubmod=$(if $(wildcard $(1)/*),,$(error $(SUBMOD_MSG)))
endif
```

quark17:
> Not sure what's up with the BDW tests.

We added a new feature that tests `bsc` PRs against the user's version of `bsc-contrib` or `bdw`, if they have a fork of that repo with the same branch name.  Because this PR's branch is called `patch-1` and you have a branch `patch-1` in your fork of `bdw`, that was chosen.  This can be seen in the `Download bdw` step of the workflow:
```
Trying: git ls-remote --heads https://github.com/Vekhir/bdw refs/heads/patch-1
Result: 050e76d7e15aad04602f094c5bc3c91e03f8ab9f	refs/heads/patch-1
Checking out user's branch
```
My suggestion is to add a descriptive string to the patch names or delete old branches, to avoid that in the future.  I think GitHub allows us to write text that will show up for users when they compose a new PR, so we could possibly mention this feature there, if that would help.

PR #880, opened 2026-01-31
Fix pModuleParam to parse 'numeric type id' and 'string type id'

npe9:
The module parameter parser (pModuleParam) only used pTypeExpr for the parameter type, which fails when parsing numeric type n or string type s because pTypeExpr consumes numeric as a type variable and then expects type parameters #(...), but the next token is type (SV keyword).

This caused P0005 for modules like: module mkFoo#(numeric type n, numeric type m)(Ifc);

Add an alternative to pModuleParam that parses (numeric|string)? type id, same as pTypedefParam, producing CType with kind KNum, KStr, or KStar.

Fixes parsing of parameterized modules with numeric/string type parameters (e.g., NVMe_Controller, SATA_Controller).

### DIFF
--- a/src/comp/Parser/BSV/CVParser.lhs
+++ b/src/comp/Parser/BSV/CVParser.lhs
@@ -25,6 +25,7 @@
 > import PreStrings
 > import PreIds
 > import Type
+> import CType(cTVarKind, Kind(KNum, KStr))
 > import IntLit
 > import Error(internalError, EMsg, WMsg, ErrMsg(..),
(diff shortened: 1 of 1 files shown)


quark17:
Oh, it looks like while I was typing that, you did add something else to the parsing of module instantiation.  But I'm not entirely following it yet.  If you can provide an example of both sides, that would help.  (And if you've explored any examples of what the error messages look like if you get the type/kind wrong, that'd be cool to see, too.)

PR #882, opened 2026-02-08
Added logic for named action.

jahagirdar:
Added support for Named actions in FSM's
[CODE.ZIP](https://github.com/user-attachments/files/25161841/CODE.ZIP)

For e.g the Attached code when compiled generates
WILL_FIRE_RL_fsm_dataPath_action
WILL_FIRE_RL_fsm_topSeq_idle

instead of
WILL_FIRE_RL_fsm_action_l17c28,
WILL_FIRE_RL_fsm_idle_l19c19

### DIFF
--- a/src/Libraries/Base2/StmtFSM.bs
+++ b/src/Libraries/Base2/StmtFSM.bs
@@ -51,7 +51,8 @@ getPIPosition l = getStringPosition( getPIString l)
 
 addPIPrefix :: String -> PosInfo -> PosInfo
 --addPIPrefix x (l,p) = ((x +++ l), p)
-addPIPrefix x y = (x +++ y)
+-- addPIPrefix x y = (x +++ y)
+addPIPrefix x y = if (x == "") then y else (x +++ "_" +++ y)
 
 data Freedom
         = Early String -- string is comment for warning message
@@ -257,6 +258,9 @@ getStmtTPosInfo (SWhile p _ _ _ _ _) = p
 getStmtTPosInfo (SRepeat p _ _) = p
 getStmtTPosInfo (SDelay p _) = p
 getStmtTPosInfo (SFor p _ _ _ _) = p
+getStmtTPosInfo (SNamed p _ _) = p
+getStmtTPosInfo (SLabel p _ _ _) = p
+getStmtTPosInfo (SJump p _) = p
 getStmtTPosInfo _ = error "unhandled case"
 
 
@@ -958,6 +962,10 @@ getNoActionCondition (SPar _ Nil) = return True
 getNoActionCondition (SPar _ (Cons st Nil)) = getNoActionCondition st
 getNoActionCondition (SPar _ _) = return False
 getNoActionCondition (SFor _ (SAction _p_init _a_init _) _ _ _) = return False
+getNoActionCondition (SNamed _ _ (Cons st Nil)) = getNoActionCondition st
+getNoActionCondition (SNamed _ _ Nil) = return True
+getNoActionCondition (SLabel _ _ _ Nothing) = return True
+getNoActionCondition (SLabel _ _ _ (Just st)) = getNoActionCondition st
 getNoActionCondition st =
     do x <- stmtTToString st
        messageM ("Case: " +++ x)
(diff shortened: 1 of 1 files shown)


quark17:
Thanks!  For context for readers, this comes from a discussion thread on the developer mailing list: https://groups.io/g/bsc-dev/topic/81853144 (if you skip down to the more recent posts in the thread)

I'm trying to understand what these commits do.  It's purely a change to the `StmtFSM.bs` library, and not anything in the parser.  It doesn't make any change to the `StmtT` datatype to support names -- it turns out that the library already has constructors `SNamed` and `SLabel` (and converted `SFNamed` and `SFLabel`).  Your attached BSV program creates `SNamed` manually.  But it turns out that the BSV parser already supports syntax for specifying names -- have you tried using that?  The BSV Language Manual doesn't document it though (and doesn't offer an alternative when it mentions that names are line-column positions), but if you follow the parsing in `src/comp/Parser/CVParser.lhs`, you can see that the following syntax creates `SNamed`:
```bsv
    Stmt y = seq : dataPath
              something(a);
             endseq;

    Stmt x = seq : topSeq
               y;
             endseq;
```
I think `SLabel` is only created by the parser inside SVAssertions parsing.  I don't understand what the difference is -- it appears that `SNamed` gives a name to a sequence, while `SLabel` appears to stand alone with Maybe a single statement and may be involved in the labelling process that happens inside the library.

Anyway, I'd suggest trying to use the colon syntax, and we can update the documentation to mention it, if it's not there.  @rsnikhil said he'll take a look -- he's the syntax expert.

Regardless of how `SNamed` is created, something needs to change in the library to use `SNamed` in the naming of the rule.  It's unclear how your changes do that, and whether they are the best approach.

If I run the current BSC on my example using colon name syntax, the names do not appear in the output.  If I try to compile your code that manually constructs `SNamed`, I get a compile-time error due to execution entering a missing case arm in the `StmtFSM` library.

Several of your changes in the PR are adding case arms for `SNamed` and `SLabel` (and in one place, `SJump`?!).  We should only add the arms where they are needed, so I would probably remove the additions for `SLabel` and `SJump`, unless you have an example that triggers those?  I would also investigate if all the arms for `SNamed` are needed.

The remaining changes are to `addPIPrefix` (to not prepend an underscore when the prefix is empty) and `ruleName` which creates the rule name.

Unfortunately, the `StmtFSM.bs` library is in serious need of comments, explaining the datatypes and functions.

The function `ruleName` takes three arguments, `text`, `n`, and `suffix`, and the previous behavior was to ignore `n` and just append `suffix` to `text`.  Your PR changes this to return `text` if the suffix is empty and otherwise return `suffix_text`, but when `suffix` is no-position it is replaced with `Ln`.

So it's unclear to me whether the changes to `addPIPrefix` and `ruleName` are actually responsible for the name appearing in the rule.  And, in any case, they should be made to work with the colon syntax -- your hand construction of `SNamed` may be causing an error merely because your construction doesn't adhere to the invariants assumed by the `StmtFSM` library.  Unfortunately, that means a little more work to understand what's going on.  It looks like your changes didn't break anything in the testsuite, though, but offhand I don't know if that's just because we're not testing enough -- I could easily believe that we're not testing the names in generated Verilog for `StmtFSM`.

An eventual PR will need testcases added -- if you're able to add that, that'd be helpful, but I can do that work, if you can provide the examples.

In the future, though, it'd be helpful to just show the BSV code rather than attaching it.  I can't always download (and unzip!) when viewing PRs, so it can delay a response.  I assume that you zipped, because you included the generated Verilog -- showing just the relevant lines from the Verilog is enough.

For convenience here's the source file you posted:
```bsv
import StmtFSM::*;

// Helper to create a named action from an action function call
function Stmt namedCall(String seqName, Action act);
    return _s__(SNamed(seqName, seqName, Cons(SAction(seqName, act, Nothing), Nil)));
endfunction

(*synthesize*)
module [Module] mkTest (Empty);
    Reg#(Bit#(4)) a <- mkRegA(0);
    
    function Action something(Reg#(Bit#(4)) r);
        return (action 
            r <= r + 1;
            // ... more complex stuff
        endaction);
    endfunction
    
    // Use the helper - call the function and wrap with name
    Stmt y = namedCall("dataPath", something(a));
    
    Stmt x = _s__(SNamed("topSeq", "topSeq", Cons(SExprS(noPosInfo, y), Nil)));
    
    FSM fsm <- mkFSM(x);
    
    rule start;
        fsm.start();
    endrule
endmodule
```

PR #883, opened 2026-02-09
Add Functor and Applicative type classes

### LINKED ISSUE #878: Adding a Functor type class (and Functor-Applicative-Monad hierarchy)
Our company's code base has grown numerous collection-ish types that could be instances of `Functor`, but not `Monad`.  Since there isn't a `Functor` type class, we now have a number of inconsistently-named `map` functions for these types.  It would be very helpful to add one.  

One issue is that `fmap` is already defined, using the `Monad` type class.  Making `fmap` member of `Functor` instead would be a breaking change for any existing types with a `Monad` instance.  However, I suspect that the number of `Monad` instances "in the wild" outside of our codebase is likely small, and writing `Functor` instances is not very difficult for someone who can write a `Monad` instance.  

In fact if we are making this change, I would suggest adopting the `Functor`/`Applicative`/`Monad` hierarchy, mirroring Haskell.  This would also let us implement [applicative-do desugaring](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/applicative_do.html) at some point, which would enable some performance improvements for any types that have an implementation of `fmap`/`<*>` that is more efficient than the one using `>>=`.

krame505:
Fix #878. 

TBD on if we should also add the applicative operators (`<*>`, `<$>`, etc.) that Haskell provides.

### DIFF
--- a/testsuite/bsc.misc/lambda_calculus/lc-sysMethod_Split.out.expected
+++ b/testsuite/bsc.misc/lambda_calculus/lc-sysMethod_Split.out.expected
@@ -24,37 +24,31 @@ dim_sysMethod_Split =
 meth_m_sysMethod_Split :: MOD_sysMethod_Split -> (Bool, MOD_sysMethod_Split, Bit #8);
 meth_m_sysMethod_Split =
     (\ (state0 :: MOD_sysMethod_Split) ->
-     let { (def_c__h114 :: Bit #1) = meth_read_RegUN (inst_c__sysMethod_Split state0) }
-     in if (bitToBool def_c__h114)
-	then let { (def_x__h210 :: Bit #8) = meth_read_RegUN (inst_rg1__sysMethod_Split state0)
-		 ; (def_y_avValue__h209 :: Bit #8) = primAdd def_x__h210 (1 :: Bit #8)
-		 ; (def_x__h134 :: Bit #8) = primAdd def_x__h210 (3 :: Bit #8)
-		 ; (def_x__h215 :: Bit #8) = meth_read_RegUN (inst_rg2__sysMethod_Split state0)
-		 ; (def_y_avValue__h214 :: Bit #8) = primAdd def_x__h215 (2 :: Bit #8)
+     let { (def_c__h124 :: Bit #1) = meth_read_RegUN (inst_c__sysMethod_Split state0) }
+     in if (bitToBool def_c__h124)
+	then let { (def_x__h204 :: Bit #8) = meth_read_RegUN (inst_rg1__sysMethod_Split state0)
+		 ; (def_x__h202 :: Bit #8) = primAdd def_x__h204 (1 :: Bit #8)
+		 ; (def_x__h154 :: Bit #8) = primAdd def_x__h204 (3 :: Bit #8)
+		 ; (def_x__h231 :: Bit #8) = meth_read_RegUN (inst_rg2__sysMethod_Split state0)
+		 ; (def_x__h229 :: Bit #8) = primAdd def_x__h231 (2 :: Bit #8)
 		 ; (act1 :: (Bool, MOD_RegUN #8, ())) =
-		       meth_write_RegUN def_x__h134 (inst_rg1__sysMethod_Split state0)
+		       meth_write_RegUN def_x__h154 (inst_rg1__sysMethod_Split state0)
 		 ; (guard1 :: Bool) = fst3 act1
 		 ; (state1 :: MOD_sysMethod_Split) = state0 { inst_rg1__sysMethod_Split = snd3 act1 }
 		 }
-	     in mktuple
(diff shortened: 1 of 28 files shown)


quark17:
Is this ready from your end?  There aren't any more typeclass changes that you might want to add?

testsuite/bsc.misc/lambda_calculus/lc-sysMethod_Split.out.expected
```diff
@@ -24,37 +24,31 @@ dim_sysMethod_Split =
 meth_m_sysMethod_Split :: MOD_sysMethod_Split -> (Bool, MOD_sysMethod_Split, Bit #8);
 meth_m_sysMethod_Split =
     (\ (state0 :: MOD_sysMethod_Split) ->
-     let { (def_c__h114 :: Bit #1) = meth_read_RegUN (inst_c__sysMethod_Split state0) }
-     in if (bitToBool def_c__h114)
-	then let { (def_x__h210 :: Bit #8) = meth_read_RegUN (inst_rg1__sysMethod_Split state0)
-		 ; (def_y_avValue__h209 :: Bit #8) = primAdd def_x__h210 (1 :: Bit #8)
-		 ; (def_x__h134 :: Bit #8) = primAdd def_x__h210 (3 :: Bit #8)
-		 ; (def_x__h215 :: Bit #8) = meth_read_RegUN (inst_rg2__sysMethod_Split state0)
-		 ; (def_y_avValue__h214 :: Bit #8) = primAdd def_x__h215 (2 :: Bit #8)
```
quark17, line 33: It's curious that `def_y_avValue__h###` became `def_x__h###`.  I can't work out why.  But I guess it's not important -- just moving from one bad name to a different bad name.  (Although the previous name has a clue that it was due to a use of `return`).

Other places no longer have `x`, which is great, and is due to an `x` argument in the Prelude being removed (eta reduction).  But I don't see an `x` that was introduced that would cause this.

testsuite/bsc.misc/sal/CTX_sysDynArrSelWithImplCond.sal.expected
```diff
@@ -18,30 +18,31 @@ BEGIN
      #) ;
   
   meth_get (state0 : STATE) : Bit{8}!T =
-    LET def_x_first__h263 : Bit{8}!T = CTX_FIFO2{8}!meth_first(state0.inst_fs_0)
-    IN LET def_x_first__h340 : Bit{8}!T = CTX_FIFO2{8}!meth_first(state0.inst_fs_1)
-    IN LET def_x_first__h417 : Bit{8}!T = CTX_FIFO2{8}!meth_first(state0.inst_fs_2)
-    IN LET def_x_first__h494 : Bit{8}!T = CTX_FIFO2{8}!meth_first(state0.inst_fs_3)
-    IN LET def_x__h572 : Bit{2}!T = CTX_RegN{2}!meth_read(state0.inst_idx)
+    LET def__element_first__h657 : Bit{8}!T = CTX_FIFO2{8}!meth_first(state0.inst_fs_0)
```
quark17, line 21: Here, an `x` has gone, which is cool, but FYI I note that the name `_element` (that `x` is no longer replaceing) isn't from the source code, it's from List functions (like mapM).

quark17 (commented):
Cool.  I've rebased the PR, fixing up conflicts.  And I've fixed the b1894 test, so it doesn't hardcode the def number, so we don't have to update the expected file every time -- instead, I've added a new `regexp_match` function to the testsuite, so the first find can return the def number, to be used in the second find.

This PR doesn't update any docs, but that's OK because Monad etc weren't documented in the libraries (despite it appearing in provisos on many functions -- though it does get a handwavey text explanation in the BSV language guide).  We'll want to add them to the library guide eventually, though.

quark17:
FYI, regarding this commit message: "I don't really understand why these changed, but we also don't really care about these backends."  The name change is not specific to those dumps -- what you're seeing are the names after elaboration, and if they change, we should care.  If those are the only tests that changed behavior, it could be that we're lacking any other tests to catch it.  It's possible, though, that we don't have many tests for names just after elaboration and only tests for final Verilog output, where the change may not be visible -- it's possible that these names are tagged as coming from the prelude, and are therefore considered "bad" and inlined or replaced with other names in the final output (but plenty of Prelude names do make it to output Verilog).

In the libraries, we often use underscore and hide-pragma to hide names of submodule instantiations.  But named arguments to functions also get preserved, and that's often the source of things like `x` in Verilog signal names.  We really should understand any change we make to those names in the Prelude -- your changes look OK (even if I can't explain why a new `x` appeared).  I'm not sure there's been a good audit of the names there, though, which we should do at some point, and also look into whether we need equivalent underscore/pragma means of hiding those names or something.

PR #884, opened 2026-02-10
Fix bsc2bsv to produce valid BSV for several constructs

matx-xida:
## Summary

This PR fixes several issues where `bsc2bsv` generated invalid or semantically incorrect BSV output, improving the fidelity of BH/Classic to BSV conversion.

### 1. Interface method arguments preserved
Methods with pragmas (like `always_ready`) but no explicit `arg_names` lost their argument types. Fixed by falling back to synthetic argument names (`x1`, `x2`, ...) when `filterIArgNames` returns empty.

### 2. Point-free function arguments preserved and applied
Functions defined point-free lost their argument types, and the synthetic arguments were not applied to the body. Fixed by generating synthetic args in the signature AND applying them to the body expression (flattening into existing `CApply` when possible).

### 3. Primed identifiers transformed
BH/Classic allows primed identifiers (`x'`, `x''`) but BSV does not. Fixed by transforming `'` to `_prime` in identifier output.

### 4. Non-builtin operators converted to function calls
User-defined operators like `+++` (string concatenation) were output as infix, but BSV only supports builtin operators as infix. Fixed by converting non-builtin operators to function call syntax using a mapping table.

### 5. Register writes handled correctly
Register writes (`:=` in BH) were incorrectly wrapped in `return()` in method bodies, and parenthesized in `Cletrec`/`Cletseq` final position, causing BSV to parse `<=` as comparison instead of register assignment. Fixed by:
- Detecting `:=` in all AST forms (`Cwrite`, `CBinOp`, `COper`) via `isRegWrite` helper
- Wrapping register writes in `action...endaction` in method bodies
- Skipping parentheses for register writes in `Cletrec`/`Cletseq` final position
- Adding a `ppBody` case for register writes in recursive let handling

## Test Plan

Added test cases in `testsuite/bsc.syntax/bsc2bsv/`:
- `IfcMethodArgs.bs` — interface method argument preservation + register write in methods
- `PointFreeFunc.bs` — point-free function argument preservation and application
- `PrimedIdent.bs` — primed identifier transformation + register write in rules
- `StringConcat.bs` — non-builtin operator conversion

Each test verifies that `bsc2bsv` output compiles successfully with `bsc`.

## Files Changed

- `src/comp/CVPrint.hs` — BSV pretty-printer fixes
- `src/comp/IdPrint.hs` — Identifier printing (primed identifier fix)
- `testsuite/bsc.syntax/bsc2bsv/*` — New test cases

### DIFF
--- a/src/comp/CVPrint.hs
+++ b/src/comp/CVPrint.hs
@@ -103,6 +103,17 @@ opToFuncName ">>>" = "compose"      -- function composition
 opToFuncName "<<<" = "composeR"     -- reverse composition
 opToFuncName s     = s              -- fallback: use operator as-is (may fail)
 
+-- Check if an expression is a register write statement
+-- In bsc2bsv flow, := stays as COper/CBinOp (ParseOp is not run)
+-- In full compiler flow, it becomes Cwrite
+isRegWrite :: CExpr -> Bool
+isRegWrite (Cwrite _ _ _) = True
+isRegWrite (CBinOp _ i _) = getBSVIdString i == ":="
+isRegWrite (COper ops) = any isAssignOp ops
+  where isAssignOp (CRator _ i) = getBSVIdString i == ":="
+        isAssignOp _ = False
+isRegWrite _ = False
+
 t :: String -> Doc
 t s = text s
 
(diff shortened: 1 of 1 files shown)


quark17:
Hi!  Improving `bsc2bsv` would be great.  Was there a reason this PR was closed?  Are you still working on it internally?

PR #891, opened 2026-03-16
Bluespec Language Server and Bluespec Build Tool

matx-yehowshua:
# Add bs-lsp language server, bbt build tool, bs-docgen, and lib-srcs install

This PR lands several quality-of-life improvements for Bluespec development, all self-contained under `util/` with minimal changes outside it.

## Changes outside `util/`

- `GNUmakefile` — new targets: `install-src`, `install-lsp`, `install-vscode-ext`, `install-bsc-docs`; `release` target extended to clean up after itself
- `src/Libraries/Makefile` / `common.mk` — new `install` step copies `.bs`/`.bsv` source files into `inst/lib-srcs/Libraries/` so `bs-lsp` can reference stdlib symbols at runtime
- `README.md` — advertises the new tools

## `util/lsp` — Bluespec Language Server (`bs-lsp`)

A full LSP server for `.bs` and `.bsv` files. Ships with:

- A hand-rolled parser/lexer (wrangling `src/comp` into an LSP-friendly library was not feasible)
- Diagnostics, hover type hints, go-to-definition, completions, and document symbols
- Cross-file and cross-stdlib resolution — type hints and go-to-def trace all the way into the installed stdlib sources
- VS Code extension (`vscode-bluespec`) and Zed extension (`zed-bluespec`)
- Project-aware source scoping via `bbt lsp-info` when a `bsc.toml` is present

## `util/bbt` — Bluespec Build Tool

A Cargo-style project tool. `bsc.toml` declares source directories, build targets, and sim targets. `bbt build`, `bbt sim`, `bbt new`. Also exposes `bbt lsp-info` for the language server to discover project sources without full-workspace scanning.

## `util/docgen` — `bs-docgen` HTML documentation generator

The repo's TeX sources use custom macros that trip up generic converters like pandoc. `bs-docgen` is a purpose-built tool with a hand-rolled TeX parser that handles those macros and uses the same BS/BSV parser as `bs-lsp` to produce comprehensive HTML documentation from the BSC sources.

You can find deployed docs [here](https://bluespec.dev/).


quark17:
Very cool.  I apologize that I may not be able to review soon.  I'll see if I can get others to have a look.

FYI, we are in favor of replacing the LaTeX documentation with something that can be better a common source for HTML, PDFs, etc.  In fact, @rsnikhil is planning to rewrite the language guides in Typst and not just convert them, but rewrite them to be more friendly to new users.

Also, I would ignore the "source doc comments" in libraries, as they are probably out of date.  They are not used at all to generate the PDF documentation.

Also, FYI, there was a `build` tool back when BSC was proprietary that would take a file of `key: value` pairs (specifying flags, top modules, backend targets, etc) and allow you to build various targets.  It was a python script, and we could probably resurrect it, if there's interest.  It was mostly useful for push-button building of designs using proprietary BSC's co-emulation and FPGA board support, since the tool knew all the flags needed and could invoke other tools besides BSC (including the FPGA tools), which is probably why we didn't include it in open BSC.

PR #893, opened 2026-03-20
Instance reduction performance improvments

krame505:
Two main changes here:
* Add a sanity check to determine if an instance can possibly match before attempting to unify in `reducePred` - reject instances that have conflicting type constructors in the instance head in positions that are not determined by functional dependencies.
* Sort preds in `satMany` by the number of type variables they contain, trying to resolve more concrete ones first.  This means we waste less time trying to reduce preds like `Bits a n` when a subsequent pred will determine `a`.  

I haven't done detailed profiling of these changes on the testsuite yet, but in in refactoring our company's code base to use type functions, this yielded a ~30x performance improvement.

The change in `satMany` does cause some error messages to change, due to preds that failed to reduce now appearing in a different order.  Also strangely, it also makes `testsuite/bsc.typechecker/typeclasses/examples/pipeline/Tb1.bsv` compile, where this test previously gave a somewhat mysterious error about ambiguous type variables.  There is a comment there claiming that there is a type class instance that is somehow violating fundep requirements and isn't caught by bsc, so there seems to be an unrelated bug that is now being somewhat mitigated with this change.


krame505:
Yeah, go ahead and merge.

PR #902, opened 2026-04-03
Stop generalizing untyped lets by default

nanavati:
The official argument is here: https://simon.peytonjones.org/let-generalised/

The Bluespec-specific detail is that `let (x, y) = split z` is currently a footgun because a generalized binding lets  `x` and `y` be used inconsistently so the observed sizes of `x` and `y` may not add up to the size of `z`.

I've added a flag, turned it off by default and adjusted the library and testsuite to handle this and the change was relatively low impact. The commonest idiom that broke was using an implicit `fromInteger` at different types and it was not very common.


src/comp/Flags.hs
```diff
@@ -153,7 +153,8 @@ data Flags = Flags {
         verilogFilter :: [String],
         warnActionShadowing :: Bool,
         warnMethodUrgency :: Bool,
-        warnUndetPred :: Bool
+        warnUndetPred :: Bool,
+        letGen :: Bool
```
quark17: Can you add this in its alphabetical location?  And also in two places in FlagsDecode.  And update the expected output in `bsc.options`.

src/comp/GenABin.hs
```diff
@@ -557,7 +557,8 @@ instance Bin Flags where
                 a_100 a_101 a_102 a_103 a_104 a_105 a_106 a_107 a_108 a_109
                 a_110 a_111 a_112 a_113 a_114 a_115 a_116 a_117 a_118 a_119
                 a_120 a_121 a_122 a_123 a_124 a_125 a_126 a_127 a_128 a_129
-                a_130 a_131 a_132 a_133) =
+                a_130 a_131 a_132 a_133
+                a_134) =
```
quark17: Here, and in two more places in this file: Continue on the same line rather than start a new line.

src/Libraries/Base2/CompletionBuffer.bsv
```diff
@@ -43,7 +43,13 @@ endfunction
 module mkCompletionBuffer (CompletionBuffer#(n, a))
   provisos (Bits#(a, sa), Log#(n, ln), Add#(1, ln, ln1));
 
-    let hi = fromInteger(valueOf(n) - 1);
+  // Used to be: let hi = fromInteger(valueOf(n) - 1);
```
quark17: Not sure this comment needs to be this wordy (and Cntrs doesn't have a comment), but OK.

quark17 (commented):
Looks good!  My only request is to insert the new flag alphabetically.  I can push those changes, if you need?  Doesn't look like separate commit history is needed, so we can just squash merge?

rsnikhil:
Does 'off by default' mean 'letGen = False', i.e., don't generalize local lets, the new behavior?

A: 'letGen=False': seems more sensible going forward (as Vytiniotis et. al. argue for ghc), but is less backward-compatible--some old B-Lang programs may not typecheck anymore.

B: 'letGen=True' as a default would be more backward compatible, but one will now want to/have to add this explicitly in future builds.

Are we ok with the (hopefully small) non-backward compatibility of (A)?

(Similar question to 'aggressive-conditions' which, not being the default, has to be added explicitly in builds)

rsnikhil:
I've also reviewed the code changes and they look ok to me (and I agree with comments by @quark17 )

I was surprised (and pleased) to see that removing the generalization required only one very small, local change (TCheck.hs L2914-ish), and of course changes in cmd-line flag management and corresponding fixups in libraries, testsuite, etc.)

In GenABin.hs L37 the ba file tag changed from bsc-ba-20230831-1 to bsc-ba-20260402-1
Is that because of this PR, or something else?

quark17:
The ba-file header needs to change because a new flag was added, which changes the binary encoding of the Flags datatype in the ba-file.

The proposal is that "letGen=False" be the default.  Some designs that used to compile will now result in a type error.  The preferred way of resolving this is by adding an explicit type to the local definitions that need to be generic (as we did for Toooba).  For users that cannot change their code, the flag `-let-gen` can be added to the BSC command line to revert to the old behavior.

Ravi's PR adds type signatures in the BSC libraries where needed.  For the testsuite, it just gives the `-let-gen` flag.  This has the benefit of testing that the flag works.

While it may mean additional verbosity for some bindings, other users that had to insert extra lines of code and type signatures when using functions like `split` will no longer have to add that extra verbosity.  And, of course, won't silently accidentally mis-use `split`.

In theory, we could add a pragma or new syntax to specify when you want a `let` to generalize (say, with a keyword `letgen`), but that additional complexity is probably not worth it.

quark17:
Also, I'm hoping that we can change `-aggressive-conditions` to be the default before the next release.  (And also `-sched-conditions`.)  It might be good to make all of these global changes in one release.

quark17:
FYI, @nanavati, this PR was submitted from your `main` and not from a branch.  I've rebased and pushed a new version, but be careful not to sync your fork until this PR is merged.

krame505 (dismissed):
Nice, that was a very straightforward change.

quark17:
Apologies, my rebase didn't include fixes for the testsuite.  I've pushed that, plus the changes I suggested in my comments.  So it's ready to merge (once green), in my opinion, whenever we're ready.

rsnikhil:
Re. making -no-let-gen, -aggressive-conditions, -sched-conditions be the default before the next release, I'm for that. We should highlight this in the release notes, since they are not backward-compatible.

quark17 (approved):

nanavati:
I agree this is ready-to-go. I'm looking forward to not having to mention the `split` footgun while teaching :)

PR #903, opened 2026-04-05
Fast instance lookup

nanavati:


quark17:
Clever!  I had a look over, and superficially it's all good.  To comment on deeper level, I'll need to get my head around exactly what it's doing (and how multiple dependencies are handled).

src/comp/MakeSymTab.hs
```diff
@@ -54,6 +55,11 @@ doTraceKI = "-trace-kind-inference" `elem` progArgs
 doTraceOverlap :: Bool
 doTraceOverlap = "-trace-instance-overlap" `elem` progArgs
 
+-- Use the legacy flat-list instance index (pre-trie).
+-- Pass -legacy-inst-index to bsc to select this path, e.g. for comparison.
+useLegacyInstIndex :: Bool
+useLegacyInstIndex = "-legacy-inst-index" `elem` progArgs
```
quark17, line 62: Is this something you want to leave in?  Any thoughts on whether it should be a trace flag or a real flag?

src/comp/PredTrie.hs
```diff
@@ -0,0 +1,217 @@
...
+import Pred(Pred(..), Class(inputPositions), expandSyn)
+
+-- ---------------------------------------------------------------------------
+-- PredTrie
+--
+-- A nested-map trie indexed by the head TyCon (via leftTyCon) of each
+-- type argument.  Each level corresponds to one type argument position.
+-- Keys in the map are Maybe TyCon:
+--   Just tc  -- this instance has constructor tc at this position
+--   Nothing  -- this instance has a type variable at this position
```
quark17, line 33: Are these the only options?  Can `m Bool` exist here, say, which has neither a concrete left type constructor nor is a variable?  I assume that you wouldn't want a bound variable to match `m Bool`, so do you need a third option here? Or maybe use `Either TyCon Bool` to capture whether the non-tycon case is a pure variable?

testsuite/bsc.typechecker/instances/TypeFunInst.bs
```diff
@@ -0,0 +1,56 @@
...
+
+-- viaTypeFun: the body uses `0 :: Bit (TAdd n 1)` where n is abstract.
+-- BSC type-checks this with the predicate SizeTag (TAdd n 1).
+-- mkQueryElem sees TAdd n 1: leftTyCon = Just TAdd_tc, isTFunTyCon → Free.
+-- Free probes all branches; byInst fails for concrete 8 and 16
+-- (cannot match TCon 8 or TCon 16 against TAdd n 1).
+-- The catch-all matches (n_fresh → TAdd n 1) and incoherent resolution
+-- commits to it at definition time → always returns 0.
+viaTypeFun :: Bit n -> Bit 8
+viaTypeFun _ = sizeTag (0 :: Bit (TAdd n 1))
```
quark17, line 43: Is this Issue #890 / PR #916?  In that, the use of `sizeTag` here should require a `SizeTag (TAdd n 1)` context on `viaTypeFun`, which would delay the resolution until the call, where it could pick up the specific instance and not the catch-all.

quark17 (commented):
I have one concern, that a left tycon and a variable are not the only two options.

Otherwise, it seems OK, but I admit that I haven't fully understood it thoroughly. My understanding is that, instead of returning the full list of instances for `reducePred` to test, you're now only returning a list that has a chance of matching -- and thus why, in `TCMisc`, you only call `canMatch` if using the legacy method.  Further, you've replaced how instance ordering (and overlap check) was done with a trie, to be more efficient: The edges of the trie are Just a left constructor or Nothing; you can walk the trie testing whether a given pred matches (allowing bound vars to only match variables).  Do I need to understand anything beyond that?

I said "allowing bound vars to only match variables", but that's only if a set of bound vars is passed in to the query function -- do you ever do that anywhere?  I see that `getCls` constructs the `genInsts` using an empty set, because of some incoherent tuple classes.  Is the intention that someday that could be cleaned up to use the bound vars for most typeclasses, when possible, but it currently isn't?

Also, numeric type functions are treated as Nothing (not Just tycon)?  Does that have the same bound-vars issue?

src/comp/MakeSymTab.hs
nanavati: I intend it for benchmarking (right now) and for debugging if we discover an instance-matching issue, so a long-term trace flag seems appropriate to me.

nanavati:
> I have one concern, that a left tycon and a variable are not the only two options.
> 
> Otherwise, it seems OK, but I admit that I haven't fully understood it thoroughly. My understanding is that, instead of returning the full list of instances for `reducePred` to test, you're now only returning a list that has a chance of matching -- and thus why, in `TCMisc`, you only call `canMatch` if using the legacy method. Further, you've replaced how instance ordering (and overlap check) was done with a trie, to be more efficient: The edges of the trie are Just a left constructor or Nothing; you can walk the trie testing whether a given pred matches (allowing bound vars to only match variables). Do I need to understand anything beyond that?
> 
> I said "allowing bound vars to only match variables", but that's only if a set of bound vars is passed in to the query function -- do you ever do that anywhere? I see that `getCls` constructs the `genInsts` using an empty set, because of some incoherent tuple classes. Is the intention that someday that could be cleaned up to use the bound vars for most typeclasses, when possible, but it currently isn't?
> 
> Also, numeric type functions are treated as Nothing (not Just tycon)? Does that have the same bound-vars issue?

The Bound alternative is for future extension. It doesn't matter for matching instances in byInst. It will matter when we start using the trie to match things in the given pool (EPreds) during satisfy - which should be extended with all the things we've deduced but isn't for a variety of reasons including performance (a linear scan is fine for a small number of things the user typed, but not for the pool of everything we've deduced).

testsuite/bsc.typechecker/instances/TypeFunInst.bs
quark17: Er, nevermind, it's probably not related to those.  It's the issue that, for overlapping instances, there's a difference in meaning depending on whether you add an explicit context (which says that resolution should be deferred) or don't (the resolution occurs with what's available, using the less-specific instance).

Are both options being tested?  If not, is it worth adding a second version of `viaTypeFun` but with a context, and showing that equivalent calls to it result in different behavior?

Separately, and perhaps beyond the scope of this PR: Should BSC warn when an instance is resolved locally that could have been deferred? (That is, the match uses bound variables in locations of overlap.)

nanavati: Honestly, what I think should be happening is bsc committing to every coherent instance match to avoid wasted work (though there is the subtlety that it can't really commit if any of the contexts it needs are matched incoherently).

So I'd like to see the backoff / retry behavior limited to incoherent matches (and I'm less concerned about the incoherent case - whose main utility seems to be limited to Has_tpl_n

On the question of allowing bound variable in matches, I think we should make bound variables rigid so those matches aren't possible in the bad cases, but that is a bigger project.

quark17: Let me try to understand.

There is a difference in meaning between these two functions, that have the same body and base type but one also has an explicit context:
```
viaTypeFun1 :: Bit n -> Bit 8
viaTypeFun1 _ = sizeTag (0 :: Bit (TAdd n 1))

viaTypeFun2 :: (SizeTag n) => Bit n -> Bit 8
viaTypeFun2 _ = sizeTag (0 :: Bit (TAdd n 1))
```
If the typeclass isn't incoherent, then there should be no difference in meaning -- well, one difference: If an instance doesn't exist at the time of this file, then v1 will fail with an error; while v2 will typecheck and either fail or pass later when called, depending on if an instance has been provided then.

Incoherence just adds the further wrinkle that the meaning could change (in the successful resolution cases).   For coherent classes, you're saying that, if an instance exists, then resolution should happen right away even for v2 (perhaps with a warning that the explicit context is unnecessary)?

In any case, as far as this PR is concerned: for any of the test files here (regardless of coherency), do you feel any need to be adjusted to test both forms?

quark17 (approved):

PR #905, opened 2026-04-08
Document v2 install needing --lib option

### LINKED ISSUE #904: --lib option needed for cabal v2-install
For `cabal install` (`v2-install`) to install a library, it needs the [`--lib`](https://cabal.readthedocs.io/en/latest/cabal-commands.html#adding-libraries-to-ghc-package-environments) option. Can we update the install instructions to include this detail.

philderbeast:
Fixes #904.

### DIFF
--- a/.github/workflows/cabal-v2.yml
+++ b/.github/workflows/cabal-v2.yml
@@ -1,0 +1,53 @@
+name: cabal v2-install
+
+on: [push]
+
+jobs:
+  v2-install-file:
+    name: Package environment
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: haskell-actions/setup@v2
+        with:
+          ghc-version: '9.12'
+          cabal-version: '3.16'
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          sudo .github/workflows/install_dependencies_ubuntu.sh
+
+      - name: Write file environment
+        shell: bash
+        run: |
+          cabal install --lib \
+            --package-env=.ghc.environment.x86_64-linux-9.12 \
+            regex-compat syb old-time split strict-concurrency
+
+      - name: Use file environment
+        shell: bash
+        run: |
+          PKGDB=$(cat .ghc.environment.x86_64-linux-9.12 | grep "^package-db" | cut -d' ' -f2)
+          GHC="ghc -package-env=$PWD/.ghc.environment.x86_64-linux-9.12" \
+            GHC_PKG="ghc-pkg --user-package-db=$PKGDB" \
+            make install-src --debug --trace
+
+      - name: Write named environment
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


quark17:
Thanks!  I think we should add a workflow that tests a v2 install, so that we can be sure that the instructions work.  (I thought I had tested it, way back, but seems like I must have done it wrong.)

I'm confused by some of what you have written though, so we might need to re-word or re-format it.  Is all that detail about standard locations and naming conventions needed?  Is it not possible to ignore all that and just have cabal install into a local file or directory of a name of our own choosing?

It might help to know who is the audience for this.  There's a bit of tension between wanting to keep the instructions simple for most users, while supporting people who need something else.  I think we added the v2 possibility, so that people who are packaging BSC for an OS package manager could use it.  (And those people don't need us to explain how to use GHC.)  But if it's not working, that suggests that no one was using the feature?  In which case, we could just leave out those instructions, or refer to another file containing the detailed info, so that this file stays simple?

Also note that the sentence about profiling now comes at the end of the v2 section, but is relevant to both v1 and v2.  We either need to duplicate it in both sections, or move it to a common location.

philderbeast:
> Also note that the sentence about profiling now comes at the end of the v2 section, but is relevant to both v1 and v2. We either need to duplicate it in both sections, or move it to a common location.

I moved this sentence and the introductory ones to a new parent section. Here's an outline of this new layout including the profiling sentence:

```
### Required Haskell libraries

To build a version of BSC that supports profiling, be aware that profiling versions of the
libraries need to be installed.

#### Installing required libraries with `cabal v1-install`

#### Installing requered libraries with `cabal v2-install`
```

philderbeast:
> It might help to know who is the audience for this. There's a bit of tension between wanting to keep the instructions simple for most users, while supporting people who need something else. I think we added the v2 possibility, so that people who are packaging BSC for an OS package manager could use it.

I wrote it for myself and for others wanting to use `cabal v2-install`. I'm familiar with the cabal's v2 commands but have hardly touched its v1 commands and I'd never used package environments before. I'd expect most newer cabal users to be in a similar situation. I was very wary of polluting the global installation.

philderbeast:
> I think we should add a workflow that tests a v2 install

I've done this, adding `.github/workflows/cabal-v2.yml`.

philderbeast:
> In which case, we could just leave out those instructions, or refer to another file containing the detailed info, so that this file stays simple?

I see that installation is missing from the [BSC User Guide](https://github.com/B-Lang-org/bsc/releases/latest/download/bsc_user_guide.pdf).

image

I'm writing some docs for another project and using [Quarto](https://quarto.org/). A nice feature this has is collapsed admonitions for hiding details. Here's a screenshot of this work in progress from https://up-do.github.io/updo/:

image

Maybe we could use something like this and keep the details?

philderbeast:
> keep the instructions simple for most users

To shorten the `v2-install` instructions we could keep only the `--package-env=file` option, the more scrutable option.

Issue #296, opened 2021-01-21
Bluesim binaries exit with status 0 on assert failure


cbiffle:
Bluesim self-checking testbenches using `dynamicAssert` wind up exiting with status 0 whether they pass or fail. This makes them difficult to integrate into conventional test frameworks (or, for that matter, Unix shell scripts) that use exit status 0 to indicate success, and non-zero to indicate failure.

The main workaround I've seen in the field is arranging for a test to print some special string, such as `PASS`, and grepping the output for it. This is not great. It would be preferable to return a non-success exit status on assert failure.

quark17:
You mention specifically Bluesim -- does Verilog sim not have the same problem?

I see that the `Assert` package defines the fail behavior like this:
```
$display (assertMessage "Dynamic" s)
$finish 0
```
The argument to `$finish` is just telling the simulator what info to print, not the error code.  I would guess that the simulator would exit with a success code (whether Verilog or Bluesim).  I would guess that we'd want to change this to use `$fatal`, which does tell the simulator to exit with an error code (of its choosing).  [This is an old package, and I don't know how often people use it, so I wouldn't be surprised if it needs some updating or rethinking.]

However, I do see other issues that would prevent a Bluesim sim from exiting with an error code, even if that change were made.  For one, Bluesim objects are executed from a Bluetcl script (`src/bluetcl/bluesim.tcl`) whose last line is `exit 0`.  The script would need to get an exit code from the `sim run` call and return that.  I don't offhand know if the `sim run` command itself is properly returning the exit code, so that might need adjusting too.  (The Bluetcl issues can be bypassed by compiling the model into an executable with a C++ driver rather than driving from Bluetcl.  See `bsc/util/bsim_standalon/`.)  For two, the implementation of `$fatal` in Bluesim (see `dollar_fatal()` in `dollar_display.cxx`) seems to be using its numeric argument as an exit code, when I believe it should have the same meaning as for `$finish`.

thoughtpolice:
I already work around this using the described method: my simulation bench just emits some unique string like `---- SIMULATION EXIT: 0 SUCCESS` and a harness runs them all and ensures that exists. Personally, I find this to be adequate, but in general *some* method to exit a simulation with a different error code would be useful.

[...]

PR #907, opened 2026-04-09
Return error exit codes for assertion failures

thotypous:
This changes assertion failures to terminate simulation with a non-zero exit status.

`dynamicAssert` and `continuousAssert` now use `$fatal(1, ...)` instead of `$display` followed by `$finish(0)`. That gives assertion failures the right semantics and allows them to produce error exit codes.

For Bluesim, this adds explicit tracking of `$fatal` in the runtime and propagates its status through `bluetcl`, so assertion failures now exit with status `1` there as well, without changing the behavior of normal `$finish`.

For Verilog simulators, this behavior depends on preserving `$fatal` in the generated Verilog, so it requires `-system-verilog-tasks` in addition to `-check-assert`.

The Assert tests were updated accordingly:
- Bluesim now checks that failing asserts exit with status `1`
- Verilog simulation now checks the same under `-check-assert -system-verilog-tasks`

Validation:
- `make install-src`
- `make clean`
- `make -C testsuite TEST_RELEASE="$(readlink -f inst)" TEST_BSC_VERILOG_SIM=iverilog bsc.interra/libraries/Assert/Assert.check`
  validates Bluesim and Icarus
- `make -C testsuite TEST_RELEASE="$(readlink -f inst)" TEST_BSC_VERILOG_SIM=verilator bsc.interra/libraries/Assert/Assert.check`
  validates Bluesim and Verilator

### DIFF
--- a/doc/user_guide/bluetcl.tex
+++ b/doc/user_guide/bluetcl.tex
@@ -676,6 +676,13 @@ executing simulation cycle.\\
 
 \begin{tabular}{|p {1.8 in}| p {3.8 in}|}
 \hline
+{\bf sim} {\bf  isfatal}&Return whether the loaded simulation last
+terminated because of \te{\$fatal}.\\
+\hline
+\end{tabular}
+
+\begin{tabular}{|p {1.8 in}| p {3.8 in}|}
+\hline
 {\bf sim} {\bf  unload}&Unload the current bluesim model.\\
 \hline
 \end{tabular}
--- a/src/Libraries/Base1/Assert.bs
+++ b/src/Libraries/Base1/Assert.bs
@@ -35,8 +35,7 @@ dynamicAssert :: Bool -> String -> Action
 dynamicAssert = if not testAssert then (\ _ _ -> noAction)
                 else (\ b s -> if b then noAction else
                         action
-                          $display (assertMessage "Dynamic" s)
-                          $finish 0
+                          $fatal 0 (assertMessage "Dynamic" s)
                      )
 
 nullModule :: (IsModule m c) => m Empty
@@ -59,6 +58,5 @@ continuousAssert = if not testAssert then (\_ _ -> nullModule)
                          "continuousAssert":
                           when not b ==>
                            action
-                            $display (assertMessage "Continuous" s)
-                            $finish 0
+                            $fatal 0 (assertMessage "Continuous" s)
                      )
--- a/src/bluesim/bluesim_kernel_api.h
+++ b/src/bluesim/bluesim_kernel_api.h
@@ -286,6 +286,9 @@ void bk_stop_now(tSimStateHdl simHdl, tSInt32 status);
  */
 void bk_finish_now(tSimStateHdl simHdl, tSInt32 status);
 
+/* Report a fatal error and end simulation. */
+void bk_fatal_now(tSimStateHdl simHdl, tSInt32 status);
+
 /* Test if $stop was called. */
 tBool bk_stopped(tSimStateHdl simHdl);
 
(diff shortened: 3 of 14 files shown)


quark17:
Awesome!  This would seem to resolve Issue #296.  We might want to update the docs for the Bluetcl `sim` command (in the BSC User Guide) to add the new subcommands.

For system tasks, neither the libraries guide or the BH guide say anything, but the BSV guide has a subsection for system tasks, however the subsection for `$error` etc is commented out.  Uncommenting that might require some work to figure out the current state of `$info`, `$warning`, and `$error` as well.  We might also consider making `-system-verilog-tasks` flag on by default, as I'm not sure that's a concern any more; if necessary, we could unhide the flag.

thotypous:
Thanks! I've now added a small user guide update for the new Bluetcl `sim` subcommands.

For the broader system-task documentation and the possibility of making `-system-verilog-tasks` the default, I wonder if it might be cleaner to treat that as follow-up work in a separate PR, since this one is fairly focused. But I'd be happy to go either way if you'd prefer it here.

quark17:
Agreed.  We can do those other things later.

My only remaining concern is that the first argument to the SystemVerilog task `$fatal` is not an error code, but a verbosity level (0, 1, or 2) like the argument to `$finish`.  If we are adhering to the SV spec, then `bluesim.tcl` should not call `sim exitstatus` but should just return 1 when fatal.  (Also, I'd probably just have a conditional on the return status and not duplicate the call to `sim unload`.)  And I'd change `Assert.bs` to pass 0 as the argument to `$fatal`.

quark17:
We could also rethink the internals of Bluesim, but I don't know if you want to go that far.

Maybe there's no need for an `exit_status` field, if we've got `fatal_called`.  We could also rename the arguments of `dollar_stop` and `dollar_finish` from `status` (which it is not) to something like `diagnistic_level` or `finish_number`.  In theory, `bk_stop_now` and `bk_finish_now` could directly take this argument, or take an exit code, or take no argument.  In theory, `dollar_fatal` could be implemented as just a call to `dollar_error` followed by a call to `dollar_finish`, even -- if we allow `$error` to set a field, that gets used to determine the final exit code.

I don't know if you want to go down any of those routes, though.  Although, I do wonder if `$error` should result in an error exit code of the simulation, when it eventually ends.  (Which might argue for replacing `sim->exit_code` with `sim->errored` that is initially 0 and gets set by `dollar_error` to 1; and then `dollar_fatal` could just call `dollar_error` and `dollar_finish`.  And `bluesim.tcl` can check for `errored` to decide the exit code.)

Also, I realize that `util/bsim_standalone/main.cxx.template` should be updated to match `bluesim.tcl` handling of the exit code.  (I can add that, if you want.)

thotypous:
Thanks, that makes sense.

I've updated the patch to keep the change focused on `$fatal`-based assertion failures:

- `Assert` now uses `$fatal(0, ...)`
- Bluesim now checks whether the simulation terminated via `$fatal` and returns exit code 1 in that case
- I kept the pre-existing `bk_exit_status` machinery untouched and only added the minimal `fatal_called` / `bk_fataled` path needed for this PR
- `util/bsim_standalone/main.cxx.template` was updated to match the same behavior
- the Bluetcl User Guide now documents the new `sim isfatal` subcommand

I also did a quick experiment with a minimal Verilog testbench to check `$error` behavior across simulators: `$error(...); $finish;` exits with status 0 on Icarus, but exits with status 1 on Verilator. So for this PR I kept the scope to `$fatal`, where we can make the behavior consistent across Bluesim, Icarus, and Verilator.

PR #908, opened 2026-04-10
Add a Rep type function for Generic

krame505:
This adds a `Rep` type function, which IMO makes generics-based typeclass implementations easier to read.  The name matches the `Rep` type family used in Haskell's `GHC.Generics` library.

We should probably also add type functions for some of the port-splitting-related utilities, but I will do that in a future PR to avoid merge conflicts with #849.

### DIFF
--- a/src/Libraries/Base1/CShow.bs
+++ b/src/Libraries/Base1/CShow.bs
@@ -56,7 +56,7 @@ instance (CShow a) => CShowTuple a where
   cshowTuple = cshow
 
 -- Default generic instance uses the CShow' type class over generic representations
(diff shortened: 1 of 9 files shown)


quark17 (approved):

nanavati:
@krame505 Looking at this PR it seems we do permit type functions in matching position, they just have to be treated as a variable, typically catch all instances? Was that an intended change or am I misreading the code?

nanavati:
Never mind. I looked more closely and realized the type functions were only in contexts and determined positions. So there's actually no issue.

PR #909, opened 2026-04-10
Expand type functions in valueOf

quark17:
This fixes Issue #890, by expanding type functions inside `valueOf` arguments.

I added tests to `bsc.typechecker/primtcons/`, although it was only testing expansion of `SizeOf` which is no longer a primitive.  I don't know if it's worth merging them with the ATF tests (and whether those should go in their own subdirectory, to not drown out the other tests in `bsc.typechecker/typeclasses/`).  I also included tests for kind-checking inside `valueOf` (that were also mentioned in Issue #890), which isn't technically a "primtcons" either.

### DIFF
--- a/src/comp/TCheck.hs
+++ b/src/comp/TCheck.hs
@@ -404,7 +404,13 @@ tiExpr as td exp@(CApply f@(CVar vo) [e@(CHasType _ t)]) | vo == idValueOf = do
     let vs = nub (tv t)
     bvs <- getBoundTVs
     case vs \\ bvs of
-        [] -> tiApply as td exp f e
+        [] -> do
+                 let base_t = case t of
+                                CQType [] bt -> bt
+                                _ -> internalError "tiExpr ValueOf base_t"
+                 (tcon_ps, _) <- expTFun base_t
+                 (ap_ps, exp') <- tiApply as td exp f e
+                 return (tcon_ps ++ ap_ps, exp')
         v : _ -> err (getPosition exp, EValueOf (pfpString v))
 
 tiExpr as td exp@(CApply f@(CVar vo) [e@(CHasType _ t)]) | vo == idStringOf = do
--- a/src/comp/TCheck.hs
+++ b/src/comp/TCheck.hs
@@ -400,10 +400,18 @@ tiExpr as td exp@(CApply f@(CVar gn) es@[(CVar i)]) | gn `qualEq` idPrimGetName 
   -- need to make sure to qualify the primitive identifier for IConv
   return ([], (CApply (CVar idPrimGetName) es))
 
-tiExpr as td exp@(CApply f@(CVar vo) [e@(CHasType _ t)]) | vo == idValueOf = do
-    let vs = nub (tv t)
-    bvs <- getBoundTVs
-    case vs \\ bvs of
+-- Pseudo-functions on types (valueOf and stringOf) are desugared
+-- into ordinary functions on dummy arguments with explicit types.
+-- This can be typechecked as normal, but we handle them specially
+-- here so that an error can be reported if the type contains any
+-- unknown variables.
+--
+tiExpr as td exp@(CApply f@(CVar pfunc) [e@(CHasType _ t)])
+  | (pfunc == idValueOf) || (pfunc == idStringOf)
+  = do
+      let vs = nub (tv t)
+      bvs <- getBoundTVs
+      case vs \\ bvs of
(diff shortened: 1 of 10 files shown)


nanavati:
Should we make a matching fix inside stringOf? I imagine the cases that trigger it might be even more contrived, but it is a loophole.

quark17:
> Should we make a matching fix inside stringOf

Oh, yes, of course!  In fact, I might merge those two cases, to avoid this kind of omission in the future.  (Only complicated by the fact that there are separate `EValueOf` and `EStringOf`, as opposed to a single error parameterized by the function name.)

quark17:
Ok, I've added the equivalent tests for `stringOf`.  And I merged the typechecking of `valueOf` and `stringOf`.  Hopefully that hasn't made it less readable, but I added a comment, which might help.

quark17:
Actually, I've implemented this wrong.  The arm that handles `valueOf` and `stringOf` is just inserting a check and then proceeds with the regular typechecking of the application.  So if `SizeOf` was failing to be expanded there, that means it's failing to be expanded in any `CHasType`.  And I was able to confirm that, with an example like the following, that also fails to expand `SizeOf`:
```
foo :: a -> String
foo _ = printType $ typeOf $ _ :: (Bit (SizeOf a))
```
So I should revert the change I made for `valueOf` and `stringOf` and instead add the `expTFun` call to the typecheck of `CHasType`.

Also, I discovered that the IExpand internal error only happens if the type doesn't have a Bits instance.  I added deriving of Bits to the test examples, but maybe I should restore them to not having the instance.

quark17:
I also notice that my code doesn't expand synonyms, so the following still fails:
```
type Syn a = SizeOf a

foo :: a -> Integer
foo _ = valueOf (Syn a)
```

But I'm also really confused where expansion is ever happening, because `CHasType` just turns into a let-expression, and `SizeOf` isn't being expanded inside those, because the following doesn't report a Bits proviso being needed for `foo` (and thus internal errors in IExpand):
```
foo :: a -> String
foo _ = let x :: Bit (SizeOf a)
            x = _
        in  printType $ typeOf $ x
```

Instead of any change for `valueOf` or `CHasType`, I think we just need to add something to `tiExpl`.  But what?  I assume we don't want to expand away all the synonyms and ATFs.  Can we expand just to get the additional provisos and just add those to the original type.  @krame505 ?

nanavati:
What happens to `Bit (SizeOf a)` when it is part of an ordinary binding? Is ctxreduce what handles it?

I think what should happen is the following: When you typecheck an explicitly typed binding with a type function that is the expected type that is pushed down. At some point type checking needs to check that type against what is inferred. What should happen at the point is the unification triggers the introduction of an equality that will expand the type function.

I think the thing that goes wrong here is we need to unify against something that *contains* a type function rather than is a type function. So the equality is never tested and the proviso is never generated.

I'm not sure what to do about this. Maybe do something in unification when generating a substitution that contains a type function? I know we wanted to ban substitutions that contained type functions and that didn't work out (@krame505 do you remember why?), but maybe a different adjustment would. Or maybe we need to do something if we trim a substitution where the trimmed part has type functions in it.

quark17:
When you say that you tried to ban substitutions with ATFs, do you mean that you just internal errored or did you try something like: replace the contained ATF with a fresh variable and add a proviso (for the equality of the new variable with the ATF)?

krame505:
The reason synonyms weren't being expanded is that you need to call `expandSyn` on the type before calling `expTFun`.

I think you're right that we need to do something for let expressions, but just expanding everything may not be what we want?  I need to do some thinking about this.  

> What happens to Bit (SizeOf a) when it is part of an ordinary binding? Is ctxreduce what handles it?

Yes, I believe so.

> When you say that you tried to ban substitutions with ATFs, do you mean that you just internal errored or did you try something like: replace the contained ATF with a fresh variable and add a proviso (for the equality of the new variable with the ATF)?

The idea was that we thought that all ATFs should be expanded during unification, such that we never construct a substitution containing an unexpanded ATF.  Thus we wouldn't have to normalize again after applying a substitution.  But I think this required writing contexts in extra places that essentially limited the utility of ATFs to places where we could previously express the same thing with a fundep type class context.

krame505:
I pushed an alternate fix that expands in `tiExpl` in #916.  I think just always expanding in `tiExpl` is the cleanest fix.

quark17:
Closing this as replaced by PR 916.

PR #913, opened 2026-04-13
Don't waste time adding edges to the uugraph when there are unlimited ports

nanavati:
Everything but the vertices will be ignored when scheduling the uses.

### DIFF
--- a/src/comp/RSchedule.hs
+++ b/src/comp/RSchedule.hs
@@ -115,7 +115,7 @@ rSchedule' moduleId rFlag rMaxs areSimult mu@(mId, uses0) =
         rMax = lookupRes mId rMaxs
 
         g :: UUGraph
-        g = uuGraph areSimult uses
+        g = uuGraph areSimult uses rMax
 
         dropEdges =
             case rFlag of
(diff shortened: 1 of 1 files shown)


quark17:
Looks fine.  Are you ready to merge?

quark17:
Wait.  It seems like this could be simplified a lot more, by just having `uuGraph` immediately return nothing if `rMax == 0`; or before even getting that far, have `rSchedule'` check for `rMax == 0` and abort early.

It seems like `uuGraph` just constructs three sets of edges, with three list comprehensions, and in each one you've added a filter `rMax /= 0`, so that an empty list is returned if `rMax` is zero.  That seems like way too far at the leaves to be checking `rMax`.

nanavati:
I didn't notice that. Happy to switch to the simpler version if there's no subtle reason the more complex one is needed.

quark17:
I guess it's not doing much wasted work, since `color` is short-circuited for `rMax == 0`.  Not sure it's worth the mental effort of trying to figure out how to reorganize `rSchedule'` to start with a check of `rMax` and just go straight to that result.  So I'm happy to merge it as it is.

quark17 (approved):

Issue #313, opened 2021-02-03
SizeOf in Vector length parameter confuses GenWrap


quark17:
The following example (which might be related to #311):
```
import Vector::*;

// This one fails
typedef SizeOf#(Bit#(2)) Length;

// This one succeeds
// typedef 2 Length;

typedef Bit#(1) Data;

interface Example;
  interface Vector#(Length, WriteOnly#(Data)) example;
endinterface

(* synthesize *)
module mkExample (Example);
  Vector#(Length, Reg#(Data)) ex <- replicateM(mkRegU);
  interface example = ?;
endmodule
```
has code to define `Length` in two different ways.  When synthesizing (`bsc -verilog`) the code using the `SizeOf` definition, BSC gives this error:
```
Error: "Example.bsv", line 14, column 8: (T0043)
  Cannot synthesize `mkExample': The interface method `example' uses type
  `WriteOnly#(Bit#(1))' which is not in the Bits class.
  During elaboration of the interface method `example' at "Example.bsv", line
  14, column 8.
  During elaboration of `mkExample' at "Example.bsv", line 14, column 8.
```
When just typechecking (without `-verilog`), it compiles.  (And the `typedef` for `Data` is not relevant -- you can inline that and see the same behavior.)

This is an interesting example that arises from the interaction of a number of features.  We may be able to fix BSC to support this, but in the meantime the result is that you should not use `SizeOf` (directly or via typedef) in the length parameter of a Vector of subinterfaces, in the interface of a module that is separately synthesized.

As you know, in BH/Classic syntax, there is no syntactic distinction between a method and a subinterface:
```
interface Ifc1 =
    f :: Vector#(4, Bit#(8))

interface Ifc2 =
    f :: Vector#(4, ReadOnly#(Bit#(8)))
```
[...]

Issue #841, opened 2025-11-10
SizeOf of known types sometimes survives elaboration


nanavati:
[GH841.bs.txt](https://github.com/user-attachments/files/23449788/GH841.bs.txt)

When compiled and synthesized the attached example fails with:
```
Internal Bluespec Compiler Error:
Please report this failure to the BSC developers, by opening a ticket
in the issue database: https://github.com/B-Lang-org/bsc/issues
The following internal error message should be included in your
correspondence along with any other relevant details:
Bluespec Compiler, version 2025.07-25-g2e4030b3 (build 2e4030b3)
chkADef dat :: ABSTRACT:  Prelude.Bit;
dat  = _if_ impl_valid____d1 impl_dat____d2 (_ :: Bit 1);
: (ADef {adef_objid = dat, adef_type = ATAbstract {ata_id = Prelude::Bit, ata_sizes = []}, adef_expr = APrim {ae_objid = Prelude::primIf, ae_type = ATBit {atb_size = 1}, aprim_prim = PrimIf, ae_args = [ASDef {ae_type = ATBit {atb_size = 1}, ae_objid = impl_valid____d1[IdP_from_rhs]},ASDef {ae_type = ATBit {atb_size = 1}, ae_objid = impl_dat____d2[IdP_from_rhs]},ASAny {ae_type = ATBit {atb_size = 1}, ae_val = Nothing}]}, adef_props = []}) ABSTRACT:  Prelude.Bit
 /= Bit 1
```

Digging into the problem, the underlying issue is that there is an unreduced `SizeOf Bool` in the output of elaboration.

quark17:
Any relation to #310, #325, #593, etc?  As 310 mentions, there are places where synonyms ought to be expanded but has been disabled because TLM code was relying on structural matching of synonyms or something.

Also, may I ask again that you just include short examples like that as code blocks, rather than as attachments (that have to be downloaded to be viewed)?

nanavati:
From what I have seen, unreduced `SizeOf` can sneak in from many places, so some sort of big hammer is necessary. I traced one case to an `iGetType` that was trying to figure out the type of an array element, for example.

quark17:
I agree that we should probably go with one of your two PRs, but would you mind having a look at some of the failing examples from other issues that seem related, but are not fixed by the PRs, in case they affect your opinion of what we should do?

There are a number of failing examples and a number of proposed fixes, so I did some analysis just to get a sense of where we are.

[...]

Issue #692, opened 2024-04-16
Type synonyms with phantom parameters can lead to strange behaviors (E.g., compiler hang)


rossc719g:
For example:

```bh
package PhantomAlias where

import Vector

-- Split a vector into many vectors of the same length.
splitN :: (Div len k n, Mul k n len, Add n n__ len) =>
  Vector len a -> Vector k (Vector n a)
splitN v = let getChunk :: Integer -> Vector n a
               getChunk i = takeAt ((valueOf n) * i) v
           in map getChunk genVector

type (PhantomVec :: # -> # -> * -> *) n k t = Vector n t
type PhantomVecReg n k t = Reg (PhantomVec n k t)

-- interface (PhantomVecReg :: # -> # -> * -> *) n k t =
--   _read  :: Vector n t            {-# always_ready #-}
--   _write :: Vector n t -> Action  {-# always_ready, always_enabled #-}

mkPhantomVecReg :: (IsModule m c, Bits t t_sz,  -- Add k k__ n,
                    Div n k w, Div n w k, Mul w k n) => m (PhantomVecReg n k t)
mkPhantomVecReg = module

  inVec :: Vector n (Wire t)
  inVec <- replicateM mkBypassWire

  let splitVec :: Vector w (Vector k t)
      splitVec = splitN $ readVReg inVec

  interface PhantomVecReg
    _read    = concat $ reverse splitVec
    _write i = zipWithM_ writeReg inVec i

mkPhantomVecRegDut :: (IsModule m c) => m (PhantomVecReg 6 2 (UInt 8))
mkPhantomVecRegDut = mkPhantomVecReg
```

As written, compiling with `-g mkPhantomVecRegDut -verilog` this will generate two errors, and then hang the compiler.

If you use the commented out interface (instead of the two type aliases), it will just give errors and not hang.
If you add the missing `Add k k__ n` proviso, it will give an error, and not hang.

I talked it over with @nanavati, and he said that type synonyms with phantom parameters should just be an error flat out, and that since they are not being flagged as an error, the compiler is free to go off in the weeds and get stuck.

Should type synonyms with phantom parameters just be a fatal error?

quark17:
[...]

Issue #312, opened 2021-02-03
Unification of type synonyms


quark17:
I'm opening this bug to record what we know about the issue.  The issue is that we have occasionally relied on the typechecker to structurally unifty type aliases.  That is, if a value has type `T#(x,y)` and its context expects `T#(a,b)`, then we rely on BSC to conclude to that `x==a` and `y==b`.  This first showed up in Bluespec's internal bug 1281, when a Sudoku design created `TSquare#(n)` as an alias to `TMul#(n,n)`.

The discussion on that internal bug 1281 shows that there are type constructors/functions that shouldn't structurally unify.  Unifying `TAdd#(x,y)` and `TAdd#(a,b)` does not imply that `x==a` and `y==b`.  We addressed this by introducing the `NumEq` typeclass, which can be used to delay unification until more information is known.  When we encounter `TAdd#(x,y)` and `TAdd#(a,b)` we can replace them with the variable `v` and add the provisos that `NumEq#(v, TAdd#(x,y))` and `NumEq#(v, TAdd#(a,b))`.

It turns out, though, that structural unification of type synonyms was still happening, and some code relied on it.  I have a patch (see below) which expands all synonyms in the symbol table.  It exposes a few a places that rely on structural unification -- and specifically also on the ability to bind type variables in function signatures when those variables are arguments to type synonym constructors.

Before I get to that, let me just say that this was uncovered when I added a fix for Bluespec internal bug 1720 (recorded here as #310) that involved adding `expandSyn` to the handling of struct field updates.  `expandSyn` is a function that expands type synonyms.  This exposed a problem in code generation where BSC had a mismatch between a value and its context in the number of type variables that are being passed -- because one side had expanded a type synonym that ignored some of its arguments.  An attempted fix for this was to expand type synonyms everywhere (so that the ignored variables were not created on either side), and this exposed the issue.

Then, a fix for Bluespec internal bug 1729 (recorded here as #311) also involved adding `expandSyn` and that again pointed out how fragile the current handling of type synonyms is in the presence of extra expanding of synonyms.

If we can find fixes for #310 and #311 that don't involve expanding synonyms (or find some other clever way to retain type synonym boundaries), then we could resolve those issues while leaving the current support for unifying type synonyms.  But the question is: is that a feature that people should be able to rely on?

One argument against supporting it is that you can define this:
```
typedef TAdd#(a,b) A#(type a, type b);
```
[...]

PR #916, opened 2026-04-13
Fix #890 by expanding type functions in tiExpl

### LINKED ISSUE #311: CtxRed needs to expand more type synonyms
As mentioned in #310, a call to `expandSyn` in `CtxRed` had to be removed because of issues with TLM libraries.  This code:
```
do -- XXX disable expanding of type synonyms until
   -- XXX failures with TLM instances are resolved
   -- XXX (vqs_extra, t1) <- expPrimTCons t0 (expandSyn t0)
   (vqs_extra, t1) <- expPrimTCons t0
```
should be replaced with this:
```
do (vqs_extra, t1) <- expPrimTCons (expandSyn t0)
```

An example can be found at `bsc.typechecker/primtcons/ExpSizeOf_InstancesBaseSyn.bsv`.  Without `expandSyn`, this example fails to resolve a context:
```
Error: "ExpSizeOf_InstancesBaseSyn.bsv", line 6, column 10: (T0031)
  The provisos for this expression could not be resolved because there are no
  instances of the form:
    ExpSizeOf_InstancesBaseSyn::C#(ExpSizeOf_InstancesBaseSyn::S, 12)
```

This is bug 1729 in Bluespec Inc's pre-GitHub bug database.

krame505:
Also adds tests including the ones added by @quark17 in #909.

This does mean that we need to write explicit contexts on top-level functions that involve type functions, but I think that is alright.  

There was some weird inconsistent behavior with type synonyms that expand to type functions in struct fields - see the test case `ExpSizeOf_FieldSyn_Minimal`.  The derived `Generic` instance was failing to type check, but it would pass if I removed the type synonym.  I traced this to `ctxRedCQType'`, where we were only calling `expTFun` without expanding synonyms.  There was a comment here that this is disabled until the failures with TLM instances are resolved.  I'm not quite sure what this was about, but there have been some fixes to type synonyms recently, and changing this back did not seem to break anything, so I am assuming that this is no longer a concern?  

The only other observable change in behavior is that the bluetcl dump of instances now has type aliases expanded in instance heads, which I think is also okay.

### DIFF
--- a/testsuite/bsc.typechecker/primtcons/CExtendSynonym.bsv
+++ b/testsuite/bsc.typechecker/primtcons/CExtendSynonym.bsv
@@ -1,0 +1,46 @@
+// Reduced version of bsc-contrib AMBA_TLM library.
+// Check that we can use type synonyms in instance heads.
+
+function Bit#(m) zExtend(Bit#(n) value)
+   provisos(Add#(n,m,k));
+   Bit#(k) out = zeroExtend(value);
+   if (valueOf(m) == 0)
+      return ?;
+   else
+      return out[valueOf(m) - 1:0];
+endfunction
+
+function a cExtend(b value)
+   provisos(Bits#(a, sa), Bits#(b, sb));
+
+   let out = unpack(zExtend(pack(value)));
+
+   return out;
+endfunction
+
+`define TLM_PRM_DCL numeric type id_size,   \
+                    numeric type addr_size
+
+`define TLM_PRM    id_size,   \
+                   addr_size
+
+typedef Bit#(id_size)     TLMId#(`TLM_PRM_DCL);
+typedef Bit#(id_size)     AxiId#(`TLM_PRM_DCL);
+
+typedef struct {
+                AxiId#(`TLM_PRM)     id;
+                } AxiAddrCmd#(`TLM_PRM_DCL);
+
+function TLMId#(`TLM_PRM) fromAxiId(AxiId#(`TLM_PRM) id);
+   return cExtend(id);
+endfunction
+
+typeclass BusPayload#(type a, type b) dependencies(a determines b);
+   function b getId(a payload);
+endtypeclass
+
+instance BusPayload#(AxiAddrCmd#(`TLM_PRM), TLMId#(`TLM_PRM));
+   function getId(payload);
+      return fromAxiId(payload.id);
+   endfunction
+endinstance
--- a/testsuite/bsc.typechecker/primtcons/CExtendSynonym.bsv
+++ b/testsuite/bsc.typechecker/primtcons/CExtendSynonym.bsv
@@ -1,5 +1,14 @@
-// Reduced version of bsc-contrib AMBA_TLM library.
-// Check that we can use type synonyms in instance heads.
+// Reduced version of the bsc-contrib AMBA_TLM library (see bsc-contrib PR 46).
+// Check that we can use type synonyms in instance heads.  The synonyms 'TLMId'
+// and 'AxiId' drop the 'addr_size' parameter; expanding them fully in the
+// instance head would leave 'addr_size' dangling and break unification.  This
+// exercises the (now conditional) synonym expansion in CtxRed.ctxRedCQType' --
+// see the XXX there and GitHub Issue #311.  Because these synonyms contain no
+// type function, the conditional expansion leaves them alone, so this compiles;
+// bsc-contrib PR 46 added explicit method types to work around the earlier
+// (unconditional-expansion) failure, and those are no longer needed here.
+// CExtendATF.bsv shows a variant that *does* still fail, because it puts a type
+// function in the synonyms and so triggers the expansion.
 
 function Bit#(m) zExtend(Bit#(n) value)
    provisos(Add#(n,m,k));
--- a/testsuite/bsc.typechecker/primtcons/ExpSizeOf_FieldSyn_BS.bs
+++ b/testsuite/bsc.typechecker/primtcons/ExpSizeOf_FieldSyn_BS.bs
@@ -1,0 +1,20 @@
+package ExpSizeOf_FieldSyn_BS where
+
+-- Classic syntax version of ExpSizeOf_FieldSyn.
+-- A struct with a field whose type uses SizeOf through a synonym chain.
+
+type T t = UInt (S t)
+type S t = SizeOf t
+
+struct Node dataType =
+  dat    :: dataType
+  unused :: T dataType
+ deriving (Bits, Eq)
+
+{-# synthesize sysExpSizeOf_FieldSyn_BS #-}
+sysExpSizeOf_FieldSyn_BS :: Module Empty
+sysExpSizeOf_FieldSyn_BS = module
+  r_node :: Reg (Node (UInt 3)) <- mkRegU
+  r_data :: Reg (UInt 3)        <- mkRegU
+  rules
+    when True ==> r_data := r_node.dat
--- a/testsuite/bsc.typechecker/primtcons/ExpSizeOf_Let_NoCtx.bs
+++ b/testsuite/bsc.typechecker/primtcons/ExpSizeOf_Let_NoCtx.bs
@@ -1,0 +1,11 @@
+package ExpSizeOf_Let_NoCtx where
+
+-- A let binding with a type annotation containing SizeOf should require
+-- a Bits context, but currently doesn't (the CHasType path in tiExpr
+-- handles inline annotations but not let bindings).
+-- This is filed as a known bug.
+
+foo :: a -> String
+foo _ = let x :: Bit (SizeOf a)
+            x = _
+        in  printType $ typeOf $ x
--- a/testsuite/bsc.typechecker/primtcons/ExpSizeOf_Let_NoCtx.bs
+++ b/testsuite/bsc.typechecker/primtcons/ExpSizeOf_Let_NoCtx.bs
@@ -1,9 +1,10 @@
 package ExpSizeOf_Let_NoCtx where
 
--- A let binding with a type annotation containing SizeOf should require
--- a Bits context, but currently doesn't (the CHasType path in tiExpr
--- handles inline annotations but not let bindings).
--- This is filed as a known bug.
+-- A let binding whose type annotation contains SizeOf requires a Bits
+-- context: expanding the SizeOf in the declared type introduces an implicit
+-- Bits context, which must be provided by an explicit context on the
+-- enclosing definition.  Here none is given, so typecheck reports the
+-- missing context (T0030).  (GitHub Issue #890)
 
 foo :: a -> String
 foo _ = let x :: Bit (SizeOf a)
--- a/testsuite/bsc.typechecker/primtcons/primtcons.exp
+++ b/testsuite/bsc.typechecker/primtcons/primtcons.exp
@@ -56,5 +56,8 @@ compile_fail_error ValueOf_KindMismatch_Res.bs T0026
 compile_fail_error ExpSizeOf_Let_NoCtx.bs T0030
 compare_file [make_bsc_output_name ExpSizeOf_Let_NoCtx.bs]
 
+# Test that cExtend works when type synonyms are expanded in instance heads
+compile_pass CExtendSynonym.bsv
+
 # ---------------
 
--- a/testsuite/bsc.typechecker/primtcons/primtcons.exp
+++ b/testsuite/bsc.typechecker/primtcons/primtcons.exp
@@ -18,7 +18,8 @@ compile_pass ExpSizeOf_Instances.bsv
 
 compile_pass ExpSizeOf_InstancesBase.bsv
 
-# Test that SizeOf is expanded even when it's buried in a type synonym
+# Test that SizeOf is expanded even when it's buried in a type synonym.
+# This was Bluespec Inc Bug 1729 (also GitHub Issue #311).
 compile_pass ExpSizeOf_InstancesBaseSyn.bsv
 
 # ---------------
@@ -33,31 +34,81 @@ compile_verilog_pass ExpSizeOf_Field.bsv
 # it succeeds.
 compile_verilog_pass ExpSizeOf_FieldSyn.bsv
 compile_verilog_pass ExpSizeOf_FieldSyn_BS.bs
+
+# Minimal reproduction: without expanding the synonym in ctxRedCQType, the
+# derived Generic instance fails to typecheck (the needed Bits context is not
+# injected), so it fails to compile.  A plain compile_pass is therefore a
+# sufficient regression guard -- no intermediate dump is needed.
 compile_pass ExpSizeOf_FieldSyn_Minimal.bs
 
-# ---------------
-# Test that SizeOf is expanded inside 'valueOf' (GitHub Issue #890)
+# Using a struct field whose type uses SizeOf through a synonym requires a Bits
+# context on the struct's type parameter.  The requirement is introduced during
+# typecheck of the field access, so it must be provided (GitHub Issue #890).
+compile_fail_error ExpSizeOf_FieldSyn_Use_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_FieldSyn_Use_NoCtx.bs]
+compile_pass ExpSizeOf_FieldSyn_Use_WithCtx.bs
 
+# A struct field whose SizeOf-using type has a type variable that is not a
+# parameter of the struct is not exempted from the Bits requirement: the field
+# must declare the context explicitly (GitHub Issues #890 and #310).  BSC
+# currently reports two T0030 errors here (one of which lacks a source
+# position), so this uses an expected count of 2 rather than compare_file.
+compile_fail_error ExpSizeOf_FieldSyn_Unbound_NoCtx.bs T0030 2
+compile_pass ExpSizeOf_FieldSyn_Unbound_WithCtx.bs
+
+# ---------------
+# Test that type functions like SizeOf are expanded in explicitly-typed
+# expressions, introducing the implicit contexts they require and reporting an
+# error when those contexts are not provided (GitHub Issue #890).  This covers
+# 'valueOf', 'stringOf', type-annotated expressions, and explicitly-typed let
+# bindings (valueOf and stringOf are themselves implemented using type-annotated
+# expressions).
+
+# valueOf
 compile_fail_error ExpSizeOf_ValueOf_NoCtx.bs T0030
 compare_file [make_bsc_output_name ExpSizeOf_ValueOf_NoCtx.bs]
-
 compile_pass ExpSizeOf_ValueOf_WithCtx.bs
 
-# Perform the same tests for 'stringOf'
+# stringOf
 compile_fail_error ExpSizeOf_StringOf_NoCtx.bs T0030
 compare_file [make_bsc_output_name ExpSizeOf_StringOf_NoCtx.bs]
 compile_pass ExpSizeOf_StringOf_WithCtx.bs
+
+# type-annotated expression, e.g. '(_ :: Bit (SizeOf a))'
+compile_fail_error ExpSizeOf_TypedExpr_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_TypedExpr_NoCtx.bs]
+compile_pass ExpSizeOf_TypedExpr_WithCtx.bs
+
+# explicitly-typed let binding
+compile_fail_error ExpSizeOf_Let_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_Let_NoCtx.bs]
 
 # Kind errors in valueOf type argument
 compile_fail_error ValueOf_KindMismatch_Arity.bs T0025
 compile_fail_error ValueOf_KindMismatch_Res.bs T0026
 
-# SizeOf in a let binding type annotation requires a Bits context
-compile_fail_error ExpSizeOf_Let_NoCtx.bs T0030
-compare_file [make_bsc_output_name ExpSizeOf_Let_NoCtx.bs]
+# ---------------
+# Type synonyms and type functions in instance heads
+# (GitHub Issue #311, bsc-contrib PR 46)
 
-# Test that cExtend works when type synonyms are expanded in instance heads
+# Plain type synonyms in an instance head are left unexpanded, so this compiles
+# without the explicit method types that bsc-contrib PR 46 previously needed.
 compile_pass CExtendSynonym.bsv
+
+# A type function (here a user-defined ATF) in the instance-head synonyms forces
+# the conditional expansion in ctxRedCQType', which re-introduces the dangling
+# parameter and still fails -- unless explicit method types are given.  This is
+# the case that remains broken (GitHub Issue #311).
+compile_fail_error CExtendATF.bsv T0035
+compile_pass CExtendATFExplicit.bsv
+
+# ---------------
+# SizeOf in the length of a Vector of subinterfaces (GitHub Issue #313,
+# Bluespec Inc Bug 1917).  This is not fixed by PR #916: the example typechecks,
+# but fails when synthesized, because GenWrap decides whether the field is a
+# method or a subinterface before typecheck and cannot reduce the SizeOf.
+compile_pass ExpSizeOf_VectorIfc.bsv
+compile_verilog_fail_error ExpSizeOf_VectorIfc.bsv T0043
 
 # ---------------
 
--- a/testsuite/bsc.typechecker/primtcons/primtcons.exp
+++ b/testsuite/bsc.typechecker/primtcons/primtcons.exp
@@ -19,7 +19,7 @@ compile_pass ExpSizeOf_Instances.bsv
 compile_pass ExpSizeOf_InstancesBase.bsv
 
 # Test that SizeOf is expanded even when it's buried in a type synonym
-compile_pass_bug ExpSizeOf_InstancesBaseSyn.bsv 1729
+compile_pass ExpSizeOf_InstancesBaseSyn.bsv
 
 # ---------------
 # Test that SizeOf introduced through the type of a field in a struct value
@@ -32,6 +32,29 @@ compile_verilog_pass ExpSizeOf_Field.bsv
 # but after adding the normtypes pass (later replaced by extra normalization in expanded),
 # it succeeds.
 compile_verilog_pass ExpSizeOf_FieldSyn.bsv
+compile_verilog_pass ExpSizeOf_FieldSyn_BS.bs
+compile_pass ExpSizeOf_FieldSyn_Minimal.bs
+
+# ---------------
+# Test that SizeOf is expanded inside 'valueOf' (GitHub Issue #890)
+
+compile_fail_error ExpSizeOf_ValueOf_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_ValueOf_NoCtx.bs]
+
+compile_pass ExpSizeOf_ValueOf_WithCtx.bs
+
+# Perform the same tests for 'stringOf'
+compile_fail_error ExpSizeOf_StringOf_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_StringOf_NoCtx.bs]
+compile_pass ExpSizeOf_StringOf_WithCtx.bs
+
+# Kind errors in valueOf type argument
+compile_fail_error ValueOf_KindMismatch_Arity.bs T0025
+compile_fail_error ValueOf_KindMismatch_Res.bs T0026
+
+# SizeOf in a let binding type annotation requires a Bits context
+compile_fail_error ExpSizeOf_Let_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_Let_NoCtx.bs]
 
 # ---------------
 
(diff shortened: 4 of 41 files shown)


krame505:
Ah, it's the AMBA_TLM libraries in contrib that are failing with that change.  I'm not sure I understand what is going on here, but adding explicit types seems to fix the failures: https://github.com/B-Lang-org/bsc-contrib/pull/46

quark17:
If I understand, in an example like this:
```
foo :: a -> String
foo _ = let x :: Bit (SizeOf a)
            x = _
        in  printType $ typeOf $ x
```
During typecheck, the use of `x` is not introducing a `Bits` context (because of the `SizeOf`), and that's not bubbling up to the top, where it would be reported as a missing needed context.  Your fix is to require a `Bits` context on `x`, so that the typecheck sees an explicit context and introduces it.  Is there an alternate fix, where uses of `x` expand the type to expose the introduced context?  I guess that potentially repeats work.  And it's probably good to require the user to add explicit contexts -- particularly if the SizeOf is buried inside of type synonyms and not immediately evident to the user.  (Though you said explicit contexts are only required on top-level functions?  Are they not required on explicitly-typed let-bindings?)

As for the XXX comment about TLM:  There were two places in the BSC source that type-alias expansion was being held back, because certain code (like the TLM libraries in bsc-contrib) were relying on the old buggy behavior, of allowing structural unification of synonyms.  I would be happy if we can eliminate those (stop holding back synonym expansion).  It may be that the TLM codes needs to be written different -- in fact I wonder if it could now be written with ATFs.

The buses there have a large set of parameters (the data and address width, the burst size, the user field width, etc) and to avoid typing these params all the type, SV pre-preprocessor macros are used, so that one name can be places where a whole comma-separated list would go.  Type synonyms are then used to get at individual fields -- the synonym takes a large argument list and returns just one of them.  In some cases, though, this caused ambiguity -- possibly because of using `TLog` on the variable, which can't match uniquely from the context.  Perhaps all of this could be better specified with type class dependencies and ATFs.

The XXXs are in two places, one in `TCheck` and one in `CtxRed`.  If you only ran into one of them, then there may be another still lurking? or maybe it was recently eliminated.  For more information see:

- Issue #692 , where I spell out the code in the TLM libraries that are the concern (probably the best place to start and which links to many of the issues below)
- Issue #310 that points to the `TCheck` code, and shows an example that fails without that expansion -- but has been since fixed? (did it reinstate the expansion or just no longer triggers the issue?)
- Issue #311 that points to the `CtxRed` code, and shows an example that fails without that expansion -- which is still open, but presumably resolved by your change?
- Issue #312 that talks about the unification problem generally (and refers to 310 and 311) -- which is still open, but we can re-evaluate if recent changes close it?
- The final paragraph of [a comment here](https://github.com/B-Lang-org/bsc/issues/221#issuecomment-675258013) (on Issue 221) that briefly summarized the XXXs
- [This comment](https://github.com/B-Lang-org/bsc/issues/841#issuecomment-3519677345) (on Issue 841) that lists various synonym issues (including 310, 311, 312, but also others) and the status etc

krame505:
> Though you said explicit contexts are only required on top-level functions? Are they not required on explicitly-typed let-bindings?

Yes, this should require an explicit context for the type in the binding to be written, somewhere.

> As for the XXX comment about TLM: There were two places in the BSC source that type-alias expansion was being held back, because certain code (like the TLM libraries in bsc-contrib) were relying on the old buggy behavior, of allowing structural unification of synonyms. I would be happy if we can eliminate those (stop holding back synonym expansion). It may be that the TLM codes needs to be written different -- in fact I wonder if it could now be written with ATFs.

I tried reverting the other place where expanding type synonyms is disabled in `tiField1`, and that still causes a lot of failures.  

I don't think I have time to keep investigating this right now - feel free to take it up if you'd like, otherwise I'll try to get back to it when I have bandwidth.

quark17:
As mentioned on the associated bsc-contrib PR, the failure there isn't what was referred to by the XXX comments (I don't think).  I've boiled down the new failure to this code (that doesn't need to import anything)
```bsv
// ---
// From BUtils

function Bit#(m) zExtend(Bit#(n) value)
   provisos(Add#(n,m,k));
   Bit#(k) out = zeroExtend(value);
   if (valueOf(m) == 0)
      return ?;
   else
      return out[valueOf(m) - 1:0];
endfunction

function a cExtend(b value)
   provisos(Bits#(a, sa), Bits#(b, sb));

   let out = unpack(zExtend(pack(value)));

   return out;
endfunction

// -----

`define TLM_PRM_DCL numeric type id_size,   \
                    numeric type addr_size

`define TLM_PRM    id_size,   \
                   addr_size

typedef Bit#(id_size)     TLMId#(`TLM_PRM_DCL);
typedef Bit#(id_size)     AxiId#(`TLM_PRM_DCL);

typedef struct {
                AxiId#(`TLM_PRM)     id;
		} AxiAddrCmd#(`TLM_PRM_DCL);

function TLMId#(`TLM_PRM) fromAxiId(AxiId#(`TLM_PRM) id);
   return cExtend(id);
endfunction

// -----

typeclass BusPayload#(type a, type b) dependencies(a determines b);
   function b getId(a payload);
endtypeclass

instance BusPayload#(AxiAddrCmd#(`TLM_PRM), TLMId#(`TLM_PRM));
   function getId(payload);
      return fromAxiId(payload.id);
   endfunction
endinstance
```
With previous BSC, it compiles (`bsc Test.bsv`) with no message.  With this PR, it has the following error:
```
Error: "Test.bsv", line 51, column 14: (T0035)
  Bit vector of unknown size introduced near this location.
  Please remove unnecessary extensions, truncations and concatenations and/or
  provide more type information to resolve this ambiguity.
```
In the example, there are two parameters (the ID size and the Addr size).  The `AxiAddrCmd` struct could have a field for the addr, using that size, but it is not needed to illustrate the problem, so I've removed it.  However, the existence of the Addr parameter is needed to trigger the bug -- if you change the macros to remove it, the example compiles fine with the new PR.

I believe the ambiguous size has nothing to do with ID, but is the Addr size, which is unused when you only extract the ID parameter, but still appears in a type in the code and therefore what we're seeing is BSC has detected a variable that is remaining at the end of typecheck and BSC is reporting an error that it hasn't been resolved.  Presumably, something in this PR changes the final typechecked code, to expose a type.  Maybe expanding a synonym has introduced a variable -- but that doesn't make sense, because expanding `AxiId` and `TLMId` actually _removes_ variables.  But maybe that's the issue: somehow we were unifying `TLMId#(idsz,addrsz)` with `TLMId#(x,y)` and thus finding that `addrsz==y`, but now we're only unifying `idsz` (after expansion) with `x` (after expansion), and so `addrsz` is still dangling somewhere without a match.  Though, the fact that an explicit type makes it go away is odd?

Also, now I wonder if one of `TLMId` or `AxiId` can go away in the example, to make it even simpler -- it may only be one of those that needs the extra parameter.

quark17:
> feel free to take it up if you'd like

I did not have the time to spend on this that I've already spent.  I'm stopping here (and going to try harder not to be nerd-sniped by emails, so I can focus on finishing my degree).

krame505:
> > feel free to take it up if you'd like
> 
> I did not have the time to spend on this that I've already spent. I'm stopping here (and going to try harder not to be nerd-sniped by emails, so I can focus on finishing my degree).

Thanks for taking the time to look into this and find a more self-contained example!  I understand, I'm in the same position with finishing my dissertation, and fixing this isn't currently a blocker issue for us at MatX.

krame505:
It looks like there are also some rather strange testsuite failures from that last commit re-enabling expanding synonyms inside `tiField1`.

krame505:
Okay for the moment, I think the path of least resistance to fixing #890 is to keep the current behavior in `ctxRedCQType'` and only expand synonyms when doing so allowed a type function to be reduced.  The current behavior of not expanding synonyms in instance heads is wrong, but that can be fixed when someone has the time to update the AMBA_TLM libraries to not rely on this.

testsuite/bsc.typechecker/primtcons/ExpSizeOf_Let_NoCtx.bs
```diff
@@ -0,0 +1,11 @@
+package ExpSizeOf_Let_NoCtx where
+
+-- A let binding with a type annotation containing SizeOf should require
+-- a Bits context, but currently doesn't (the CHasType path in tiExpr
```
quark17: I think this comment is out of date and the behavior is now fixed by later commits in this PR?  Can you update the comment?  (Also, in the future, instead of vaguely saying that something is filed as a bug, I think it'd be better to give the actual bug number.)

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -19,7 +19,7 @@ compile_pass ExpSizeOf_Instances.bsv
 compile_pass ExpSizeOf_InstancesBase.bsv
 
 # Test that SizeOf is expanded even when it's buried in a type synonym
-compile_pass_bug ExpSizeOf_InstancesBaseSyn.bsv 1729
+compile_pass ExpSizeOf_InstancesBaseSyn.bsv
```
quark17, line 23: Now that the bug number is removed from the command, can you insert a comment that records that this was Bluespec Inc Bug 1729?  And add that it is also GitHub Issue 311.

testsuite/bsc.typechecker/primtcons/ExpSizeOf_FieldSyn_BS.bs
```diff
@@ -0,0 +1,20 @@
...
+
+-- Classic syntax version of ExpSizeOf_FieldSyn.
+-- A struct with a field whose type uses SizeOf through a synonym chain.
+
+type T t = UInt (S t)
+type S t = SizeOf t
+
+struct Node dataType =
+  dat    :: dataType
+  unused :: T dataType
```
quark17, line 11: I think we should add two variations on this test.  One about the user of the struct and one about the struct itself.

For the user of the struct, the test that I would like to see is this:
```bh
fnNeedsCtx :: Node a -> T a
fnNeedsCtx n = n.unused
```
or even this, if it matters whether T is in the type or not:
```bh
fnNeedsCtx :: Node a -> Bool
fnNeedsCtx n = (n.unused == 0)
```
Because I'd like to test that BSC introduces a Bits context due to the use of the struct field that requires it.  The current file (which we can keep) doesn't use it and, even if it did, doesn't confirm that the handling of the instance occurs during typecheck.  (In the new example, the field should be renamed from `unused` since it's now used.)

For the struct itself, I suggest we test the following example:
```bh
struct S =
  sfield :: T dataType
```
Here, because `dataType` is not bound in the struct type, I would expect that the user should get an error about a missing `Bits` context, and should have to write the following:
```bh
struct S =
  sfield :: (Bits dataType sz) => T dataType
```
I'm guessing that BSC will not complain, and that's because of the XXX in `TCheck.tiField`?  If so, then I'd like us to make sure there's a bug open for that.  Bug #310 was originally about the XXX in `tiField`, but was closed because the specific example started working?  If I've identified another example that hasn't been fixed, then we should re-open #310 (or open a new bug) and put that bug number in comments for the new example.

If I understand, this example (in BSV form) was the attached example for Bug 310.  It has a bound type (which is expected from expansion?).  But we should have an example also for an unbound type (that should not be exempted?).  And also examples that actually use the field.

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -32,6 +32,32 @@ compile_verilog_pass ExpSizeOf_Field.bsv
 # but after adding the normtypes pass (later replaced by extra normalization in expanded),
 # it succeeds.
 compile_verilog_pass ExpSizeOf_FieldSyn.bsv
+compile_verilog_pass ExpSizeOf_FieldSyn_BS.bs
+compile_pass ExpSizeOf_FieldSyn_Minimal.bs
```
quark17, line 42: The comment in the file says that there's a bug (Bits not being included in the Generic instance).  Should this be dumping an intermediate stage and checking for that failure?

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -32,6 +32,32 @@ compile_verilog_pass ExpSizeOf_Field.bsv
...
+
+# ---------------
+# Test that SizeOf is expanded inside 'valueOf' (GitHub Issue #890)
```
quark17: I wonder if the header should be corrected here, since it's really a fix for explicit types, which includes explicitly-typed let bindings, explicit types on expressions (e.g. `pack (x::T)` in BH), and both `valueOf` and `stringOf` which are implemented using typed expressions.  We could even divide this into three sections with their own headers, if you prefer (and just mention GH 890 in all three).

I don't think you have an example of typed expressions?  Can we add one, for completeness (and to not assume anything about the implementation, that some other test will cover it)?

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -32,6 +32,32 @@ compile_verilog_pass ExpSizeOf_Field.bsv
...
+compare_file [make_bsc_output_name ExpSizeOf_ValueOf_NoCtx.bs]
+
+compile_pass ExpSizeOf_ValueOf_WithCtx.bs
+
+# Perform the same tests for 'stringOf'
+compile_fail_error ExpSizeOf_StringOf_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_StringOf_NoCtx.bs]
+compile_pass ExpSizeOf_StringOf_WithCtx.bs
+
+# Kind errors in valueOf type argument
```
quark17, line 92: If we had a separate place for tests for pseudo type functions, we could put these tests there.  But I don't see that we do, so including them here is fine.

testsuite/bsc.typechecker/primtcons/CExtendSynonym.bsv
```diff
@@ -0,0 +1,46 @@
...
+function TLMId#(`TLM_PRM) fromAxiId(AxiId#(`TLM_PRM) id);
+   return cExtend(id);
+endfunction
+
+typeclass BusPayload#(type a, type b) dependencies(a determines b);
+   function b getId(a payload);
+endtypeclass
+
+instance BusPayload#(AxiAddrCmd#(`TLM_PRM), TLMId#(`TLM_PRM));
+   function getId(payload);
```
quark17, line 52: If I understand, this is related to bsc-contrib PR 46, which added explicit types to `getId` because BSC was failing to unify somehow after synonym expansion was fully applied.  You then added a condition to the expansion, to not expand unless type functions are exposed -- so now this file compiles fine without explicit types.

Can you add a mention of "bsc-contrib PR 46" to the comment and that this relates to the XXX in `ctxRedCQType'` (which is GH 311)?

Also, can you construct a variant of this file that does have an ATF in it, such that expansion does happen and typecheck fails (but typecheck works if explicit types are given)?  That way we have an example of what is still broken -- and mention the example on GH 311.  (As I say in bsc-contrib PR 46, it may not even be something deep in unify that is the problem, it could simplify be how we desugar or order the type inferences or something, that just needs to be fixed somewhere.)

Also, then, I assume you want to close bsc-contrib PR 46, as it's not needed -- or do you still intend to add the explicit types there?  (The other way of addressing it in bsc-contrib is to remove the unnecessary call to `cExtend`, since that's not need at all.)

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -32,6 +32,32 @@ compile_verilog_pass ExpSizeOf_Field.bsv
...
+compile_fail_error ValueOf_KindMismatch_Arity.bs T0025
+compile_fail_error ValueOf_KindMismatch_Res.bs T0026
+
+# SizeOf in a let binding type annotation requires a Bits context
+compile_fail_error ExpSizeOf_Let_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_Let_NoCtx.bs]
+
+# Test that cExtend works when type synonyms are expanded in instance heads
+compile_pass CExtendSynonym.bsv
```
quark17, line 102: Add a comment that this relates to GH 311 and bsc-contrib PR 46.  (And, per my comment on the file, add additional tests for the case that still fails.)

testsuite/bsc.typechecker/primtcons/primtcons.exp
quark17, line 92: Is it worth adding equivalent tests for `stringOf`?

testsuite/bsc.typechecker/typeclasses/ATFUnifyError.bs
```diff
@@ -1,7 +1,7 @@
 package ATFUnifyError where
 
--- Test that unification of two different type function applications fails
--- when they cannot be proven equal.
+-- Test that unification of a type function application with a concrete type
```
quark17: Can you explain why this change?  I'd understand if you added a second test, but you're eliminating a test to add a new one -- is the old test not worth having?

testsuite/bsc.typechecker/typeclasses/ATFUnifyHKError.bs
```diff
@@ -7,5 +7,10 @@ class (Container :: * -> (* -> *) -> *) a s | a -> s where
 
 data Foo f = Foo (f Bool)
 
-conv :: a -> Foo (Elem a) -> Foo Maybe
-conv _ = id
+-- Reading a struct field whose type uses a higher-kinded ATF (Elem a),
```
quark17: Here, too, it seems like you're overwriting a test to add a new one -- do you not want to keep the old test, too?

quark17 (commented):
In an offline conversation, we agreed to go ahead with this version (conditional expansion).  However, I've identified a few questions and concerns in the testsuite, including additional tests to add.  Are you able to address that?

quark17:
Also, does this fix Bug #313 (SizeOf in a type synonym used in an interface method typechecks fine, but when synthesizing results in an error about not being bitifiable)?  If so (or even if not), can you add the example from #313 to the tests, demonstrating the current behavior?

testsuite/bsc.typechecker/primtcons/ExpSizeOf_Let_NoCtx.bs
krame505: Done — updated the comment to describe the current (fixed) behavior: the `SizeOf` in the let-binding's declared type introduces the implicit `Bits` context, and because it isn't provided here the checker now reports T0030. I referenced the actual issue (GitHub #890) rather than "a known bug", per your note about giving the number.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Done:
```
# Test that SizeOf is expanded even when it's buried in a type synonym.
# This was Bluespec Inc Bug 1729 (also GitHub Issue #311).
```

testsuite/bsc.typechecker/primtcons/ExpSizeOf_FieldSyn_BS.bs
krame505: Done (and I kept the original `ExpSizeOf_FieldSyn_BS.bs` unchanged — its field really is unused, so the name still fits).

**User of the struct:** `ExpSizeOf_FieldSyn_Use_NoCtx.bs` (fails, T0030) and `ExpSizeOf_FieldSyn_Use_WithCtx.bs` (passes). The NoCtx version reads the field (`n.wide == 0`) with no `Bits` context and the error names the introduced `Bits a a__`, confirming the requirement is handled during typecheck. The used field is renamed to `wide`.

**The struct itself:** `ExpSizeOf_FieldSyn_Unbound_NoCtx.bs` / `..._WithCtx.bs`. Good news — BSC *does* complain: with the unbound type variable it reports T0030 and requires the field to declare `(Bits dataType sz) => T dataType`, exactly as you expected. So the `tiField1` XXX is not wrongly exempting this case, and I don't think a bug needs (re)opening for it. One cosmetic wart: BSC emits two T0030s here, one with "Unknown position" (no source location), so I registered the test with an expected count of 2 rather than pinning that output via `compare_file`. Happy to file a minor bug for the missing position if you think it's worth it.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: I don't think a dump is needed. Without the `ctxRedCQType'` fix the needed `Bits` context isn't injected into the derived `Generic` instance, and that instance then *fails to typecheck* — i.e. the failure is a plain compile error, which `compile_pass` already guards. I made this explicit in the comment (in both the test file and the .exp).

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Done — reworked the header so the section is about type functions in explicitly-typed expressions generally (`valueOf`, `stringOf`, type-annotated expressions, and explicitly-typed let bindings — noting valueOf/stringOf are themselves implemented via typed expressions), all under GH #890, with a sub-header per case. Added the missing typed-expression example: `ExpSizeOf_TypedExpr_NoCtx.bs` (T0030) / `ExpSizeOf_TypedExpr_WithCtx.bs` (passes), using `(_ :: Bit (SizeOf a))`.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Agreed — there's no dedicated place for pseudo-type-function tests, so I've kept them here.

testsuite/bsc.typechecker/primtcons/CExtendSynonym.bsv
krame505: Right on all counts.

- Expanded the `CExtendSynonym.bsv` header to mention bsc-contrib PR 46, note that the synonyms drop `addr_size`, and reference the XXX in `ctxRedCQType'` / GH #311. It compiles only because the plain synonyms contain no type function, so the conditional expansion leaves them alone.
- Added the still-broken variant you asked for: `CExtendATF.bsv`. It puts a user-defined ATF (`IdW#(Bit#(n)) = n`, playing the role of `SizeOf`) into the id-type synonyms, so expanding them in the instance head *does* reveal a type function → the conditional expansion fires → the dangling `addr_size` returns → **T0035**. Adding an explicit method type makes it compile (`CExtendATFExplicit.bsv`), which supports your hunch that this is about inference ordering rather than anything fundamental in unify. (The same failure happens with builtin `SizeOf` in place of the ATF.)
- Since CExtendSynonym now compiles without the explicit types, bsc-contrib PR 46 is no longer needed, so I'm closing it.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Done — added the GH #311 / bsc-contrib PR 46 references here in the .exp, and registered the still-failing case (`CExtendATF.bsv` → `compile_fail_error ... T0035`) alongside the passing `CExtendATFExplicit.bsv`. See the reply on `CExtendSynonym.bsv` for details.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Already there — `ExpSizeOf_StringOf_NoCtx.bs` (T0030) + `ExpSizeOf_StringOf_WithCtx.bs`, mirroring the valueOf tests. The restructured section keeps them under their own `# stringOf` sub-header.

testsuite/bsc.typechecker/typeclasses/ATFUnifyError.bs
krame505: You're right — I've kept both. Restored the original `id`-based test under its original name (`ATFUnifyError.bs`) and added the struct-field-access version as a new file (`ATFFieldUnifyError.bs`). The `id`-based test still fails with T0030 (now with one extra "Introduced at" location, from expanding the type function in the declared type), so I refreshed its expected output. I also fixed the comment, which described it as testing "two different type function applications" — it actually unifies one ATF application (`Elem a`) with a concrete type (`Bool`).

testsuite/bsc.typechecker/typeclasses/ATFUnifyHKError.bs
krame505: Same as `ATFUnifyError` — kept the original `id`-based test (`ATFUnifyHKError.bs`, with refreshed expected output) and added the struct-field-access version as a new file (`ATFFieldUnifyHKError.bs`).

krame505:
Pushed a follow-up commit addressing the review (testsuite-only; `make localcheck` passes in both `primtcons` (31) and `typeclasses` (99)). Per-comment replies are inline; the highlights:

- Kept the original `id`-based `ATFUnify{,HK}Error` tests and added struct-field-access variants (`ATFFieldUnify{,HK}Error`).
- Added struct-field context tests (`ExpSizeOf_FieldSyn_Use_*`, `ExpSizeOf_FieldSyn_Unbound_*`), a typed-expression test (`ExpSizeOf_TypedExpr_*`), and regrouped the explicit-type tests under a corrected header.
- Added `CExtendATF{,Explicit}.bsv` as the still-broken ATF variant (GH #311), and recorded bug numbers / bsc-contrib PR 46 in comments. I'll close bsc-contrib PR 46 since it's no longer needed.

On **Bug #313**: this does **not** fix it. Confirmed on the branch — the #313 example typechecks fine but still fails to synthesize with T0043 (`WriteOnly#(Bit#(1))` not bitifiable), because GenWrap makes the method-vs-subinterface determination before typecheck and can't reduce `SizeOf#(Bit#(2))`. I added the example as `ExpSizeOf_VectorIfc.bsv`, registered as `compile_pass` + `compile_verilog_fail_error T0043`, documenting the current behavior (it should become a `compile_verilog_pass` if #313 is fixed).

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -33,31 +34,81 @@ compile_verilog_pass ExpSizeOf_Field.bsv
...
 compile_pass ExpSizeOf_StringOf_WithCtx.bs
 
+# type-annotated expression, e.g. '(_ :: Bit (SizeOf a))'
+compile_fail_error ExpSizeOf_TypedExpr_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_TypedExpr_NoCtx.bs]
+compile_pass ExpSizeOf_TypedExpr_WithCtx.bs
+
+# explicitly-typed let binding
+compile_fail_error ExpSizeOf_Let_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_Let_NoCtx.bs]
```
quark17, line 89: There's no `WithCtx` version for `Let`?  Up to you if you want to add one, for completeness.

testsuite/bsc.typechecker/primtcons/primtcons.exp
```diff
@@ -33,31 +34,81 @@ compile_verilog_pass ExpSizeOf_Field.bsv
...
+# typecheck of the field access, so it must be provided (GitHub Issue #890).
+compile_fail_error ExpSizeOf_FieldSyn_Use_NoCtx.bs T0030
+compare_file [make_bsc_output_name ExpSizeOf_FieldSyn_Use_NoCtx.bs]
+compile_pass ExpSizeOf_FieldSyn_Use_WithCtx.bs
+
+# A struct field whose SizeOf-using type has a type variable that is not a
+# parameter of the struct is not exempted from the Bits requirement: the field
+# must declare the context explicitly (GitHub Issues #890 and #310).  BSC
+# currently reports two T0030 errors here (one of which lacks a source
+# position), so this uses an expected count of 2 rather than compare_file.
```
quark17: This seems like an argument for using `compare_file`: We want to detect when details like position change in odd cases.

Offhand, do you know why we get two messages?  Is one from ctxreduce and one from typecheck?

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Added `ExpSizeOf_Let_WithCtx.bs` — the same let binding, compiling successfully with an explicit `Bits` context on the enclosing definition.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Done — switched to `compare_file`, pinning both messages (including the "Unknown position" one).

On why two messages: both come from typecheck, not ctxreduce — and there are actually *four* underlying errors, deduplicated to two. The struct declaration derives four definitions that each bind the field: the `PrimMakeUninitialized` and `PrimMakeUndefined` instances for the wrapper struct generated for the polymorphic field, and the `Generic` instance's `from`/`to` methods. Each field binding goes through `tiField1`, where `expTFun` on the field's type introduces the `Bits` proviso; it can't be satisfied, so each definition reports T0030 (`EWeakContext` via `handleWeakContext` in `tiExpl'''`). The `nub` in `TypeCheck.tiDefns` then collapses the four to the two distinct texts you see (verified with `-continue-after-errors`, which skips the nub and shows all four). The three bindings of the wrapper's generated `val` field id have no source position — hence "Unknown position" — while `Generic.to` constructs `MyS { sfield = a1.val }` with the source `sfield` id and so carries its position. (Giving the generated wrapper field the original field's position would presumably fix the "Unknown position" wart.)

I've recorded this explanation in comments in the test file and the .exp.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Yeah good idea, done.

testsuite/bsc.typechecker/primtcons/primtcons.exp
krame505: Done.  

The errors are actually both from type checking, from instances derived for the wrapper struct that gets generated for the higher-rank field when deriving the generic instance.  I added an explaination.

quark17 (approved):

PR #918, opened 2026-04-14
Make dVars traversal more efficient by folding sets across applications.

nanavati:


quark17:
What does the recursion check do?  If something is recursive, what happens?  I'm just wondering if it's possible to create an example that was recursive, but not in the more precise way, to observe the difference (and add it to the testsuite).

Anyway, it looks fine -- aside from testsuite failures.  I'd be curious to know what changed -- presumably something more than unique IDs.  Probably it's expected and the testsuite just needs to be updated.

quark17:
The change to `bsc.bugs/bluespec_inc/b292/` is OK.  (I wonder if we'd see less changes there if we just sorted the defs by unique number?  It feels like changes due to a set or map being flattened in an unstable way.)

The change in `bsc.misc/sal/` and `bsc.misc/lambda_calculus/` is because a def named `state` (which comes from the source file) has been replaced with `x` (presumably a name from the library).  This is introduced by the second commit (the change to the recursion check).  I would guess that, by changing when simplification happens, either `x` is no longer being replaced by `state` or `state` is now being replaced by `x`.  I don't think your commit is to blame, it's just exposing a name in the library that should be marked "hide", say.  So I'll go ahead and push a commit to update the expected files.

quark17 (dismissed):

quark17 (approved):

Issue #863, opened 2025-12-19
Clock-crossing errors should have better positions without the -cross-info flag


nanavati:
This is noted in the testsuite here: https://github.com/B-Lang-org/bsc/blob/ceee6dab29a180d539219ef550759b6348f5779e/testsuite/bsc.mcd/Misc/mcd.exp#L49 (and was apparently Bluespec bug 1238). This is just capturing the issue on GitHub.

quark17:
Here are the comments from Bluespec Inc's bug 1238:

----

**DESCRIPTION**

In bsc.mcd/Misc/mcd.exp there is a comment that the positions in this error need to be improved:
```
Error: "ClockCheckCond.bsv", line 2, column 18: (G0007)
  Reference across clock domain in method out .
  Method calls by clock domain:
    Clock domain 1:
      Prelude.read at "Prelude.bs", line 1428, column 26,
    Clock domain 2:
      Prelude.read at "Prelude.bs", line 1458, column 26,
```
The references are probably to the "read" of the RegN and RegUN imports, and not to the places where _read is called in the user's design.

Not sure what can be done, but the exp file says that a bug should be filed.

----

**COMMENT 1**

A similar thing happens in another example:
```
Error: "ClockCheckMethodArg.bsv", line 8, column 6: (G0007)
  Reference across clock domain in rule test .
  Method calls by clock domain:
    Clock domain 1:
      Prelude.write at "Prelude.bs", line 534, column 45,
    Clock domain 2:
      Prelude.read at "ClockCheckMethodArg.bsv", line 9, column 8,
```
Also, the method/rule name should be put in quotes, in these messages.  And the Ids should be displayed with pfPrint -- "Prelude.write" is the Classic form of qualification.

----

**COMMENT 2**

Oops.  It appears that the exp file is compiling with the `-cross-info` flag on, which gives positions in the user code -- the user shouldn't have to turn on that flag to get better positions, though.

It's possible that when the original exp file comments were made, the position was worse.  In any case, the positions with cross-info are still not perfect:
```
Error: "ClockCheckCond.bsv", line 2, column 18: (G0007)
  Reference across clock domain in method out .
  Method calls by clock domain:
    Clock domain 1:
      Prelude.read at "ClockCheckCond.bsv", line 2, column 18,
    Clock domain 2:
      Prelude.read at "ClockCheckCond.bsv", line 14, column 19,
```
[...]

PR #919, opened 2026-04-14
Plug memory leaks in the evaluator

nanavati:


src/comp/IExpandUtils.hs
```diff
@@ -2451,25 +2451,31 @@ addHeapCell tag cell = do
...
   (p, r) <- addHeapCell tag newcell
-  let result = IRefT t p r
+  let poss = S.singleton $ getIExprPosition e
+  let result = IRefT t p poss r
   when doTraceHeap $ traceM ("addHeapWHNF " ++ ppString cell_name ++ " [" ++
                              prPositionConcise (getPosition cell_name) ++ "] " ++
                              ppReadable (result,t,pe))
-  return result
+
+  -- The deepseq forces the type and position in the IRefT so that
```
quark17: Do you want to repeat this comment on the other `deepseq` below?

src/comp/IExpand.hs
```diff
@@ -568,11 +568,11 @@ eqPtrs heap ptrs =
         step p (dsm, ptrm) =
                 let e = sub (heapOf p)
                     sub (IAps f ts es) = IAps (sub f) ts (map sub es)
-                    sub e@(IRefT t i _) =
+                    sub e@(IRefT t i _ _) =
                         case IM.lookup i ptrm of
                         Nothing -> e
                         -- hd_ref errors because we don't use it once we have the iheap
-                        Just i' -> IRefT t i' (internalError "eqPtrs ref")
+                        Just i' -> IRefT t i' (internalError "eqPtrs ref") (internalError "eqPtrs ref")
```
quark17, line 580: Did you want the strings to be slightly different, so they can be told apart?

testsuite/bsc.long_tests/log2_loop/pure/log2_loop.exp.golden
```diff
@@ -1,12 +1,4 @@
...
-#test_veri_only_bsv_modules_options PureLoop {} $rts_flags out.expected
-
-# Instead, use this:
-compile_verilog_fail PureLoop.bsv {} $rts_flags
-
-# And confirm that the test failed in the way we expect
-find_n_strings [make_bsc_vcomp_output_name PureLoop.bsv] "Heap exhausted" 1
+# This passes with -M16M but I am leaving some breathing room for things like
+# profiling builds.
+set rts_flags {+RTS -H16M -M32M -Sstderr -RTS}
```
quark17, line 3: I seem to recall that BSC might ignore `-M` values that are too small, but I can't recall what is too small.  Does `-M32M` give a warning (with newer GHC) about it being ignored?

testsuite/bsc.mcd/Misc/mcd.exp
```diff
@@ -48,26 +48,23 @@ test_c_veri_bsv_modules LaunderClock {mkLaunderClock}
 
 # these errs should have better position w/o the -cross-info flag (bug 1238)
 if {$vtest == 1} {
-compile_verilog_fail_error ClockCheckRule.bsv G0007 1 "" "-cross-info"
-compare_file ClockCheckRule.bsv.bsc-vcomp-out
+compile_verilog_fail_error ClockCheckRule.bsv G0007 1
+compare_file_bug ClockCheckRule.bsv.bsc-vcomp-out "" 1238
```
quark17: I would be careful with `compare_file_bug` because the file can change, just not in the expected way, and we'd fail to notice it.  I would tend to have a second test that the output matches the bad output that we expected.

Also, can you add a mention that this is GitHub Issue 863?  (Bug 1238 is the old BSInc number)

testsuite/bsc.mcd/Misc/mcd.exp
```diff
@@ -48,26 +48,23 @@ test_c_veri_bsv_modules LaunderClock {mkLaunderClock}
...
+compare_file_bug  ClockCheckRule2.bsv.bsc-vcomp-out "" 1238
 
-compile_verilog_fail_error ClockCheckMethod.bsv G0007 1 "" "-cross-info"
-compare_file ClockCheckMethod.bsv.bsc-vcomp-out
+compile_verilog_fail_error ClockCheckMethod.bsv G0007 1
+compare_file_bug ClockCheckMethod.bsv.bsc-vcomp-out "" 1238
 
-compile_verilog_fail_error ClockCheckCond.bsv G0007 1 ""
-copy ClockCheckCond.bsv.bsc-vcomp-out ClockCheckCond.bsv.nocrossinfo.bsc-vcomp-out
-compare_file ClockCheckCond.bsv.nocrossinfo.bsc-vcomp-out
```
quark17, line 62: If we're no longer comparing against this file, should it be deleted?

quark17 (commented):
Very cool.  Looks good, aside from the comments below.

If I understand, this adds the ability to collect multiple positions for a ref, but somewhere we must still only be using one of them -- can you remind me where that happens?

Also, what is the effect on those clock messages?  If I understand, without the cross-info flag, the message would have no positions, and with the flag it would have position.  So is the current behavior one of those?  If it's no position, why is that?  Are we not capturing the multiple positions, such that the cross-info one is also available?

No worries if there's no improvement.  Just maybe make sure to add a comment to Issue 863 with an update on the status.

testsuite/bsc.long_tests/log2_loop/pure/log2_loop.exp.golden
nanavati: I haven't seen a warning with the GHCs I typically use (9.4.x and 9.6.x)  I'm using. I remember I was able to go lower (12M maybe?) and get the test to fail.

src/comp/IExpand.hs
nanavati: They're basically the same problem (touching an invalid part of the IRefT). I wouldn't be opposed to different strings, but I also don't think they'll make a significant debugging difference in practice.

src/comp/IExpandUtils.hs
nanavati: We do the deepseq in several places. Maybe it should be a utility function? (a specific one about IRefT, not a generic deepseq x x deal)

src/comp/IExpand.hs
quark17: If you don't think it will matter when debugging, to know which of the two was triggered (say by adding "1" or "2" to the string, or other identifier), then I'm OK with how it is.

testsuite/bsc.long_tests/log2_loop/pure/log2_loop.exp.golden
quark17: For the record, here is the GHC issue that I was thinking of, that mentions the `-H` and `-M` RTS flag bugs and gotchas: https://gitlab.haskell.org/ghc/ghc/-/issues/24487

src/comp/IExpandUtils.hs
nanavati: Made and now use `mkIRefT` utility to cover exactly this.

testsuite/bsc.mcd/Misc/mcd.exp
nanavati: I made the testing more careful - checking against the original .expected to trigger XPASS if something is fixed while matching the current expected bad output directly so we can catch drift. Does that sound like a good strategy?

testsuite/bsc.mcd/Misc/mcd.exp
nanavati: Yes, I think the new testing strategy (good expected and bad expected both checked) resolves this. What do you think?

nanavati:
> Very cool. Looks good, aside from the comments below.
> 
> If I understand, this adds the ability to collect multiple positions for a ref, but somewhere we must still only be using one of them -- can you remind me where that happens?
> 
> Also, what is the effect on those clock messages? If I understand, without the cross-info flag, the message would have no positions, and with the flag it would have position. So is the current behavior one of those? If it's no position, why is that? Are we not capturing the multiple positions, such that the cross-info one is also available?
> 
> No worries if there's no improvement. Just maybe make sure to add a comment to Issue 863 with an update on the status.

Currently everywhere -- the set is collected but nothing reads it yet.
And yes, clock messages now always show the old no-flag behavior: the
-cross-info rewriting that stamped positions onto Ids was the leak source.
Plumbing the collected position sets into the G0007 report is the intended
fix for #863.

src/comp/IExpandUtils.hs
```diff
@@ -2493,7 +2503,7 @@ addHeapNF tag t pe ws = do
 -- flags <- getFlags
 -- mapIExprPosition flags?
   when doTraceHeap $ traceM ("addHeapNF " ++ ppReadable (result,t,pe))
-  return result
+  return $ deepseq result result
```
quark17: If you're going to update this comment, remove the `deepseq` and use `mkIRefT` in the assignment to `result` above.

src/comp/IExpandUtils.hs
```diff
@@ -2458,30 +2458,40 @@ addHeapCell tag cell = do
...
   (p, r) <- addHeapCell tag newcell
-  let result = IRefT t p r
+  let poss = S.singleton $ getIExprPosition e
+  let result = mkIRefT t p poss r
   when doTraceHeap $ traceM ("addHeapWHNF " ++ ppString cell_name ++ " [" ++
                              prPositionConcise (getPosition cell_name) ++ "] " ++
                              ppReadable (result,t,pe))
-  return result
+
+  return $ deepseq cell_name result
```
quark17: Why was `deepseq` of `cell_name` added here? and why was it not added similarly to `addHeapUnev` above?

If you intend it in both places where `improveCellName` is called, you could put the `deepseq` in that function?

testsuite/bsc.mcd/Misc/mcd.exp
```diff
@@ -46,28 +46,38 @@ test_c_veri_bsv_modules LaunderClock {mkLaunderClock}
...
+compare_file ClockCheckRule2.bsv.bsc-vcomp-out \
+    ClockCheckRule2.bsv.bsc-vcomp-out.bad-expected
+
+compile_verilog_fail_error ClockCheckMethod.bsv G0007 1
+compare_file_bug ClockCheckMethod.bsv.bsc-vcomp-out "" 863
+compare_file ClockCheckMethod.bsv.bsc-vcomp-out \
+    ClockCheckMethod.bsv.bsc-vcomp-out.bad-expected
+
+compile_verilog_fail_error ClockCheckCond.bsv G0007 1
+
```
quark17: You could remove this blank line between steps of the same test, if you want

quark17:
Looks good.  There were just a few minor remaining things I noticed.

src/comp/IExpandUtils.hs
nanavati: Done

testsuite/bsc.mcd/Misc/mcd.exp
nanavati: Done

src/comp/IExpandUtils.hs
nanavati: The main reason was that the names that made it to `addHeapUnev` were usually forced already so a second `deepseq` seemed redundant. But in fact there are corner cases I hadn't accounted for so it makes more sense to centralize the forcing in `addHeapCell` and `updHeap` so there's no way on the heap without a deeply forced name. That is the property to maintain.

quark17:
Since merging the port-splitting PRs, this has a conflict with `IExpand` that is not immediately obvious to me how to resolve.

nanavati:
The conflict is a modify/delete. #849 replaced walkNF's type-guarded ActionValue-ref arm

```haskell
[e_ref@(IRefT t ptr ref)] | (isitActionValue_ t) || (isitAction t) -> ...
```

with the selector-id-keyed arm (`i_sel` ∈ {`avValue_`, `avAction_`, `fst`, `snd`}).

This branch's only change to that arm was widening the `IRefT` pattern for the new position field (the body is untouched, and the new binding is unused there).

I think the right way to resolve the conflict is by taking 849's side wholesale, i.e. delete the old arm. Why:

- The selector-keyed arm covers it: it matches any single argument, heap refs included, so the "inner selector wound up on the heap via evalHeap's move" case still recurses.

- The old arm's only residual coverage is a selector *outside* those four applied to an ActionValue/Action-typed ref. That is a malformed selection, which #849 deliberately routes to the `nfError "walkNF sel"` diagnostic instead of silently recursing. I think that is the right move for that malformed case (unless we think the failure is serious enough to require `internalError`).

- The surviving arm binds a bare `[e]` with no `IRefT` pattern, so none of this branch's arity changes apply inside the conflicted hunk; they all land in non-conflicting hunks elsewhere and should merge cleanly.

quark17:
Thank you.  I now see that Git/GitHub present the conflict as two differing blocks of text, and that misled me; if I go look at what this PR actually changed, it was only updating a pattern match of `IRefT t ptr ref` to be `IRefT t ptr _ ref` -- so obviously we could resolve the conflict by taking the code in `main` and just adding the field where it might needed.

Taking a bigger picture look: there is probably a lot of new code that wasn't flagged as a "conflict" and was silently inserted, that therefore didn't get the memory-leak analysis that was done at the time this PR was written.  Should we be concerned that new leaks have been introduced?

I'll rebase this PR and, if the testsuite passes, I'll merge it.

quark17 (approved):

quark17:
I've pushed a rebase, that addresses conflicts and fixes a missing IRefT needed to compile.  I also kept in a change to the log2_loop test that used `-K20`, however before this was adding the flag when no flag existed, but now it's decreasing the value (from 30, on 'main').  The change was adding `-steps` which is now on `main` and was decreasing the heap sizes, so I figure we can see if this drop in stack is OK for the CI -- otherwise we can put it back to 30.

PR #920, opened 2026-04-18
Resolve GHC build warnings

quark17:
Fix two GHC warnings introduced by PR #847: unused import and a missing case arm (internal error for unexpected CDefn disjuncts).

### DIFF
--- a/src/comp/GenSign.hs
+++ b/src/comp/GenSign.hs
@@ -20,7 +20,7 @@ import Position(noPosition)
 import Pragma
 import CSyntax
 import CFreeVars
-import CType(leftTyCon, tyConArgs, isTConArrow, allTyCons, allTConNames)
+import CType(leftTyCon, tyConArgs, isTConArrow, allTyCons)
 import Pred(Class(..), predToCPred, expandSyn)
 import SymTab
 import Assump(Assump(..))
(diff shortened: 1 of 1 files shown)


src/comp/GenSign.hs
```diff
@@ -761,5 +761,6 @@ getPackagesUsedByExports currentPkg (CSignature _ _ _ defns) =
     getPackageFromDefn (CprimType (IdKind i _)) = getIdPackage i
     getPackageFromDefn (CIinstance _ _) = Nothing  -- Instances don't have a name
     getPackageFromDefn (CPragma _) = Nothing
+    getPackageFromDefn d = internalError ("getPackageFromDefn: " ++ ppReadable d)
```
nanavati: Only question I have if this internalError should mention that the construct in question shouldn't have survived past typecheck? I guess maybe match what IConv does since it has the same issue?

It's not a big deal whatever you decide, just figured it was worth mentioning.

nanavati (dismissed):

src/comp/GenSign.hs
quark17: IConv says nothing, beyond `show` or `ppReadable` of the values.  I can add a comment here, though, or something in the string.

PR #922, opened 2026-04-19
Turn on -Werror in GHCFLAGS

nanavati:
Locally suppress the known, long-standing warnings

This would prevent some of the warning issues we've had, though I worry that the warnings to turn off might be tricky when they are in new GHC versions only. Let's see what CI does.

### DIFF
--- a/src/comp/AUses.hs
+++ b/src/comp/AUses.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-incomplete-record-updates #-}
 {-# LANGUAGE PatternGuards #-}
 -- AUses
 --
--- a/src/comp/Makefile
+++ b/src/comp/Makefile
@@ -183,6 +183,7 @@ GHCFLAGS = \
 	-hide-all-packages \
 	$(FVIA) \
 	-Wall \
+        -Werror \
 	-fno-warn-orphans \
 	-fno-warn-name-shadowing \
 	-fno-warn-unused-matches \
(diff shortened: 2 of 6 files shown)


quark17:
I'm all for turning this on!  FYI, Issue #469, which was originally about new warnings in GHC 9.2 (that were mostly resolved), mentions [in a comment at the end](https://github.com/B-Lang-org/bsc/issues/469#issuecomment-1756703937) the new warnings in GHC 9.8+ for head/tail.

quark17:
Is there an alternative to not turning off `-warn-incomplete-record-updates`?  We do want to detect if new code is introduced that does something wrong.  As I understand, the warnings we're seeing are about code that is fine but GHC for some reason has a problem with uses of `~`.

nanavati:
Incomplete record updates are funny - we always do them after selecting the field in question (so it would crash before it got to the record update). The problem generally is the field we want to update is on a lot but not all of the alternatives of an ADT.

One possibiity would be using ViewPatterns to create something that let you extract / replace the field safely? Or make the updates applying an update function which touches and updates the field when there and ignores it when it isn't?

src/comp/AUses.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-incomplete-record-updates #-}
```
philderbeast, line 1: Are we still supporting GHC < 8? The `-fno-warn-*` [warning syntax is deprecated](https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html) and its replacement is `-Wno-*`.

PR #923, opened 2026-04-20
Detect coherent matches that depend on incoherent ones

nanavati:
These behave incoherently even if the top-level match appears coherent, so we want to detect this case and warn / error as appropriate. Also, having a precise coherent / incoherent boundary will help dictionary lifting.


src/comp/Id.hs
```diff
@@ -161,7 +161,8 @@ data IdProp = IdPCanFire
               -- used by the BSV parser to keep track of which array types
               -- were introduced from bracket syntax
               | IdPParserGenerated
-              | IdPIncoherent           -- Used to track incoherent instance matches for future use
+              | IdPIncoherent           -- Used to track incoherent instance matches
+              | IdPCAF                  -- is a CAF (introduced by LiftDicts)
```
quark17: I was going to mention that the acronym `CAF` should be explained, but I don't see that this is used anywhere (not anything called `LiftDicts`).  I guess this related to the follow on PR (925); should it be removed from this PR then?

src/comp/TCheck.hs
```diff
@@ -2547,9 +2547,9 @@ tiExpl''' as0 i sc alts me (oqt@(oqs :=> ot), vts) = do
         (rs_amb, rs_unamb) = partition (any (`elem` amb_vars) . tv) rs
 
     -- Apply the substitution to the code fragments
-    let alts''     =  apSub s alts'             -- new alternatives
-        asbs       =  apSub s sbs1              -- new dict bindings
-        me''       =  apSub s me'               -- update guards
+    let alts'' = apSub s alts'
```
quark17: Do you want to preserve the comments?

testsuite/bsc.typechecker/typeclasses/coherence/coherence.exp
```diff
@@ -0,0 +1,40 @@
...
+# IncoherentBase via a proviso.  Each Super_* test uses IncoherentBase as a
+# superclass constraint instead.
+
+# ---- Instance-level proviso constraints ----
+
+# coherent class depending on incoherent resolution: compile error T0159.
+compile_fail_error Instance_Coherent.bs T0159
+
+# default class (no annotation): T0159 error when flag off, T0158 warning when flag on.
+compile_fail_error Instance_Default_FlagOff.bs T0159
```
quark17: Are FlagOff and FlagOn identical files?  If so, I would have just one file and copy it to the two names.

testsuite/bsc.typechecker/typeclasses/coherence/coherence.exp
```diff
@@ -0,0 +1,40 @@
...
+# to all callers, so behaviour is coherent from the callers' perspective.
+# No T0158 warning is expected.
+
+test_c_veri_bs_modules_options Super_Coherent {} {}
+find_n_warning Super_Coherent.bs.bsc-vcomp-out T0158 0
+
+test_c_veri_bs_modules_options Super_Default_FlagOff {} {}
+find_n_warning Super_Default_FlagOff.bs.bsc-vcomp-out T0158 0
+
+test_c_veri_bs_modules_options Super_Default_FlagOn {} {}
```
quark17: This isn't compiling with the flag on!

Also, if the FlagOn and FlagOff files are identical, I would use one file and copy it.

src/comp/TCMisc.hs
```diff
@@ -571,7 +613,11 @@ reducePred eps dvs (VPred w pp@(PredWithPositions pr@(IsIn c ts) pos)) = do
                                      ppReadable (v', i', m_tv, inst_subst, fd_subst, incoherent))
                      let Inst _ _ (_ :=> h) _ = i
                      let minst = toMaybe incoherent h
-                     return $ Just (qs, sb, fd_subst, minst)
+                         -- Mark the binding incoherent for LiftDicts.
```
quark17: Another reference to LiftDicts, which I guess is coming in the follow-on PR (925) but doesn't exist here yet.

quark17 (commented):
Looks reasonable, but a few comments below.

src/comp/Id.hs
nanavati: CAF = "constant applicative form" (a term used when talking about GHC's handling of similar expressions), but it isn't used here so I'll move the definition and comment to 925

src/comp/TCheck.hs
nanavati: Yes, fixed.

testsuite/bsc.typechecker/typeclasses/coherence/coherence.exp
```diff
@@ -0,0 +1,42 @@
...
+
+test_c_veri_bs_modules_options Super_Coherent {} {}
+find_n_warning Super_Coherent.bs.bsc-vcomp-out T0158 0
+
+test_c_veri_bs_modules_options Super_Default_FlagOff {} {}
+find_n_warning Super_Default_FlagOff.bs.bsc-vcomp-out T0158 0
+
+# The expected sim output is identical with the flag on
+copy sysSuper_Default_FlagOff.out.expected sysSuper_Default_FlagOn.out.expected
+test_c_veri_bs_modules_options Super_Default_FlagOn {} {-incoherent-instance-matches}
```
quark17: Again, if FlagOff and FlagOn are the same file, would it make sense to have one file and copy it?

src/comp/TCMisc.hs
```diff
@@ -217,6 +218,40 @@ mkATFClassPred tag posType clsId pIdxs tIdx atfArgs targetType = do
...
+          (S.toList newlyIncoherent)
+    return sbs'
+  where
+    diagnoseOne typeMap diMap clsMap i = do
+      ai <- getAllowIncoherent
+      case M.lookup i clsMap >>= allowIncoherent of
+        Just True -> return ()   -- incoherent class: suppress
+        mallow    ->
+          if fromMaybe ai mallow
+            then twarn (pos, WTransitiveIncoherentMatch tStr rootPredStr rootInstStr)
```
quark17: Do these need to be two separate tags?  Can we use the same tag, and just warn or error as appropriate?

The messages are identical, except that one starts with "coherent proviso" and the other with just "proviso" -- but the choice is based on a global flag and not a property of the specific proviso.  Could we just use the same message in both error and warn situations? or, at worst, a parameter to the tag that indicates when to say something different.

quark17 (approved):

PR #924, opened 2026-04-20
Fix bugs related to recording and dropping uses of instances

nanavati:
These tests involve orphan instances because without orphans there is usually a use of a typeclass or a type that makes an imported module used regardless of how instances are or are not recorded.


nanavati:
This is implemented on top of #903

testsuite/bsc.driver/unused_imports/unused_imports.exp
```diff
@@ -44,6 +44,50 @@ compile_pass_no_warning InstanceUseBS.bs
...
+# The fixed sat branch does not spuriously record the package.
+compile_fail_error IncoherentUse.bsv T0030 1
+compile_fail_error IncoherentUseBS.bs T0030 1
+
+# Test 6f: Incoherent match allowed (-incoherent-instance-matches).
+# GeneralDescribable commits via the incoherent sat branch and is recorded
+# (no T0157 for it).  BoolDescribable only triggered the predUnify incoherence
+# detection (byInst returns Nothing for it) so it is never committed and T0157
+# correctly fires for it.  Two T0126 warnings fire (typecheck + ctxReduce).
+compile_pass_warning IncoherentUse.bsv T0126 2 "-incoherent-instance-matches"
```
quark17, line 88: This checks for two T0126 warnings, but it does not confirm that there were no other warnings (such as T0157).  If you don't want to compare an expected output, what about adding `find_n_warning ... T0157 0`?

quark17 (commented):
Looks fine, except for the one testsuite issue above.

testsuite/bsc.driver/unused_imports/unused_imports.exp
quark17: This is the output of BSC:
```
Warning: "IncoherentUse.bsv", line 22, column 12: (T0126)
  Proviso Helper::Describable#(a) satisfied by incoherently matching the
  instance Helper::Describable#(a)
Warning: "IncoherentUse.bsv", line 22, column 12: (T0126)
  Proviso Helper::Describable#(a) satisfied by incoherently matching the
  instance Helper::Describable#(a)
Warning: "IncoherentUse.bsv", line 4, column 8: (T0157)
  Package `BoolDescribable' is imported but not used
```

I don't know why I thought there would be no occurrences of T0157, as you do say that "T0157 correctly fires".  Either way, it isn't being checked.

It's unfortunate that we two copies of the message -- without `-v` to show the stages, it looks like BSC reported a duplicate message.

quark17: I see: there's no T0157 for Helper but there is T0157 for BoolDescribable.  Checking for 1 occurrence wouldn't check that it's the correct occurrence, so I added a comparison against an expected file.

PR #925, opened 2026-04-21
Lift dictionaries and simplify more aggressively in isimplify

nanavati:
Reimplemented again, on top of the reworked interning stack: #1048 ← #1055 (ftv cache) ←
this. The dedup no longer compares evidence expressions at all — structural comparison
(`eqEvidence`) is removed; each lifted dictionary carries a recorded identity, and
merges happen only when that identity *verifies* equal.

## What

Lifts typeclass dictionaries to package top level as ISyntax `IDef`s over interned
types, spliced in at `iConv` time, and deduplicates them — within a package and across
packages — by an identity recorded once at lift time (LiftDicts) and shipped through the
.bo in `DefProps` on the IDef (DefProp moves to its own module so it can mention IType):

- **`DefP_DictRendering`**: a slot-normalized rendering of the dictionary's own evidence
  level — an injective string over the evidence shapes
  (CApply/CTApply/CSelectT/CStructT/CLitT/CVar), every leaf atom length-prefixed, every
  qualified name encoded as two atoms so (A, B.C) can never conflate with (A.B, C).
  References to other lifted dictionaries appear only as kid-slot markers (D0, D1, ...)
  and types only as type-slot markers (T0, T1, ...).
- **`DefP_DictKids`**: the referenced lifted dictionaries, in slot order.
- **`DefP_DictTypes`**: the referenced types, in slot order. Types stay out of the
  string because interned types are in-process DAGs: a textual rendering would expand
  the DAG (exponentially in the worst case), while an `[IType]` serializes through the
  .bo writer's shared type machinery and compares in O(1) per slot by interned equality.

FixupDefs buckets candidates by interned type; within a bucket a dictionary redirects to
the first import-order candidate whose identity **verifies** equal: rendering strings
equal (sound — the rendering is injective at its own evidence level), type slots equal
element-wise by interned identity, and kid slots verified recursively pairwise by the
same procedure, memoized on the (Id, Id) pair. **No digest ever commits a merge**; every
decision reads the identities themselves. A dictionary without a rendering (incoherent
resolution, or evidence outside the renderable shapes) matches only itself by qualified
name — the never-dedup behavior its own package gave it. Kid recursion terminates
because lifted evidence is acyclic (letrec-bound dictionaries are never lifted); a
visited set guards against corrupted input by refusing the merge. When a dictionary is
dropped, surviving defs' kid lists are rewritten through the redirect map so downstream
packages verify against names they can see. .bo/.ba format tags bumped for the three new
DefProp constructors.

## Why the recorded identity

Soundness first: type-keyed sharing is unsound — two packages with different visible
orphan instances can each coherently resolve the same predicate to different instances —
so any merge must verify evidence, not types (the liftdicts T1 testcase pins this).

Against the previous scheme here (structural comparison of evidence read back out of
imported defs): the recorded identity is far cheaper — no expression walks over imported
packages per fixupDefs call — and it survives passes that rewrite the stored
expressions, instead of silently losing dedup when the post-fixup body no longer matches
the shape the comparison expected. The cost was measured, not hypothetical: on one wide
production module (a MatX `wrapper_fixup`), structural-evidence dedup spent **101
minutes of wall time**; the verified recorded-identity design compiles the same module
in **seconds**.

## Validation

- `testsuite/bsc.misc/liftdicts`: **34/34**, including the orphan-divergence soundness
  case (T1: same type, different orphan evidence — must not merge), same-evidence dedup
  (T2), diamond (T3), and divergence one level down (T4: the recursive kid verification
  must keep both). Each test also runs the design and checks its output.
- Incoherent dictionaries record no identity by construction, so `IdPIncoherent` plays
  no role in deduplication.


krame505 (commented):
I don't have enough bandwidth to give this a super deep review, but nothing here seems obviously wrong.

nanavati:
Rebased onto current main and repushed (the branch was CONFLICTING after the July merges). Notes on what changed beyond the replay:

* The interleaved commits whose content merged via #923 were dropped; their evolved upstream forms (`SolvedBinds.markIncoherent`, the `WIncoherentMatch`/`WTransitiveIncoherentMatch` warnings) are used as-is — `TCMisc` ends up with zero delta against main.
* The `Bin Flags` instance is regenerated against the chunked encoding (136 fields), and both binary-format tags are bumped to `bsc-bo/ba-20260714-1`, covering the IdProp additions (`IdPCAF` = 38, `IdPEvidenceFP` = 39) and the `liftDicts` flag — addressing the earlier format-tag review gap.
* New in the follow-up commit: the evidence fingerprint is now stored as a fixed-size **digest** rather than the unbounded canonical rendering, with **verification** wherever equality acts — a loud in-compile collision check in LiftDicts, and structural verification against the pre-fixup definitions in FixupDefs before any reference redirect or definition drop. A digest collision degrades to a lost deduplication, never a wrong merge.

Validated: closure typecheck clean, full build, and the touched suite directories green (including `liftdicts`, `typeclasses`, `signal_names`, b1490 with upstream's unified caps + a 320M exception for MyBool, and the two CSE-order regolds noted in the tip commits).

nanavati:
I've discovered an ugly performance issue with the evidence fingerprinting. Fix coming but it is going to be a few steps.

Issue #731, opened 2024-08-20
implicit pack/unpack + custom Bits instance can result in extraneous logic.


rossc719g:
The compiler inserts an implicit unpack when reading a port/register, and an implicit pack when writing.  For "standard" types like "UInt" or "Bit", these are effectively no-ops.  But when using a type with a custom Bits instance, they can generate extraneous logic. Given that the source and destination are both packed, these could be elided, and remove the extraneous logic that is generated.

See https://github.com/rossc719g/bsc_examples/blob/main/pack_unpack/README.md for a full example.

The example includes a workaround using a wrapper type.

krame505:
I think a fix for this would be to introduce some primitives:
```
primitive primPack :: (a - > Bit n) -> a -> Bit n
primitive primUnpack :: (Bit n -> a) -> Bit n -> a
```
and use these as wrappers around `pack` and `unpack` at synthesis boundaries.  The evaluator would permit these to cancel each other out as a special case, otherwise it would just use the wrapped functions.

nanavati:
> I think a fix for this would be to introduce some primitives:
> 
> ```
> primitive primPack :: (a -> Bit n) -> a -> Bit n
> primitive primUnpack :: (Bit n -> a) -> Bit n -> a
> ```
> 
> and use these as wrappers around `pack` and `unpack` at synthesis boundaries. The evaluator would permit these to cancel each other out as a special case, otherwise it would just use the wrapped functions.

I implemented this in draft PR #1008 (full design, semantic contract, and test results in its description), following the approach sketched above, extended with a propagation strategy: the coercions stay symbolic in the evaluator until a consumer genuinely demands the value, so matching pairs cancel without ever elaborating the instance bodies. `TestDoubleNorm` (#926) becomes assign _read = r ;`. Two supporting fixes split out on the way: #1006, and #1007. The latter turned out to be the original bug 1490 (cf. #659), a latent exponential that the transformation made easy to hit.

One point worth flagging here for anyone following this issue, since it is the semantic decision embedded in the fix: cancellation applies in both directions, as a documented law of the Bits class (mutual inverses up to resolution of don't-cares: the "up to" matters because derived tagged union instances violate strict equality via their don't-care padding). The observable consequence is that non-canonical bits arriving from outside are no longer laundered through a decode/re-encode round trip — garbage in, garbage out — and there is deliberately no user-facing opt-out. Objections to that stance belong here or on the PR.

[...]

PR #926, opened 2026-04-21
Tweak ITransform to get more pack . unpack to optimize properly

nanavati:


src/comp/ITransform.hs
```diff
@@ -423,6 +424,13 @@ iTrAp ctx p@(ICon _ (ICPrim _ PrimIf)) [t] [cnd, thn, els]
...
                                       -> iTrAp2 ctx p [t] [cnd,thn,iTrue]
 
+                -- if c then _ else e  -->  e
+                -- This is more aggressive than just matching UNoMatch, but
+                -- it appears to make a positive difference.
+                -- Note that enabling the symmetric simplification for thn
+                -- when els is a don't care disturbs the expected pack . unpack
+                -- structure and makes some things worse while fixing others.
+                (_, ICon _ (ICUndet {}), _) -> (els, True)
                 _ -> (IAps p [t] [cnd, thn, els], False)
```
quark17, line 434: Can we insert a blank line before the fall-through arm?

testsuite/bsc.long_tests/log2_loop/actionvalue/log2_loop.exp.golden
```diff
@@ -1 +1 @@
-test_veri_only_bsv_modules_options Log2Test  "" "+RTS -H150M -M300M -Sstderr -RTS" out.expected
+test_veri_only_bsv_modules_options Log2Test  "" "-steps 2000000 +RTS -H150M -M300M -K30M -Sstderr -RTS" out.expected
```
quark17, line 1: Why does this now churn more?  Is that a trade-off we should be concerned with?

testsuite/bsc.verilog/derived_bits/TestNoopCase.bs
```diff
@@ -0,0 +1,54 @@
...
+--       3'd1, 3'd2, 3'd4: q1 = r[3:0];
+--       default:          q1 = r[3:0];  // every branch is identical
+--     endcase
+--   end
+--   // identical always block for _write_1 / q2
+--
+-- Should be:
+--
+--   assign _read = r;
+--   assign r$D_IN = _write_1;
```
quark17, line 37: Would it be better to have the test look for this line, rather than (or in addition to) the absence of `case`?

testsuite/bsc.verilog/derived_bits/TestNoopTernary.bs
```diff
@@ -0,0 +1,63 @@
...
+--       3'd1, 3'd2, 3'd4: q1 = r[3:0];
+--       default:          q1 = r[3:0];
+--     endcase
+--   end
+--   // identical always block for _write_1 / q2
+--
+-- Should be:
+--
+--   assign _read = r;
+--   assign r$D_IN = _write_1;
```
quark17, line 37: Would it be better to have the test look for this line, rather than (or in addition to) looking for the absence of `?`?

testsuite/bsc.verilog/opt/opt.exp
```diff
@@ -88,3 +88,5 @@ if { $vtest == 1 } {
 
 # ---------------
 
+compile_verilog_pass TestDoubleNorm.bs
+find_n_strings_bug mkTestDoubleNorm.v "?" 0
```
quark17: Why `bug` and why `0`?  If there's a bug in BSC's behavior, can you file it and include the number here?

The source file says that we only expect the `read` path to simplify, but the `write` path should normalize, which is going to include `?`.  So the exp file should be checking for 1 occurrence, not 0, right?  Though it'd be better (or in addition) to check for `assign _read  = r;` rather than just `?`, I'd think?

testsuite/bsc.verilog/undet/undet.exp
```diff
@@ -19,11 +19,5 @@ if { $vtest == 1 } {
 
     # Undefined values before the end currently trigger unnecessary != tests
```
quark17: The part of the comment that you left does not make sense by itself.  Can you write a proper comment here?

src/comp/ITransform.hs
nanavati: Done

testsuite/bsc.verilog/derived_bits/TestNoopCase.bs
nanavati: Yes, done.

testsuite/bsc.verilog/derived_bits/TestNoopTernary.bs
nanavati: Yes, done.

testsuite/bsc.verilog/opt/opt.exp
nanavati: So this is effectively #731 which has now been recorded. And you're right, there should still be normalization on the write path so one `?` should survive.

testsuite/bsc.verilog/undet/undet.exp
nanavati: Comment fixed

testsuite/bsc.long_tests/log2_loop/actionvalue/log2_loop.exp.golden
nanavati: I think this churns more because if-structures that used to be left alone now simplify, which leads to more evaluation.  Specifically this case involves a recursive loop where crawling the structure can lead to more `ICDef` crossings and, hence, more steps. Since this is an evaluation stress-test and I didn't have to move the steps limit anywhere else I don't think it is a big deal.

quark17 (approved):

PR #929, opened 2026-04-22
Performance improvement: Use a size/type map instead of crawling the …

nanavati:
…list of VMItems.

### DIFF
--- a/src/comp/AVerilog.hs
+++ b/src/comp/AVerilog.hs
@@ -352,7 +352,7 @@ aVerilog errh flags pps aspack ffmap =
 
         -- remember to not redeclare any signals that needed to be
         -- declared for submodule instantiations
-        isPreDeclared = isDeclFromList inst_declared_signals
+        isPreDeclared = isDeclFromList $ S.fromList inst_declared_signals
 
         -- make a map to hold the defs, for easy access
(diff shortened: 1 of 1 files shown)


src/comp/AVerilog.hs
```diff
@@ -1639,18 +1641,20 @@ mergeCommonDecl ins =
 
 -- ==============================
 
-getSizeAndTypeM :: VId -> [VMItem] -> Maybe (Maybe VRange, VDType)
-getSizeAndTypeM i [] = Nothing
-getSizeAndTypeM i (VMDecl (VVDWire sz (VVar i') _) : _)
-    | i == i' = Just (sz, VDWire)
-getSizeAndTypeM i (VMDecl (VVDecl t sz vs) : _)
-    | (t == VDWire || t == VDReg) && i `elem` [ i | VVar i <- vs ] = Just (sz, t)
```
quark17, line 1647: This used to check that the type was either `VDWire` or `VDReg`.  I don't see that the creation of the map or the lookup ever checks this.  Does that matter?

nanavati: I think it isn't necessary because you can only get VDWire or VDReg in the input but a belt-and-suspenders check is a good idea so I added one.

quark17 (approved):

PR #930, opened 2026-04-23
Add a top-level target to run the full testsuite

nanavati:


GNUmakefile
quark17, line 29: Did you want to add the new target(s) in the help message?

GNUmakefile
```diff
@@ -72,6 +72,10 @@ check-suite:
 check-suite-parallel:
 	$(MAKE) -C testsuite checkparallel
 
+.PHONY: full-suite-parallel
```
quark17: Do you want to complete the matrix of possibilities by having a target for the full suite not parallel?  It's reasonable to say that long tests should be reserved for parallel, and not include a non-parallel option.

I would keep the convention of starting with `check` (and `check-suite`), perhaps naming it `check-suite-full-parallel` (or just `check-suite-full` if you prefer).

quark17 (approved):

GNUmakefile
nanavati: Done

GNUmakefile
nanavati: Done

PR #932, opened 2026-04-24
Use TAdd in the instance head instead of a TSub context in TupleSize

krame505:
This reverts commit 9d77afce69e7c961452363d57ef198f16e58a6a4.

This was a change I apparently made in #729, for reasons that I don't recall.  However it turns out to cause a performance regression in our code base on a module with a huge number of input ports of complex types.

### DIFF
--- a/src/Libraries/Base1/Prelude.bs
+++ b/src/Libraries/Base1/Prelude.bs
@@ -3572,7 +3572,7 @@ instance (TupleSize' a n) => TupleSize a n where {}
 
 class TupleSize' a n | a -> n where {}
 instance TupleSize' a 1 where {}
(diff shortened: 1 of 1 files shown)


quark17 (approved):

quark17:
Yeah, I avoid `TSub` because the undefined cases make it not entirely compatible with `TAdd`.  Though, in this case, is it the use of a type function in the head that improves the performance?  That is, would this equally perform poorly?:
```bh
instance (TupleSize' b n, Add n 1 m) => TupleSize' (a, b) m where {}
```

krame505:
I didn't try that variant, but I suspect that the difference is TAdd resolves during matching without creating an additional context to solve.

quark17:
But then, why doesn't TSub also resolve without needing an additional context?

Issue #489, opened 2022-07-20
Float to fixed of a denormalized number


michael-roe:
https://github.com/B-Lang-org/bsc/blob/main/src/Libraries/Base3-Math/FloatingPoint.bsv,
line 3156 says:

// todo: does this work for subnormal?

it seems that, no, it does not work for subnormal floating point numbers.

This ends up being an issue for us with CPUs written in Bluespec, because the failure to handle this case shows
up in randomized testing of the CPU.

quark17:
Thank you for reporting this.  We are exploring the FloatingPoint library on a set of tests from IBM for IEEE 754R ([here](https://research.ibm.com/haifa/projects/verification/fpgen/test_suite_download.shtml)), but it doesn't include tests for converting from float to fixed, so it wouldn't have caught this issue.  If you have examples that are failing and can provide the inputs and expected outputs, for testing, that would be helpful.  If you're able to submit a fix, that would be wonderful too.

michael-roe:
The example of it failing is when a negative subnormal number is rounded down (rounding mode is towards negative infinity), it should be rounded down to -1, but the current code rounds it up to 0.

michael-roe:
P.S. I am also running IBM's FPGen tests. I am running them on the Flute RISCV soft core; as this is written in Bluespec, it indirectly tests the Bluespec floating point libraries.

The float-to-int bug was found using a different randomized tester.

PR #933, opened 2026-04-24
Fix FloatToFixed of a denormalized number

michael-roe:


michael-roe:
This resolves issue #489

quark17:
Awesome, thank you!

This should have test cases to go with it.  Probably by adding new files to `testsuite/bsc.libs/FloatingPoint/` and updating the `FloatTest.exp` script.  Do you have BH/BSV examples that illustrate the changed behaviors?  Are you able to add them to the testsuite?

quark17:
I'm happy to help with the testsuite, but assigning this to @rsnikhil to confirm the floating point rounding mode behavior.

michael-roe:
I have my own test cases for this bug, but they probably need to be rewritten before they can be merged into your test suite:

https://github.com/michael-roe/fpu_tests

I’ll take a look at modifying them and adding them to this PR.

quark17:
I just pushed two commits (to be squashed merged with the original commit).

One removed the "todo" comments, since we've now addressed that, I believe?  But I left in the comment "this is not well tested" on the FixedPoint instance for FixedFloatCVT.

And I added your testcase.  There was only one BSV file in `fpu_tests`, which I renamed and added, with a file of expected output.  Can you confirm that this is the test and that the output looks correct?  I noticed that it is only testing the FixedPoint instance, not the UInt and Int instances, even though you made the change in all three.  (The change is the same in all three -- I assume there's no reason why one might have different behavior than the others?)

Thanks again!

quark17 (approved):

PR #936, opened 2026-04-25
Use strict ByteString with offset indexing for binary deserialization

nanavati:
Replace the [Word8] linked list in BinData's In monad state with a strict ByteString and integer offset. This eliminates millions of cons-cell allocations and pointer chases during .bo/.ba file reading.

### DIFF
--- a/src/comp/BinData.hs
+++ b/src/comp/BinData.hs
@@ -50,7 +50,7 @@ import IntLit
 import Undefined
 import Prim hiding(PrimArg(..))
 
-import Util(Hash, hashInit, nextHash, showHash)
+import Util(Hash, hashInit, nextHashByte, showHash)
 
 import Data.List(sort, intercalate)
 import Control.Monad(replicateM, liftM, ap)
@@ -59,6 +59,7 @@ import Data.Array.Unboxed
 import Data.Bits
 import Data.Word
 import GHC.IsList(IsList(..))
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as B
 import qualified Data.Text.Lazy as T
 import qualified Data.Text.Lazy.Encoding as TE
@@ -325,7 +326,7 @@ section s action = do { beginSection s; action; endSection }
 -- in addition to unconsumed bytes and an optional hash of
 -- consumed bytes.
 
-data IS = IS BinTable [Byte] !(Maybe Hash) -- Integer
+data IS = IS !BinTable !BS.ByteString !Int !(Maybe Hash)
 
 -- In monad is a state transformer type monad
 newtype In a = In (IS -> (a,IS))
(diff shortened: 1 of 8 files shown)


src/comp/BinData.hs
```diff
@@ -347,16 +348,15 @@ getN :: Int -> In [Byte]
...
-          internalError "BinData.getB: unexpected end of byte stream"
-        consume_byte (IS bc (b:bs) (Just h)) =
-            -- trace ((show n) ++ ": " ++ (showHex (ord b) "")) $
-            let h' = nextHash h [b]
-            in seq h' $ (b,(IS bc bs (Just h')))
-        consume_byte (IS bc (b:bs) Nothing) =
-          -- trace ((show n) ++ ": " ++ (showHex (ord b) "")) $
-          (b,(IS bc bs Nothing))
+getB = In $ \(IS bc bs off mh) ->
+  if off >= BS.length bs
```
quark17: Would it be more efficient to replace the separate `length` and `index` calls with pattern match of a call to `indexMaybe`?  (Also, `index` is a partial function, so this probably introduces a new warning.)

src/comp/FileIOUtil.hs
```diff
@@ -83,9 +84,9 @@ readFilePath errh pos verb name path =
     readFilesPath' errh pos verb [name] path >>= mapM (decodeUtf8orError errh name)
 
 readBinFilePath :: ErrorHandle -> Position ->
-                   Bool -> String -> [String] -> IO (Maybe ([Word8], String))
+                   Bool -> String -> [String] -> IO (Maybe (BS.ByteString, String))
 readBinFilePath errh pos verb name path =
-    readFilesPath' errh pos verb [name] path >>= return . mapFst B.unpack
+    readFilesPath' errh pos verb [name] path >>= return . mapFst (BS.concat . B.toChunks)
```
quark17: Would it make sense to just to read the file in with `BS.readfile` rather than what's happening here of `BS.concat . B.toChunks. B.readFile`?

Because you changed two of the uses to `B.readFile` to `BS.readFile`, but the one inside `readFilePath_tryFile` is unchanged and is what's used here.  I'm guessing that for non-binary files, you wanted to preserve the use of `B.readFile` so the returned `String` is lazy or something?

src/comp/Util.hs
```diff
@@ -607,12 +607,13 @@ hashInit = Hash 0 4000000063
 -- showpair (x,y) = "(" ++ (showHex x ("," ++ (showHex y ")")))
 
 nextHash :: Hash -> [Word8] -> Hash
-nextHash h s = foldl' f h s
-    where f :: Hash -> Word8 -> Hash
-          f (Hash x y) c =
-              let y' = (rotate x 5) + (toEnum (fromEnum c))
-                  x' = y + y' + 1442968193
-              in Hash x' y'
+nextHash h s = foldl' nextHashByte h s
```
quark17: Are you keeping this around just in case?  `BinData` was the only user and now uses `nextHashByte`, so I believe you could remove it?

If you wanted to be consistent, `nextHashByte` would be `nextHash8` along with `nextHash32` and `nextHash64`.  I notice that neither of those are used -- not sure whether it's worth removing code that we no longer use, or it leave it as a utility that could be used in the future.

quark17 (commented):
Seems reasonable.  Just a few comments.

src/comp/GenABin.hs
```diff
@@ -36,16 +36,20 @@ import qualified Data.ByteString as B
 header :: [Byte]
 header = B.unpack $ TE.encodeUtf8 $ T.pack "bsc-ba-20260418-1"
 
+headerBS :: B.ByteString
```
quark17, line 39: Ah, you pushed changes while I was reviewing :)

What was the reason for deciding to do this the other way around?  Seemed like the other way was cleaner, but no big deal either way.

src/comp/BinData.hs
nanavati: I was originally not using `indexMaybe` because it is newer, but it is safe for the entire CI matrix so there is no issue.

src/comp/FileIOUtil.hs
nanavati: Yes, the difference is deliberate. I want to preserve the laziness on non-decoding paths, which need that laziness, I believe.

src/comp/Util.hs
nanavati: Yes, I didn't previously investigate the uses of the hash functions. Now that only one is uses `nextHashByte` I just kept that and deleted the others (leaving no existing naming convention to match so `nextHashByte` seems fine).

src/comp/GenABin.hs
nanavati: This way we can keep the `[Byte]` as the canonical definition, which the write path still uses. The important thing is that both header versions are top-level CAFs so there should be no reevaluation either way.

nanavati:
Found while verifying the fixups: this branch couldn't reliably build under CI's -M5G heap cap -- GenABin.hs alone peaks at ~16 GB (the strict In monad meets the 135-field Bin Flags chains and the simplifier goes superlinear; predates the fixups). Fixed in the last commit by splitting Bin Flags serialization into NOINLINE chunks of 15 fields -- `.ba` format unchanged. GenABin.hs now builds in ~0.9 GB, and the full compiler passes CI's exact settings locally.

quark17 (approved):

PR #937, opened 2026-04-26
Filter iverilog 12+ "Excess hex digits" warnings from test output

nanavati:


testsuite/config/verilog.tcl
```diff
@@ -281,10 +281,13 @@ proc clean_iverilog_output { infile outfile } {
 
     set readmem_script {-e {/^\$readmem/d}}
     set readmem_script2 {-e {/^WARNING:.*\$readmem/d}}
+    # Iverilog 12+ warns about excess hex digits when $readmemh data
+    # is wider than the target register
+    set readmem_script3 {-e {/^WARNING:.*Excess hex digits/d}}
```
quark17, line 286: Can you show an example of what the message looks like?

I'm wondering if this pattern targeted enough, and whether there's a way to be more precise for readmem.  I guess if we silence other errors, that's probably not a concern for the testsuite, though.

Also, does this warning suggest anything that we should change in our code generation?

nanavati:
Closing in favor of #995, which fixes the one real occurrence at the source.

For the record: the message (iverilog 12.0) is

    WARNING: RegFileLoad.v:57: Excess hex digits (2 of '0x63') while reading 8-bit words.

and it fires on data-file contents, not on anything BSC generates. An audit of
the suite found MPEG4's 'ffffffff' sentinel (fixed by #995) is the only actual
trigger -- the AES .table files would also warn, but only the unused
SboxRom/RconRom variants reference them. So after #995 the filter has
no remaining use case and only serves to hide data errors we should fix.

PR #938, opened 2026-04-26
Fix iMkCons to match the frontend's Cons representation

nanavati:
Use the List_$Cons struct type with _1/_2 field IDs instead of PrimPair, and fix iConType arity from 2-arrow to 1-arrow to match the single struct argument.


krame505 (approved):

quark17 (approved):

quark17:
Interesting.  Out of curiosity, how did you run across this?  It appears to currently only be used in the implementation of primRealToDigits.

PR #939, opened 2026-04-27
Stop using standard library names when naming heap cells

nanavati:


testsuite/bsc.lib/rwire/mkTestENBypassWire2.v1.v.expected
```diff
@@ -40,7 +43,9 @@ module mkTestENBypassWire2(CLK,
 
   // remaining internal signals
   wire [63 : 0] counter_MUL_counter___d5;
-  wire [31 : 0] testWire__read__h212;
+
+  // inlined wires
+  assign testWire$wget = toggle ? counter_MUL_counter___d5[31:0] : 32'd0 ;
```
quark17, line 48: Note that this has exposed the internals of `mkBypassWire`.  The interface visible to the user is register `_read`, but inside the module defined in PreludeBSV, there is a VWire with `wget`.

I suspect that we've decided that `_read` is a Prelude name and therefore refused to use it.  Or maybe that `testWire__read` has picked up the Prelude-ness from `_read`.

testsuite/bsc.names/signal_names/Method.bsv.bsc-vcomp-out.expected
```diff
@@ -43,14 +43,14 @@ rg :: ABSTRACT:  PreludeBSV._PreludeBSV.VRWire110 = RWire
 							       WVAL -> Prelude.Bit 8
 -- AP local definitions
 rg_whas_CONCAT_rg_wget___d3 :: Bit 9;
-rg_whas_CONCAT_rg_wget___d3  = rg_whas____d1 ++ wget__h65;
+rg_whas_CONCAT_rg_wget___d3  = rg_whas____d1 ++ rg_wget____d2;
```
quark17, line 46: This is an interesting change.  The source instantiates `rg <- mkRWire`, so the name `rg` should be in there.  I wonder if `wget` (without `rg`) is due to the `mkRWire` module explicitly constructing an interface struct and returning it, therefore the names are the fields are explicit names that BSC thinks need to be preserved.

testsuite/bsc.misc/sal/CTX_sysMultiArityConcat.sal.expected
```diff
@@ -17,12 +17,12 @@ BEGIN
      #) ;
   
   rule_RL_d (state0 : STATE) : [ BOOLEAN, STATE ] =
-    LET def__read__h37 : Bit{3}!T = CTX_RegUN{3}!meth_read(state0.inst_src1)
-    IN LET def__read__h87 : Bit{5}!T = CTX_RegUN{5}!meth_read(state0.inst_src3)
-    IN LET def__read__h62 : Bit{4}!T = CTX_RegUN{4}!meth_read(state0.inst_src2)
+    LET def_src1___d1 : Bit{3}!T = CTX_RegUN{3}!meth_read(state0.inst_src1)
```
quark17, line 20: Here, the source instantiates `src1 <- mkRegU` and it's interesting that the name before was just `_read` without `src1`.

quark17 (commented):
I'm a little concerned that this is too big of a hammer and hides other issues.

For the most part, there are names `x`, `b`, and `v` that have been inlined.  I believe that these come from function/method argument names that were being preserved but are now being inlined.  Alternate ways to handle this might be: create a new property for argument names and only filter names that are both prelude and argument (so that we're not filtering other names); support "hide" attributes on the arguments in the source.  I also think it'd be worth replicating one of the examples but in user code (not prelude code), which should have the same naming issues; in case we want to address this differently in the future (both all code, not just prelude).

There are some places where method names are being inlined -- in one place, I do think it's exposing internal implementation, replacing the names the user knows.  (I also think that's definitely a situation that could occur in user non-prelude code, that we'd eventually want to address directly, and should make an example for it.)  I think this is different from the argument names case, so a more targeted filtering for that would not filter these.

I don't know what the right answer is and I'm not opposed to this PR as is.

quark17 (approved):
It'd be nice to understand why we see the internals of `mkBypassWire` exposed (in the change I noted in `bsc.lib/rwire`) and whether it could be prevented; but really we're dealing with a pile of heuristics that should be redone someday anyway, so I'm OK with this.  (If anyone has examples of signal names that they'd like to be better, feel free to submit them; we could start a directory in the testsuite even.)

PR #940, opened 2026-04-27
Reduce type normalization in IExpand by preserving type information

nanavati:
This is on top of #939 because it cleans up signal names that change would otherwise disturb.


quark17:
Looks reasonable and good to clean things up.

nanavati:
Redid this so it no longer depends on #939. With the naming heuristic fix (which fixed some issues with expressions not being CSEd properly), the remaining name changes seem reasonable enough.

testsuite/bsc.scheduler/relax-schedule/IBuffer2.bsv.bsc-sched-out.expected
```diff
@@ -73,7 +73,7 @@ order: [handleIBufferRequest,
     1)]),
  (responseQueue_rq_the_fifof.enq,
   [(if responseQueue_rq_enqw_whas__4_AND_responseQueu_ETC___d22 then
-      responseQueue_rq_the_fifof.enq responseQueue_rq_enqw_wget____d17,
+      responseQueue_rq_the_fifof.enq v__h1181,
```
quark17: Any clue why this now has an IExpand name (`__h`) that's kept, instead of being replaced in AConv (`__d`) with a name based on the expression?

src/comp/AVeriQuirks.hs
```diff
@@ -286,14 +286,26 @@ aQExp top (APrim aid t p es) | p == PrimAdd || p == PrimSub = do
 -- All multipliers should already be assigned directly to a
 -- def, and they should all have output size equal to the
 -- sum of the input sizes.  We check those properties here.
+-- The arguments must be in named defs (via mkDefS) so that Verilog
```
quark17, line 289: This looks like a functional change for Mul/Rem/Quote.  Do you have test cases that demonstrate the difference in Verilog behavior?

testsuite/bsc.scheduler/relax-schedule/IBuffer2.bsv.bsc-sched-out.expected
nanavati: `v` is a source variable (which is why the naming heuristic prioritises it), but it's a library source variable from `Vector.bs` (e.g. `pack (V v) = packArray v`) that leaks through inlining as a keep-marked heap-cell name.  We shouldn't be preferring those.

 This is what #939 is for.  I confirmed that by rebuilding with #939 on top, where the line goes back to AConv's `__d` name: responseQueue_rq_the_fifof.enq responseQueue_rq_enqw_wget____d17

src/comp/AVeriQuirks.hs
nanavati: I just added some.

quark17:
Is this PR replaced by #957? or in addition to it?  I'm happy to merge this, but only if it's still valid?

FYI, I just rebased this on the HEAD of main -- that eliminated the need for changing testsuite expected files (because of a naming heuristic in 939, which was previously merged).

src/comp/IExpandUtils.hs
```diff
@@ -2758,81 +2767,58 @@ unheapAllNFNoImp e = do
...
 -- Used when you absolutely need to get an IRefT back
 -- for arrays
 {-# INLINE toHeapCon #-}
-toHeapCon :: String -> HExpr -> Maybe Id -> G HExpr
+toHeapCon :: String -> IType -> HExpr -> Maybe Id -> G HExpr
 -- expand out ICDef as in toHeap
-toHeapCon tag (ICon i (ICDef t e)) cell_name = do
-  e' <- cacheDef i t e
-  toHeapCon tag e' cell_name
+toHeapCon tag t (ICon i (ICDef t' e)) cell_name = do
```
quark17, line 2797: We probably want the same "Don't use t for caching because" comment here, as on `toHeap`?

quark17:
I notice that this has a commit that adds tests for mul/quot/rem Verilog quirks.  Those relate to PR 939, and I guess were left out of that PR.  Happy to include them here.

PR #941, opened 2026-04-28
docs: add Terra repository as a way to install ghc-strict-concurrency packages

Owen-sz:
I recently added ghc-strict-concurrency to Terra, a third-party Fedora package repository - https://github.com/terrapkg/packages/tree/frawhide/anda/langs/haskell/ghc-strict-concurrency

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -198,10 +198,11 @@ sudo dnf install \
    ghc-old-time-prof \
    ghc-split-prof
 ```
-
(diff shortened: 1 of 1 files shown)


kenta2:
I have confirmed that bsc builds on Fedora 43 using `ghc-strict-concurrency-devel` from the Terra repository.

kenta2 (approved):

Issue #243, opened 2020-08-05
Literal '1 not accepted in case match pattern


arjenroodselaar:
I am trying to use the literal '1 as a case match pattern but am presented with the following error:
```
Error: "Schmitt.bsv", line 17, column 17: (P0097)
  Numeric literals with X or Z not supported: `'1'
```

For completeness a minimal-ish repro:
```
package Schmitt;

interface Schmitt #(numeric type sz);
    method ActionValue#(Bit#(1)) filter(Bit#(1) x);
endinterface

module mkSchmitt (Foo#(sz))
        provisos (Add#(msb, 1, TLog#(sz)));
    Reg#(Bit#(sz)) xs <- mkReg('1);
    Reg#(Bit#(1)) x <- mkReg(1'b1);

    method filter(x);
        let _xs = Bit#(Bit#(sz))'{xs[msb:1], x};
        let _x =
            case (_xs) matches
                '0: return 1'b0;
                '1: return 1'b1;
                default: return x;
            endcase;

        xs <= _xs;
        x <= _x;
        return _x;
    endmethod
endmodule : mkSchmitt

endpackage : Schmitt
```

The use case above where this came up is in a schmitt trigger like filter with a polymorphic sampling window size. In this case I want to match the "all zeros" and "all ones" cases, and according to paragraph 2.3.1 in the Reference Guide these literals would do just that. I can use the '1 literal in other places, as demonstrated by the initialization of "xs". Am I doing this wrong? And if so, what's the appropriate way to implement this functionality?

bpfoley:
The BSV parser explicitly disallows `'1` and `'0` in pattern literals at the moment. You can see the code that forbids it in [CVParser.lhs:2723](https://github.com/B-Lang-org/bsc/blob/master/src/comp/Parser/BSV/CVParser.lhs#L2723).

A workaround for now is to use pattern guards.

```
import Vector::*

      rule do_update;
          let _xs = Bit#(sz)'{xs[valueOf(msb):1], _x0};
          xs <= _xs;
          Vector#(sz,Bit#(1)) vz = replicate(0);
          Vector#(sz,Bit#(1)) vo = replicate(1);
          let zeros = pack(vz);
          let ones  = pack(vo);
          x <=
             case (_xs) matches
                 .* &&& (_xs == zeros): return 1'b0;
                 .* &&& (_xs == ones):  return 1'b1;
                 default: return x;
              endcase;
      endrule
```
```

[...]

PR #945, opened 2026-04-28
Unbased unsized literals in BH

matx-chris:
This PR adds the `'0` and `'1` syntax for unbased unsized literals to the Bluespec Haskell parser.

Notes:
* `'` is an identifier character, so a literal like this must start after a token boundary; i.e., `foo'0` is an identifier, but `foo '0` is a function applied to an unbased unsized literal.
* '0' is still a character literal
* This is similar to a lot of Haskell syntax for TH and Data Kinds promotion, it doesn't actually conflict with any valid Haskell syntax, which is nice.

### DIFF
--- a/src/comp/Lex.hs
+++ b/src/comp/Lex.hs
@@ -66,6 +66,8 @@ data LexItem =
         | L_lcurl_o | L_rcurl_o | L_semi_o
         -- pragma
         | L_lpragma | L_rpragma
+        -- unbased unsized bit literals ('0 all-zeros, '1 all-ones)
+        | L_unbasedUnsized Bool
         -- pseudo items
         | L_eof
         | L_error LexError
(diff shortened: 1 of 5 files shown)


nanavati:
LGTM, my only question is how does this compare to what BSV does with '0 ad '1?

matx-chris:
> LGTM, my only question is how does this compare to what BSV does with '0 and '1?

I believe it does the same thing.  I followed the same implementation.  They both produce `cVar (idConstAllBitsUnsetAt pos)` and `cVar (idConstAllBitsSetAt pos)` in the parser.  There's a slight difference in `'x` and `'z`, which are accepted by the BSV lexer, but then rejected in the parser.  Here, I didn't accept them in the lexer either.  The reason is that I wanted to minimize the syntax grab.

matx-chris:
I realized I missed one case, which is `'0` as a pattern.  (Note: `'1` as a pattern is an error in BSV as well, so this only applies to `'0`.)  Added that as well, for consistency.

quark17:
> I realized I missed one case, which is `'0` as a pattern. (Note: `'1` as a pattern is an error in BSV as well, so this only applies to `'0`.) Added that as well, for consistency.

I don't know why `'1` in pattern matching is an error in BSV -- note that this is Issue #243.

matx-amy:
reviewed, :+1:

rsnikhil:
Looks good to me.

rsnikhil (approved):
Approved (comments, if any, in PR comment list)

PR #947, opened 2026-05-06
Fix a bug where unused import warnings were issued for uses removed in ctxreduce

matx-chris:
The unused import warnings was computed after ctxreduce, which caused some actual uses to be missed.  A test case to demonstrate this is included, and the change fixes that test case.

### DIFF
--- a/testsuite/bsc.driver/unused_imports/unused_imports.exp
+++ b/testsuite/bsc.driver/unused_imports/unused_imports.exp
@@ -75,6 +75,10 @@ compile_pass_warning_bug TransitiveSynonymBSBad.bs T0157
 compile_pass_no_warning TypeclassConstraint.bsv
 compile_pass_no_warning TypeclassConstraintBS.bs
 
-# Test 14: Ambiguous constructor - should warn about Helper2
+# Test 14: type alias used only in a type constraint - should NOT warn
+compile_pass TypeAliasConstraintHelperBS.bs
+compile_pass_warning_bug TypeAliasConstraintBS.bs T0157
+
+# Test 15: Ambiguous constructor - should warn about Helper2
 compile_pass_warning AmbiguousConstructor.bsv T0157 1
 compare_file AmbiguousConstructor.bsv.bsc-out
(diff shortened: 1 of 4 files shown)


krame505 (commented):
Looks good to me.

testsuite/bsc.driver/unused_imports/unused_imports.exp
```diff
@@ -75,6 +75,10 @@ compile_pass_warning_bug TransitiveSynonymBSBad.bs T0157
 compile_pass_no_warning TypeclassConstraint.bsv
 compile_pass_no_warning TypeclassConstraintBS.bs
 
-# Test 14: Ambiguous constructor - should warn about Helper2
+# Test 14: type alias used only in a type constraint - should NOT warn
```
quark17: Would you mind correcting the comment here?  It's not just that it's used only in a constraint, but that it's only used in a constraint that is resolved (and thus removed) during CtxReduce.

quark17 (commented):
Looks good.  Just a minor request to fix a comment, if you don't mind?

Nothing is special about aliases: any type that is only used in a removed context would be missed.  But if you do that, then you're probably using other names from the file elsewhere, so BSC wouldn't consider the import unused.  The only example I can think of is if you just have an unnecessary context:
```
package Imp where
data Foo = Foo
instance DefaultValue Foo where
  defaultValue = Foo

package Top where
import Imp
f :: (DefaultValue Foo) => Integer
f = 0
```

If we eventually allowed selective importing and reported warnings about individual names that are unused, then it would be visible, such as the following (where `Foo` might be incorrectly warned about):
```
package Top where
import Imp(fooToInteger, Foo)
f :: (DefaultValue Foo) => Integer
f = fooToInteger defaultValue
```

testsuite/bsc.driver/unused_imports/unused_imports.exp
matx-chris: Done.

quark17 (approved):

PR #948, opened 2026-05-13
ValidateBits type class

### LINKED ISSUE #879: Type class for validating Bit values are legal packed values for a type's Bits instance
We would like to add a type class `ValidateBits` to check whether a `Bit` value is, in fact, valid for its `Bits` instance:
```
class (Bits a n) => ValidateBits a n where
  validateBits :: Bit n -> Bool
```
This class should probably be derivable and not implemented using generics, since we do not want generic default implementations for `ValidateBits` when there is a custom manual `Bits` implementation.

matx-chris:


src/Libraries/Base1/Prelude.bs
```diff
@@ -3365,6 +3378,10 @@ instance Bits File 32
                  1 -> FD x[30:0]
                  0 -> MCD x[30:0]
 
+instance ValidateBits File 32
+  where
+    unpackValid x = Valid (unpack x)
```
quark17, line 3383: It might better to derive ValidateBits, because this instance seems incorrect.  The type has 3 tags, which needs 2 bits, so there's one tag value that's unused and should return Invalid, I assume?  (And be tested in the test file.)

src/Libraries/Base1/Prelude.bs
```diff
@@ -1634,6 +1641,9 @@ instance Bits (Int n) n where
     pack (Int x) = pack x
     unpack b = Int (unpack b)
 
+instance ValidateBits (Int n) n where
+    unpackValid b = fmap Int (unpackValid b)
```
quark17, line 1645: For `Int` and `UInt`, would there be any efficiency improvement to just defining this as `Valid b`?  I assume that recursively calling `unpackValid` is conceptually cleaner and safer?

matx-chris: My thought was that it was better to be obviously correct, and the performance here isn't likely to matter and likely goes away with inlining anyway.  I could definitely be wrong.  If you'd prefer not to delegate to the Bit instance, I'd be fine with doing it directly as well.

src/Libraries/Base1/Prelude.bs
matx-chris: I think we want this to match the `Bits` instance, where `unpack` is total.  There are three constructors, but there's still one tag bit, and `InvalidFile` is handled as a special case.  (Note: `MCD 0` doesn't round-trip; after `pack` and `unpack` it will land as `InvalidFile`, but that is an existing quirk of the `Bits` instance, not new here.)

In general, I'd say we should never derive `ValidateBits` unless `Bits` is also derived.  If there's a hand-written `Bits`, then we also want a hand-written `ValidateBits` that matches.  Perhaps we should even make this a warning.

matx-chris (commented):
FYI, this isn't really done yet, but thanks for the early review!

src/Libraries/Base1/Prelude.bs
quark17: Oh, you're totally right.  I hadn't looked at the Bits instance.  Thanks for pointing that out.

src/Libraries/Base1/Prelude.bs
quark17: Sounds good.  And it matches the Bits instance, which I had overlooked.

PR #951, opened 2026-05-20
Don't rewrite a file if the content is the same

flafflar:
I wanted to use the Meson buildsystem to build my bluespec project, but bsc unnecessarily touches files, even though the contents are the same as before. This PR fixes this by checking if the file contents are the same before writing a file.

### DIFF
--- a/src/comp/FileIOUtil.hs
+++ b/src/comp/FileIOUtil.hs
@@ -51,6 +51,7 @@ import Data.Word
 import qualified Data.Text.Lazy as T
 import qualified Data.Text.Lazy.Encoding as TE
 import qualified Data.Text.Encoding.Error as TEE
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as B
 
 import Util(concatMapM, mapFst)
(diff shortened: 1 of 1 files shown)


quark17:
Can you explain more specifically the situation where BSC needs to construct some files but is unnecessarily touching others?  Which types of files does this occur with?  Are you using the `-u` flag or not using it?

If someone is using `-u`, then BSC might choose to recompile something if a dependency has a newer date, and in that case the user might want BSC to update the file's date, even if the contents are the same.  (Should this code be conditional on whether the flag is given, then?)

In any case, a testcase would be good to have, for the test suite.  Otherwise, the code seems reasonable to me (aside from any decision about whether this should be more targeted), but I don't have a sense of whether there's any performance concerns about the way it's written, so I'd like to get someone else's opinion on it.

flafflar:
When compiling a `.ba` file, bsc will also rewrite the `.bo` file, even though the contents are the same. This will update the timestamp of the `.bo` file, so every file that depends on that package will be recompiled, even though the file didn't really change.

I don't use `-u` because I use a build system (meson) to build my project, so I don't know if it would affect that in any way. To be honest, I can't think of a scenario where you would want to touch a file that has not changed.

PR #952, opened 2026-05-21
Add Applicative instance for Vector (and ListN)

krame505:
Fixed-size collection types like `Vector` and `ListN` have a natural, rather useful `Applcative` instance, where `pure` is `replicate` and `liftA2` is `zipWith`.

Note that there is also technically a `Monad` instance possible for `Vector`, where the intuition is that `join` would take the diagonal elements of a 2D vector.  `Monad` in `Bluespec` doesn't have `join`, only `bind`, which would look like this:
```
instance Monad (Vector n) where
  bind xs f = genWith $ \ i -> f (xs !! i) !! i
```
I chose not to include it as it does not seem especially useful; I can't think of a reasonable use case that calls for more than just `Applicative`.  It also quite inefficiently constructs n^2 elements and throws most of them away, and the potential for confusion seems high.  But I could be convinced otherwise.


matx-chris (approved):

quark17 (dismissed):

quark17:
Looks fine.  Can we add testcases?

krame505:
I added testcases.

quark17 (approved):

PR #953, opened 2026-05-21
Add Foldable and Traversable utilties

krame505:
Add `Foldable` and `Traversable`, ala Haskell.  Also includes some Bluespec-specific fold variants in the type class.  `Foldable` is split into `Foldable` and `Foldable1`, since the non-empty fold variants on `Vector` and `ListN` have an extra `Add` proviso.

This is based on #952 because it requires the missing `Functor ListN` instance.


src/Libraries/Base1/Traversable.bs
```diff
@@ -0,0 +1,56 @@
...
+instance Traversable (Vector.Vector n) where
+  traverse f v = fmap Vector.toVector (traverse f (Vector.toList v))
+  mapM = Vector.mapM
+  sequence = Vector.sequence
+
+instance Traversable Maybe where
+  traverse _ Invalid   = pure Invalid
+  traverse f (Valid x) = fmap Valid (f x)
+
+-- The `_` variants discard the rebuilt container and return `f ()`.
```
matx-chris: It can typically be much more efficient to implement these directly and avoid state threading just to discard the state.  Are you sure we shouldn't move them into the type classes, even if we don't implement them directly yet?

The _ variants could actually be moved to Foldable, too.  They shouldn't need Traversable as far as I can tell.

krame505: Good catch, I guess the `_` functions only require `Foldable` in Haskell too.  They aren't members of the type class in Haskell, but I see no reason not to allow specializing them.

matx-chris (approved):

nanavati:
I want to emphasize that we have quite a bit of room to add specialized primitives for many Foldable and Traversable operations. Make sure everything we might want to implement that way is in the typeclasses (because they are a good target when deciding what to optimize).

krame505:
I added `null` and `length` since those are also in Haskell's Data.Foldable.  Do you think it is worthwhile adding `maximum`, `minimum`, `sum`, `product`?

krame505:
I guess there is no harm in adding more methods later on, as long as they have default implementations.

quark17 (dismissed):
Looks fine.  Any chance of testcases (I know it's a lot to test)?

krame505:
I added some testcases.  It looks like you need to re-approve this (and #952) now.

quark17 (approved):

PR #954, opened 2026-05-23
Internal evaluator type tracking fix for PrimStringSplit

nanavati:


testsuite/bsc.codegen/strings/strings.exp
```diff
@@ -26,6 +26,10 @@ compare_file CharLiteralBad_Multiple.bsv.bsc-vcomp-out
 # Test non-error behavior
 test_veri_only_bsv StringCharFunctions
 
+# Verify string prims produce correctly-typed results
+compile_verilog_pass StringCharFunctions.bsv {} {-trace-eval-types}
```
quark17, line 32: What's being tested here?  There's no check of expected output; are you just testing that the trace flag doesn't cause a failure?

The file `StringCharFunctions` is separately being tested for its output, above.  But the new file `StringSplitTypes` is not being tested that way.

quark17 (commented):
The evaluator was clearly giving the wrong type -- was that not a problem, because the value was only ever consumed without needing to consult the type?  Until, maybe, you tried to trace the types?  Just trying to understand what you observed that brought this to your attention.

Fix looks fine.  Just want to make sure the testsuite changes are sufficient.

testsuite/bsc.codegen/strings/strings.exp
nanavati: Yes, turning on the trace flag turns on more internal type checking in the evaluator (which is how I ran into the issue in the first place - I turned on the checking to deal with something else and it got stuck on this issue) and we're verifying that doing that doesn't create problems.

quark17 (approved):

PR #955, opened 2026-05-23
Cache ATF resolutions for fullTypeNormalizer

nanavati:
## Summary
- Cache associated type function (ATF) resolutions computed during context reduction and type checking, and use them in `fullTypeNormalizer` to avoid redundant type normalization during IExpand
- Falls back to the `iConvT`/`iToCT` path on cache miss (for polymorphic resolutions that can't be cached)
- Adds `-trace-atf-cache` and `-trace-atf-cache-miss` debug flags for monitoring cache behavior
- Replaces fragile `findWithDefault (-1)` lookups in `MakeSymTab` with `fromJustOrErr` for better error reporting

## Test plan
- [x] Compiler builds successfully (all binaries: bsc, bluetcl, bsv2bsc, showrules, vcdcheck, dumpbo, dumpba)
- [ ] Run existing testsuite
- [ ] Profile on regression designs to confirm speedup


krame505 (dismissed):
Looks good to me.  How much of a performance win did this end up being?

src/comp/CtxRed.hs
```diff
@@ -17,31 +18,33 @@ import TCMisc
...
 
 import Debug.Trace(traceM)
 import IOUtil(progArgs)
 
 doTraceCtxReduce :: Bool
 doTraceCtxReduce = "-trace-ctxreduce" `elem` progArgs
 
-cCtxReduceIO :: ErrorHandle -> Flags -> SymTab -> CPackage -> IO CPackage
+cCtxReduceIO :: ErrorHandle -> Flags -> SymTab -> CPackage ->
+               IO (CPackage, S.Set Id, CATFCache)
```
quark17, line 38: Perhaps a comment on this function, just briefly saying what the return values are

src/comp/ISyntaxCheck.hs
```diff
@@ -86,17 +92,24 @@ eqTypeNum flags symt r t1 t2
...
-          addBoundTVs (M.elems m)
+          addBoundTVs boundTVs
           addExplPreds eqs
           vp <- mkVPredFromPred [] (IsIn numEqCls [t1', t2'])
-          satisfy eqs [vp]
-    in  case runTI flags False symt satisfyEq of
-          (Right ([],_), _, _) -> True
-          res -> --trace("eqTypeNum: not satisfied: " ++ ppReadable (t1, t2, res)) $
-                 False
+          satisfyFV boundTVs eqs [vp]
```
quark17: Is this change from `satisfy` to `satisfyFV boundTVs` related to caching or a separate fix/improvement?

testsuite/bsc.evaluator/primtcons/primtcons.exp
```diff
@@ -28,3 +28,5 @@ test_primtcon_normalization "TestTMul_VectorReg.bs" "mkTest_TestTMul_VectorReg"
 test_primtcon_normalization "TestTMul_StatefulIndex.bs" "mkTest_TestTMul_StatefulIndex"
 test_primtcon_normalization "TestTLog_UIntParam.bs" "mkTest_TestTLog_UIntParam"
 test_primtcon_normalization "TestTLog_TypeVar.bs" "mkTest_TestTLog_TypeVar"
+
+test_primtcon_normalization "TestATFCacheMiss.bs" "mkTest_TestATFCacheMiss"
```
quark17, line 32: If I understand, this would result in an internal error (`assert failed: IAp`) and after this PR it will compile fine.  In theory, you could turn on trace flags and confirm that a cache miss is happening, by checking the output, but that's not necessary.

testsuite/bsc.evaluator/primtcons/TestATFCacheMiss.bs
```diff
@@ -0,0 +1,27 @@
...
+
+{-# verilog mkTest_TestATFCacheMiss #-}
+mkTest_TestATFCacheMiss :: Module Empty
+mkTest_TestATFCacheMiss = module
+  r :: Reg (UInt 5) <- mkReg 7
+  rules
+    "test": when True ==>
+      action
+        let result = checkPair r
+        if result then $display "pass" else $display "FAIL"
```
quark17: This display has nothing to do with pass/fail of the test (and in fact I think it will display FAIL currently?) -- perhaps just display `result` here.  Nothing checks the expected output anyway.

testsuite/bsc.typechecker/typeclasses/ATFPolyCacheMiss.bs
```diff
@@ -0,0 +1,42 @@
...
+
+{-# verilog sysATFPolyCacheMiss #-}
+sysATFPolyCacheMiss :: Module Empty
+sysATFPolyCacheMiss = module
+  r :: Reg (UInt 5) <- mkReg 7
+  rules
+    "test": when True ==>
+      action
+        let result = checkPair r
+        if result then $display "pass" else $display "FAIL"
```
quark17: Again, replace this with just a display of `result`, since it's not actually related to pass/fail of the test (and nothing checks it).

testsuite/bsc.typechecker/typeclasses/typeclasses.exp
```diff
@@ -114,6 +114,10 @@ compile_verilog_pass ATFPolyStructIfc.bs
 compile_verilog_pass ATFPolyStructLog.bs
 compile_verilog_pass ATFPolyStructMultiParam.bs
 
+# Test that a user-defined ATF cache miss is resolved by normTFun
+compile_verilog_pass ATFPolyCacheMiss.bs {} "-dexpanded=expanded_%m.out"
```
quark17, line 118: The evaluator test would fail before, but this one compiles fine currently.  So what's being tested here? and how is it different from the other test?

quark17 (commented):
Looks reasonable.  Just a few questions/fixups.

nanavati:
Pushed a81d445e: the cache now records only **coherent** resolutions, and the hit path is pinned by tests (a same-package hit, and a cross-package hit delivered through the `.bo` cache, both asserted via `-trace-atf-cache` — previously only the miss path was tested, so a lost recording site was invisible to a green suite).

Rationale for the incoherent-arm change: an incoherent match is an information-dependent choice (it varies with `-allow-incoherent-matches`, in-scope instances, and import order), so persisting its ATF projection into the `.bo` cache would deliver a context-dependent answer to importing compiles as if it were canonical. Dropping the recording restores the pre-cache confinement without any language-level change.

The broader question — whether ATF-bearing classes should be allowed to opt out of coherence at all — is deliberately **not** addressed here; filed as #1042 for discussion outside the release cycle.

PR #956, opened 2026-05-25
Add wiretypemap command to bluetcl

nanavati:
Bluetcl's port/type correlation is limited. It doesn't always work in the presence of modules using SplitPorts and drops a lot of interior signals that do have useful types.

Besides fixing some crashes, this PR adds a wiretypemap command to bluetcl that includes the type of every wire whose type is known. Depending on optimizations and flags not all the candidate wires will appear in a vcd, but the intent is to capture everything we know. Note that Bluesim doesn't always output the same logical wire the way Verilog would, so there are some extra wire candidates to specifically cover what Bluesim generates.

The wiretypemap is tested against generated VCDs to confirm the VCD correlation is working as expected.


src/comp/bluetcl.hs
```diff
@@ -3569,6 +3600,272 @@ getModPortInfo apkg pps tifc = do
...
+--   * inst.port   (Bluesim scope-relative; correlator concatenates while
+--                  descending into the sub-scope)
+--   * inst        (Verilog reg-Q_OUT shortcut, when reg)
+-- For CReg instances, the same candidates are also emitted after applying
+-- cregToReg's port-name rewrite, so the Reg-converted Verilog names are
+-- covered without depending on which flags were active at build time.
+--
+-- Over-emission is intentional: a correlator building a name -> type
+-- dict can index by VCD wire name and match whichever form appears.
+getWireTypeMap :: APackage -> [HTclObj]
```
quark17: Is it worth pulling out the part of this that is not specific to HTcl, and putting it somewhere else, and only doing the conversion to HTclObj here?

Like how the `TypeAnalysis` package defines the analysis and here we just convert it to Tcl for the Tcl interface.

testsuite/archive_logs.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/sh
 
 find . \
-    -name bsc.log -o \
     -name bsc.sum -o \
+    -name '*.log' -o \
+    -name '*.out' -o \
```
quark17, line 9: The only concern is whether the tar-file gets too large.  I recall that I left some files out, because it was a problem -- but don't remember the specifics.

quark17 (commented):
I haven't looked closely, but seems good at a high level, just one comment about code organisation.

testsuite/archive_logs.sh
nanavati: Not sure, but I know I needed more detail for debugging bluetcl issues so I'd rather have the info than not have it. If it becomes too much of a problem, we can dial back somehow.

quark17: Ok.  I might leave the explicit `bsc.log` in, for documentation purposes.  The localcheck that's run in each directory produces a `bsc.log` and `bsc.sum` file, so we want to make sure to always capture those.

src/comp/bluetcl.hs
nanavati: Done in 2017a127 — the analysis now lives in a new `WireAnalysis` module as `getWireTypeMap :: APackage -> [(String, IType)]` (dedup included in its contract), and bluetcl keeps only the two-line conversion to Tcl objects, mirroring the TypeAnalysis division of labor. I named it WireAnalysis rather than putting it in TypeAnalysis itself: that module is typechecker-world (CType/SymTab/TIMonad) with no ASyntax dependency, while this analysis is entirely about the elaborated APackage — and it covers internal wires (rule fire signals, inlined-primitive wires), not just ports.

krame505 (dismissed):
Did you test this when stacked on #967?

PR #957, opened 2026-05-25
Fix quadratic blowups and stack overflows

nanavati:
This branch addresses several long-standing problems that surface in
  large modules -- particularly those with many CRegs and many local
  defs.

  ### What changed
  
  **Quadratic fixes** (each replaces an `elem` over a list with O(log n)
  `Set`/`Map` membership):

  - `InlineCReg`: `Set` lookup for `creg_ports`.
  - `AExpand.aoptExpandTest`: precompute `instVars` as a `Set` once,
    share via `ExpandData`.
  - `AVerilog.ds7`: `Map.member` on the existing `defmap7` instead of
    `elem` over `M.keys`.

  **Stack-safety fixes** for lazy structures that grow with module size:

  - `SCC.decrList`: `foldr` -> `foldl'`.
  - `Balanced.foldm.recFn`: bang-pattern on the recursive `as1` list
    and force the combined value strictly, so neither builds a deferred
    thunk chain.
  - `IExpand.eqPtrs`: switch the accumulator fold to `foldl'`, force
    the tuple components, and use `SCC.tsort` instead of
    `GraphWrapper.tSortInt` (the latter uses `Data.Graph`'s recursive
    DFS).
  - `ITransform.runCSE`: switch to `Control.Monad.State.Strict`, force
    inserted maps and the running `idNo`.
  - `AOpt.addDef`: same treatment.
  - `SpeedyString.newSString`: switch to `Data.IntMap.Strict`, force
    the inserted maps before `writeVar`.

  **CSE canonical names**: the `eqPtrs` rework introduces a naming
  regression in generated code. To fix it, this PR adopts and
  generalises the heuristic from #940.
  
  **Tests and harness**

  - New tests under `testsuite/bsc.bugs/perf-creg-blowup/` (one `.exp`
    driving three synthetic `.bsv` modules):
    - `CregInlineBlowup` (4096 CRegs x 4 ports, trivial rules) --
      catches the InlineCReg / AVerilog / AExpand quadratics in one go.
    - `CregMemBlowup` (256 CRegs x 4 ports plus heavy per-rule
      arithmetic) -- primarily the AVerilog ds7 quadratic;
      needs `+RTS -K3G -RTS`.
    - `AoptScanBlowup` (8192 plain Regs + 512 rules of arithmetic depth
      128) -- primarily the AExpand quadratic; needs `+RTS -K3G -RTS`.

    The `.exp` documents which fix is each compile's primary signal and
    which others it implicitly requires.  Each compile is invoked with
    `-elab` so the harness also exercises `.ba` generation.

  - `check_intermediate_files` now also verifies `.ba` files for sim
    compiles and verilog compiles invoked with `-elab`.

  ### Impact
  
  On a real-world module of ~7,000 CRegs and ~800,000 local defs that
  motivated this work, the AOpt phase dropped from "10+ minutes and
  counting" to under three minutes; total compile went from `2242s`
  (incomplete) to `~1500s` (complete). The strictness fixes also drop
  the stack budget needed for moderate inputs significantly.


quark17:
Does this replace #940 then?

testsuite/config/unix.exp
```diff
@@ -1533,16 +1533,58 @@ proc dumpbo_pass { source xfail } {
...
+            }
+            # Resolve relative to the test's subdir (bsc_compile_*
+            # cd's back to srcdir after running bsc, so a bare relative
+            # path would miss the file).
+            if {[file pathtype $ba_file] eq "relative"} {
+                set check_path [file join $test_dir $ba_file]
+            } else {
+                set check_path $ba_file
+            }
+            if {[file exists $check_path]} {
```
quark17, line 1640: For `.bo` files, we run `dump_bo` to verify that the file is readable.  (Only if the utility exists; when it doesn't, we do no check, not even that the file exists.)

Here, you're only checking that the `.ba` file exists.  We could run `dump_ba` to also check that the file is readable.

What led you to want to check for the `.ba` file?  Do you want to also run `dump_ba` on them?

quark17 (commented):
Looks good.  I would check whether the expected-file updates go away when rebased on the HEAD of main.

testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
```diff
@@ -70,66 +70,54 @@ module mkDesign(clk,
...
-	     { 1'd0, notB__h260 } +
-	     6'd1 ;
+	     { 1'd0, partRem4__h722 } + { 1'd0, notB__h260 } + 6'd1 ;
   assign _0_CONCAT_IF_0_CONCAT_IF_0_CONCAT_result_a_BITS_ETC___d27 =
-	     { 1'd0,
-	       IF_0_CONCAT_IF_0_CONCAT_result_a_BITS_8_TO_4_P_ETC___d24 } +
-	     { 1'd0, notB__h260 } +
-	     6'd1 ;
+	     { 1'd0, partRem3__h675 } + { 1'd0, notB__h260 } + 6'd1 ;
   assign _0_CONCAT_IF_0_CONCAT_result_a_BITS_8_TO_4_PLUS_ETC___d18 =
```
quark17, line 116: Interesting that the def name is still based on the old expression?

quark17: You updated `iTransFixupDefNames`, but I see that `AConv` has very similar code, which you did not similarly update.  Perhaps that accounts for what we're seeing (as `__d` names are from AConv, which creates the "names from AExpr").

See this comment in AConv:
```
-- Restore original names to CSEd definitions, when possible.
-- But prevent non-CSE defs from becoming CSE'd into other expressions.
-- Similar to what is being done in ITransform, in iTransFixupDefNames.
```

nanavati:
> Does this replace #940 then?

No it generalizes the heuristic #940 uses and applies it in a new place (eqPtrs).

testsuite/config/unix.exp
nanavati: I think I checked for the .ba file because sometimes blowups led to crashes that didn't generate it. Adding `dumpba` checks to parallel the `dumpbo` checks is a good idea and I've done that.

testsuite/bsc.bugs/bluespec_inc/b302/mkDesign.v.expected
nanavati: I'm experimenting with a new Id quality tier for these RHS-derived names in #1005

quark17:
Note that, when I rebased this, I deleted the commit that updated `bsc.bugs/bluespec_inc/b302/`.  That was an update to names in the expected Verilog output, but those changes were reverted by the final commit, so were no longer needed.

PR #958, opened 2026-05-26
Add type-level PrintType utility

krame505:


src/Libraries/Base1/PrintType.bs
```diff
@@ -0,0 +1,148 @@
...
+-- saturated).  We fully apply the constructor to dummy args (`()` or `0`)
+-- and look at its Generic representation, then drop the dummy tail.
+
+class (PrintConType :: * -> $ -> *) c s | c -> s where {}
+
+instance (PrintConType' (Rep c) s) => PrintConType c s where {}
+
+class (PrintConType' :: * -> $ -> *) r s | r -> s where {}
+
+-- We saturate the constructor with dummy args (`()` or `0`); for nullary,
```
nanavati, line 132: Do you also saturate with ""?

nanavati: Oh, I see you saturate with the ones you know about.

nanavati (approved):
This will be handy. Does this mean TypeError is coming soon?

testsuite/bsc.lib/PrintType/sysTestPrintType.out.expected
```diff
@@ -0,0 +1,28 @@
+Prelude.Bit 8
```
matx-chris, line 1: IMO it would be really great if it would omit `Prelude.` on prelude types.

krame505: I made this match the value-level `printType`.  IMO if we want to remove the prefix for `Prelude` types then we should change both?

matx-chris: Okay, this is fine, then.  I agree we can make the change in both places at once, if we do (and not in this PR, in any case)

quark17 (approved):

PR #928, opened 2026-04-21
BackendNamingConvensions: fixed issue with multiplicity on methods with multiple inputs

bitmappergit:
On methods with more than one input the ordering of these bindings in the list comprehension caused them to not match the ordering of the mapping in AVerilogUtils.
 
I ran into this issue while implementing a multi-port register file with multiple write ports (not sure how well this is going to work anyway but better to not have this bug I guess!)

Essentially BackendNamingConventions was producing this before:
(["upd_0_1","upd_0_2","upd_1_1","upd_1_2","upd_2_1","upd_2_2","upd_3_1","upd_3_2","upd_4_1","upd_4_2","upd_5_1","upd_5_2","EN_upd_0","EN_upd_1","EN_upd_2","EN_upd_3","EN_upd_4","EN_upd_5"],["WADDR_1","WADDR_2","WADDR_3","WADDR_4","WADDR_5","WADDR_6","WDATA_1","WDATA_2","WDATA_3","WDATA_4","WDATA_5","WDATA_6","WEN_1","WEN_2","WEN_3","WEN_4","WEN_5","WEN_6"])

After this change:
(["upd_0_1","upd_1_1","upd_2_1","upd_3_1","upd_4_1","upd_5_1","upd_0_2","upd_1_2","upd_2_2","upd_3_2","upd_4_2","upd_5_2","EN_upd_0","EN_upd_1","EN_upd_2","EN_upd_3","EN_upd_4","EN_upd_5"],["WADDR_1","WADDR_2","WADDR_3","WADDR_4","WADDR_5","WADDR_6","WDATA_1","WDATA_2","WDATA_3","WDATA_4","WDATA_5","WDATA_6","WEN_1","WEN_2","WEN_3","WEN_4","WEN_5","WEN_6"])

PR #959, opened 2026-05-28
Port splitting part 3

krame505:
* Avoid flattening all input ports into curried method arguments, and instead preserve tuples of port values for inputs all the way through ASyntax as ATuples.
* Add handling of port splitting for `noinline` functions, introducing a `primNoInline` in the evaluator to tag the implementation with the port names, like is done with `primMethod` for method implementations.
* Some additional cleanup in ASyntax land.

This is based on the output port splitting PR (#849), and should probably be merged concurrently with it, but for ease of review I am leaving this change out of that branch.  The diff of what is changed in this pull request can be viewed at https://github.com/MatX-inc/bsc/compare/split-output-ports...Matx-Inc:bsc:input-ports-tuples-v2.

Note that I made an alternative version of the input port change that did not use ATuples for the input ports, and instead changed `ACall` to take a list of lists of `AExpr`s, which can be found at https://github.com/MatX-inc/bsc/compare/split-output-ports...krame505:bsc:input-ports-tuples.  ~~That is arguably more elegant but is a somewhat more invasive change, and adds more overhead on the single-port-argument case.~~  Edit: There are now additional changes in this pull request (fixing noinline, other ASyntax cleanup) that are not present in that branch, so the comparison may not be very helpful.


quark17:
Something's not working in the comparison link.  And the commits don't seem to line up with part 2, so maybe it needs to be updated before I can review it?

krame505:
I rebased #849 on main and this PR hasn't been rebased on #849 yet, which is maybe why you are seeing different commits?  Still, the comparison link seems to work fine for me? https://github.com/MatX-inc/bsc/compare/split-output-ports...Matx-Inc:bsc:input-ports-tuples-v2

quark17:
The link works, but shows many differences which are not in this PR, which drowns out what's actually to be reviewed in this PR.

krame505:
Ah, you are right.  I guess this needs to be rebased again - will do that as soon as I've finished my pass over your comments on the #849.

krame505:
Okay, this is now cleanly rebased on #849 and the diff link works.  This should now be ready for review.

src/comp/GenABin.hs
```diff
@@ -34,7 +34,7 @@ import qualified Data.ByteString as B
 -- .ba file tag -- change this whenever the .ba format changes
 -- See also GenBin.header
 header :: [Byte]
-header = B.unpack $ TE.encodeUtf8 $ T.pack "bsc-ba-20260427-1"
+header = B.unpack $ TE.encodeUtf8 $ T.pack "bsc-ba-20260616-1"
```
quark17: The header will need to be bumped again in this PR.

src/comp/AState.hs
```diff
@@ -998,41 +1025,61 @@ mkEmuxss tl cnd rdb value_method_ids om (((o, m), f), emrss) =
...
-mkEmuxs :: ([AExpr] -> [AExpr]) -> ([AExpr] -> AExpr) ->
+mkEmuxs :: ([AExpr] -> [AExpr]) -> ([AExpr] -> [AExpr]) ->
+           ([AExpr] -> AExpr) ->
            ExclusiveRulesDB -> [AId] -> OrderMap ->
            AId -> AId -> Maybe Integer -> MethPortBlob ->
            ([ADef], [ADef], [ADef], AExprSubst)
-mkEmuxs tl cnd rdb value_method_ids om o m ino emrs =
-    let meth_id = mkMethId o m ino MethodResult
-
+mkEmuxs tl tlG cnd rdb value_method_ids om o m ino emrs =
```
quark17: There is no reason for a new `tlG` argument -- the only callers provide the same function for both `tl` and `tlG` -- so just use `tl` in both places.

src/comp/ASyntax.hs
```diff
@@ -533,23 +543,25 @@ type AInput = (AId, AType)
 data AAbstractInput =
         -- simple input using one port
         AAI_Port AInput |
+        -- a single source argument that is split into multiple hardware ports
+        -- (e.g. a struct argument whose fields each become a port).
+        AAI_MultiPort [AInput] |
```
quark17: Why a new constructor?  Why not change the data in `AAI_Port` to be a list?

The only place where they're created is here (AConv):
```
case convGroup group of
    [p] -> AAI_Port p
    ps  -> AAI_MultiPort ps
```

And the only place where they're consumed is here (ASyntax):
```
absInputToPorts (AAI_Port p) = [p]
absInputToPorts (AAI_MultiPort ps) = ps
```

src/comp/ASyntax.hs
```diff
@@ -773,21 +787,21 @@ instance NFData AAssumption where
 
 -- the APred is the implicit condition to the scheduler
 data AIFace =   AIDef { aif_name      :: AId,
-                        aif_inputs    :: [AInput],
+                        aif_inputs    :: [AAbstractInput],
```
quark17: Here (and for `AIAction` and `AIActionValue` below), I think `AAbstractInput` is the wrong type to use -- since it includes clocks, resets, inout.  If what you want to express is that each input can now be lists of ports, then just change this to be `[[AInput]]`.

src/comp/ASyntax.hs
```diff
@@ -1854,7 +1893,12 @@ instance PPrintExpand AExpr where
                <> if (null es) then empty else (parens (hsep ( punctuate comma docArgs )) )
                    where
                    docArgs = map (pPrintExpand m d defContext) es
-    pPrintExpand m d ec (AMethValue _ i meth) = pPrint d 1 i <> text "." <> ppMethId d meth
+    pPrintExpand m d ec (AMethValue _ i meth) =
+        pPrint d 1 i <> text "." <> ppMethId d meth
+    pPrintExpand m d ec (ATuple _ es) =
+        pparen (useParen ec) $ parens (commaSep (map (pPrintExpand m d defContext) es))
```
quark17: There is no need for a surrounding `pparen`, which would add redundant parentheses.

src/comp/ASyntax.hs
```diff
@@ -1854,7 +1893,12 @@ instance PPrintExpand AExpr where
...
+    pPrintExpand m d ec (ATupleSel _ e idx) =
+        pparen (useParen ec) $ pPrintExpand m d defContext e <> text ("[" ++ itos idx ++ "]")
```
quark17: The `pPrintExpand` call on `e` should use `pContext` (not `defContext`) to force parentheses around the expression if needed (so the brackets are applied to the whole thing).

I'm also not sure that a surrounding `pparen` is needed -- is there ever a case where the surrounding context would need parens around `foo[n]` for disambiguation?

src/comp/BackendNamingConventions.hs
```diff
@@ -559,37 +559,41 @@ createMapForVMod inst_id (Method meth_id _ _ mult ins mo me) = -- trace (ppReada
...
-      -- make the names x_<port>_n for multi-ported methods
-      method_input_names = [ addNum meth_n arg_n |
-                             meth_n <- meth_mult, arg_n  <- [1 .. length ins]]
-      addNum fs n =
-          concatFString [fs, fsUnderscore, (mkNumFString (toInteger n))]
-
+      -- For method "x", names are "x_<argN>" (plus a "_PORT_<portM>" suffix
+      -- when an argument is split across several ports) — one entry per
+      -- hardware port, preserving the source-arg grouping.  For methods
+      -- with multiplicity > 1, the meth_n in meth_mult already has the
```
quark17, line 581: Note that there's a bug in multiplity names that PR #928 fixes.  I'd like to get that fix in.  Should we do it after this, or do you want to roll them in together?  (PR 928 has an example source file in the comments, which needs to be set up as a test in the testsuite.)

src/comp/AVerilogUtil.hs
```diff
@@ -650,6 +713,11 @@ vExpr vco (APrim _ _ PrimConcat es@(e:_)) | allSame es = VERepeat (VEConst (gene
 vExpr vco (APrim _ _ PrimConcat es) = VEConcat (map (vExpr vco) es)
 -- Avoid repetition syntax for strings
 vExpr vco (APrim _ _ PrimStringConcat es) = VEConcat (map (vExpr vco) es)
+-- A constant bit-extract composes with an inner constant select (e.g. reading
+-- a field out of a combined split port that is itself a slice), so route it
+-- through vSelectBits to avoid an illegal chained index like x[a:b][c:d].
```
quark17: There are places where BSC insists that certain expressions must be lifted to their own definition.  If we had added TupleSel to the set of things that must be lifted, then maybe this situation (of chained indexing, here and below for signext) wouldn't have happened, because the outer index would be applied to a def.  I don't know which is better -- just mentioning the possibility.

src/comp/bluetcl.hs
```diff
@@ -3299,28 +3299,31 @@ rawIfcFieldName (RawInout i _ _ _ _) = i
...
-    (AIDef i args _ _ def
-     (Method _ clk rst mult ins mo@(Just out) Nothing) _) =
-    let -- include the type in the "mo"
-        mo' = Just (out, adef_type def)
-    in  RawMethod i mult clk rst (mapFst Just args) ins mo' Nothing
-rawIfcFieldFromAIFace pps
-    (AIAction args _ _ i _
-     (Method _ clk rst mult ins Nothing me@(Just _))) =
+rawIfcFieldFromAIFace _ iface@(AIDef i _ _ _ def
+     (Method _ clk rst mult ins outs Nothing) _) =
```
quark17: This line is the rest of the `iface` that you moved up a line.  Before, this was clear from the indentation, but now that `iface` is moved up, the rest is indented far less, which loses the connection for me.  Is there a reason you preferred it this way? (here and in the next few arms)

src/comp/ForeignFunctions.hs
```diff
@@ -153,7 +153,9 @@ data ForeignCall = FFCall   { fc_func :: ForeignFunction
 
 fc_args :: ForeignCall -> [AExpr]
 fc_args (FFCall _ expr)  = ae_args expr
-fc_args (FFAction _ act) = tailOrErr "action has no condition" (aact_args act)
+-- aact_args includes the cond as the first element; strip it so the
+-- argument list matches the foreign function's parameter list.
+fc_args (FFAction _ act) = drop 1 (aact_args act)
```
quark17: I prefer the previous definition, which raises an error if the list is empty.  Was there some reason you decided to rewrite this?

src/comp/IExpand.hs
```diff
@@ -2621,12 +2668,20 @@ walkNF e =
...
     case e of
-        ref@(IRefT _ _ r) -> do
+        ref@(IRefT _ ptr r) -> do
             hc <- getHeap r
             case hc of
-                HUnev { hc_hexpr = e } -> do
-                    e' <- unheapAll e
-                    internalError ("walkNF unev: " ++ ppReadable (e, e'))
+                HUnev { hc_hexpr = hu_e } -> do
+                    -- Force the unevaluated cell to WHNF and retry.  Without
```
quark17: Whoa.  Can you explain why this is needed?  Why is this not an indication that we've failed to evaluate something earlier?

Or ... is this reached when the source code has forgotten to include a wrapmethod?  Should we not just error in that case?

src/comp/IExpand.hs
```diff
@@ -3151,10 +3206,38 @@ conAp' i (ICPrim _ PrimIsRawUndefined) _ (T t : E e : as) = do
...
+      let (argTys, resTy) = itGetArrows ft
+          -- pair each (non-zero-width) port name with its size, flattening a
+          -- bitified tuple type into the bit-sizes of its ports
+          mkPorts names ty = zip names (filter (/= 0) (bitTupleSizes ty))
+          -- inputs are grouped per argument (kept as a 2-d list)
+          ips = zipWith mkPorts inNames argTys
+          ops = mkPorts outNames resTy
+      in  return $ P p $ ICon fi (fc { foports = Just (ips, ops) })
+    -- not a foreign-function value (shouldn't happen): leave unchanged
+    _ -> return (P p fe')
```
quark17: If this shouldn't happen, should we internal-error?

src/comp/LambdaCalc.hs
```diff
@@ -1003,8 +1009,8 @@ convStmt modId avmap (AStmtAction cset (ACall obj meth as)) = do
 
   -- convert the condition
   c_expr <- convAExpr c
-  -- convert the arguments
```
quark17, line 1007: Was there a reason to lose this comment?

src/comp/LambdaCalcUtil.hs
```diff
@@ -566,9 +568,11 @@ instance PPrint AStmt where
   pPrint d p (AStmtDef def) =
       pparen (p > 0) (text "AStmtDef" <+> pPrint d 1 def)
   pPrint d p (AStmtAction cset act) =
+      -- prepend the recomputed condition to the action's argument list
       let c = getCondExpr cset
-          as = aact_args act
-          act' = act { aact_args = (c:as) }
+          act' = case aact_args act of
+                   (_:as) -> act { aact_args = c : as }
```
quark17: What's going on here.  Before, the condition was always inserted.  Now you're saying that the list already has a condition and you're replacing it?  Did something change?

src/comp/SAL.hs
```diff
@@ -1133,8 +1141,8 @@ convStmt avmap (AStmtAction cset (ACall obj meth as)) = do
 
   -- convert the condition
   c_expr <- convAExpr c
-  -- convert the arguments
```
quark17, line 1140: Was there a reason to remove this comment?

src/Libraries/Base1/Prelude.bs
```diff
@@ -4608,18 +4622,25 @@ data (MetaField :: $ -> # -> *) name idx = MetaField
...
--- Should eventually include the output port names, when we support multiple output ports.
-primitive primMethod :: List String -> a -> a
+-- Tag a method with metadata: the input port names for each method argument,
+-- and the output port names.
+primitive primMethod :: List (List String) -> List String -> a -> a
+
+-- Like primMethod, but for a noinline function: records the (possibly split)
+-- input and output port names directly on the foreign-function value, so that
+-- a call to it connects to the same port names that the generated module
+-- exposes.  Unlike primMethod, it does not wrap the result value (which would
```
quark17, line 4632: Can you explain this?  In what way does primMethod "wrap" the result, and why is primNoInline different?

testsuite/bsc.verilog/splitports/splitports.exp
```diff
@@ -50,6 +92,88 @@ if { $vtest == 1 } {
...
+# Result is a 3-element tuple (ShallowSplit newtype): the shallow split peels
+# off the first field (fst) and leaves the rest in one combined port (snd).
+# Reading the individual fields out of the combined snd port slices a value
+# that is itself a slice, which must compose into a single index rather than
+# an illegal chained index like wire[8:0][8:1].
+test_c_veri_bs_modules NoinlineSplitTuple3 {module_triple}
+if { $vtest == 1 } {
+    find_regexp module_triple.v {output \[7 : 0\] triple_fst;}
+    find_regexp module_triple.v {output \[8 : 0\] triple_snd;}
+    find_regexp sysNoinlineSplitTuple3.v {triple__f1\[8:1\]}
```
quark17: I'll have to compile this PR and run the example, to see what the Verilog looks like -- because I'm confused where `_f1` came from.  Why does the parent not have, say, `triple_snd[8:1]` or `triple$triple_PORT_2[8:1]`?

quark17 (commented):
I've had a look over the whole PR.  Overall, seems reasonable -- but a few questions and comments below.

quark17:
Oh, I forgot to include a couple other comments:

There were places where the types became lists of lists (or lists of lists of pairs for port name and size).  In theory, we could define type aliases to make those more readable -- like a type for the tuple, called `Port` say, that's shared with output ports; or define `MethodInput` as `[Port]` and then you could write `[MethodInput]` instead of `[[(Id,Type)]]`, say.  What you have is fine -- but thought I'd mention it.

In a few places (particularly ASyntax), you reformat code that's not otherwise changing, and specifically you remove the spaces in `(c : es)` to be `(c:es)`.  I'm not sure this is always more readable -- particularly in one place (another file other than ASyntax but I forget which) where the names on either side have a final tick, it becomes a jumble.  No need to revert these, but wanted to mention it.

src/comp/BackendNamingConventions.hs
krame505: I can just add it in to this PR.

src/Libraries/Base1/Prelude.bs
krame505: * `primMethod` lives in the `WrapField` instance's `toWrapField` — the to-wrapper (definition) direction. When GenWrap builds a synthesized module's wrapper interface, each method goes through `primMethod ∘ toWrapMethod`, tagging it with the input/output port names that module will expose.
* `primNoInline` lives in `fromWrapNoInline` — the from-wrapper (call-site) direction, and only for noinline functions. At a use of a noinline function, its wrapper value is tagged via primNoInline (then fromWrapMethod turns it back into the callable function) so the call connects to the same ports the generated noinline module exposes.

src/comp/AState.hs
krame505: Done.

src/comp/ASyntax.hs
krame505: Whoops, I thought I fixed this already.  I think I messed up something in a rebase.

src/comp/ASyntax.hs
krame505: Yeah, and that also means that `AAI_MultiPort` isn't needed.  I don't remember why I changed this to `AAbstractInput` in the first place.

testsuite/bsc.verilog/splitports/splitports.exp
krame505: This is an internal signal whose name seems to have come from the pattern match expression.  
```verilog

module sysNoinlineSplitTuple3(CLK,
			      RST_N);
  input  CLK;
  input  RST_N;

  // register r
  reg [7 : 0] r;
  wire [7 : 0] r$D_IN;
  wire r$EN;

  // remaining internal signals
  wire [16 : 0] triple__f1;

  // register r
  assign r$D_IN = r + 8'd1 ;
  assign r$EN = 1'd1 ;

  // remaining internal signals
  module_triple instance_triple_0(.triple_a(8'd7),
				  .triple_b(8'd3),
				  .triple_fst(triple__f1[16:9]),
				  .triple_snd(triple__f1[8:0]));
...
```

src/comp/bluetcl.hs
krame505: I thought it looked weird the way it was, but it doesn't matter to me. Changed it back.

src/comp/IExpand.hs
krame505: Yeah good catch.

src/comp/ForeignFunctions.hs
krame505: No reason, another casualty of a poorly-reverted abandoned refactor I think.

src/comp/IExpand.hs
krame505: Right.  I guess I previously encountered this same issue on the output side, this needs `primDeepSeqTupleBits` in the `WrapMethod` instance.

src/comp/AVerilogUtil.hs
krame505: Good point.  I guess I don't see a downside to just doing the lifting.  Slightly more verbosity sometimes, but probably makes the result easier to follow.

krame505 (commented):
Okay I've addressed all of Julie's comments.  This is finished from my end, unless Ravi has anything else.

krame505:
> There were places where the types became lists of lists (or lists of lists of pairs for port name and size). In theory, we could define type aliases to make those more readable -- like a type for the tuple, called Port say, that's shared with output ports; or define MethodInput as [Port] and then you could write [MethodInput] instead of [[(Id,Type)]], say. What you have is fine -- but thought I'd mention it.

Missed this comment - yeah that's reasonable, will do.

> In a few places (particularly ASyntax), you reformat code that's not otherwise changing, and specifically you remove the spaces in (c : es) to be (c:es). I'm not sure this is always more readable -- particularly in one place (another file other than ASyntax but I forget which) where the names on either side have a final tick, it becomes a jumble. No need to revert these, but wanted to mention it.

I reverted all the pure formatting/naming changes - these were left over from me trying one approach and then partially changing it back.

src/comp/Parser/BSV/CVParser.lhs
```diff
@@ -1507,16 +1509,19 @@ returns a single identifier formed by joining the components with underscores.
 >                            Just (VeriPt p) ->
 >                              [(mkRdyId name,
 >                                V.Method (mkRdyId fullname)
->                                    clk rst 0 [] (Just p) Nothing,
+>                                    clk rst 0 [] [p] Nothing,
 >                                False)]
 >                            Just _ -> internalError "pMethodVeriProt(4)"
+>        -- BVI lists each method arg port individually; TCheck regroups
```
nanavati: I'd like to understand what is happening here in more detail. I would have expected the VModInfo to already know the argument to port structure and for that to just be threaded back through GenWrap. Then I would have expected `import "BVI"` to always just emit a one argument / one port version of things unless we extended it. This comment makes it sound like we lost information we have to reconstruct, which is something we want to avoid as much as possible.

src/comp/Parser/Classic/CParser.hs
```diff
@@ -148,7 +148,8 @@ pModule = l L_module `into` \ pos ->
                 ||! literal (mkFString "const") .> VPconst
                 ||! literal (mkFString "unused") .> VPunused
                 ||! literal (mkFString "inhigh") .> VPinhigh
-        mkMethod i n vps mo me = Method i Nothing Nothing n vps Nothing Nothing
+        -- TCheck regroups per source-language argument once the type is known
```
nanavati: Same concern about reconstruction here.

src/comp/ACheck.hs
```diff
@@ -111,10 +111,18 @@ chkAIface aa@(AIInout { aif_inout = r }) =
 chkCond :: AType -> Bool
 chkCond = isBit1
 
+-- A method argument must either be a Bit-typed expression (a single
```
nanavati, line 114: This is not exactly the structure  I expected. I was expecting that we would permit method arguments to be either Bit n or tuples of valid method arguments, so that we could have argument structure that matched the original type structure. I was hoping to avoid tuple-flattening altogether (since it is inherently O(n^2) in the typechecker). Reducing n by a constant factor is still valuable, but I was hoping for more.

Did you try that and there was a reason it didn't work out or is it something you didn't try?

src/comp/ACleanup.hs
```diff
@@ -142,17 +142,22 @@ cleanupActions flags pred as =
...
-                                do
-                                    argid <- newName
-                                    let argtyp = (aType arg)
-                                    addDef (ADef argid argtyp
-                                        (APrim argid argtyp PrimIf [cond, arg, arg']) [])
-                                    return (ASDef argtyp argid))
-                                         (zip args args'))
-                        let newcall = (ACall id methodid
-                                ((ASDef aTBool newid):newargs))
+                        -- For SplitPorts args (ATuple), merge per element so
```
nanavati, line 145: I think this does handle nested tuple structure?

src/comp/APaths.hs
nanavati, line 1: My general question about APaths is do you have enough tests to make sure that paths inside the argument structure are precise - so, for example, if a particular element of a `SplitPorts` tuple has a combinational path to a particular method output (among many) is that path captured precisely with no extraneous dependencies?

src/comp/ASyntax.hs
```diff
@@ -527,6 +538,11 @@ getArraySize t = internalError ("getArraySize: " ++ ppReadable t)
 -- then to be AAbstractInput.)
 type AInput = (AId, AType)
 
+-- One source-language method argument, decomposed into the hardware input
+-- ports it occupies (a single port for an unsplit argument, several for a
+-- split struct/tuple).  A method's arguments are then a list of these groups.
+type AMethodInput = [AInput]
```
nanavati, line 544: This makes me wonder what would/could happen if we allowed "methods" to take clocks and/or resets as arguments. Could we turn that into a way of representing input clocks within interfaces rather than forcing them to be module arguments? Not today's problem, but I thought I didn't want to forget to capture.

src/comp/ASyntaxUtil.hs
```diff
@@ -242,6 +253,28 @@ instance ATypeC ADef where
     aType d = adef_type d
     aSize d = aSize $ adef_type d
 
+-- The bit range [hi:lo] occupied by element `idx` (1-based) of a tuple type,
```
nanavati, line 256: In what context does bit extraction matter when handling tuple inputs or outputs? I would have expected expanded to have pre-calculated all of the concats and extracts needed at inputs and outputs. What scenario am I missing?

src/comp/AVerilogUtil.hs
```diff
@@ -31,8 +31,8 @@ module AVerilogUtil (
                      VConvtOpts(..)
                     ) where
 
-import Data.List(nub, partition, genericLength, union, intersect, (\\),
-                 uncons)
+import Data.List(nub, partition, genericLength, genericIndex, union, intersect,
+                 (\\), uncons)
```
nanavati, line 35: Usually using a List operation like `\\` leads to a buried O(n^2) problem somewhere down the line. Is that accurate here too?

src/comp/Parser/BSV/CVParser.lhs
quark17: Yeah, my assumption was that we don't yet support multiple ports in import-BVI.  If we do, there should be a test case for it, and I don't recall seeing one?

nanavati (dismissed):
I'd like to explore the possibility of not changing the type structure at all - just carrying it through directly without flattening any tuples. This nested structure might make it harder to trace individual ports down to the subfields they eventually go to and come from, but maybe there are clarity or structure gains that make everything 

I can see this running into trouble with coherence if we're not careful. but hopefully we can avoid it the way we avoid in in some of the tuple-manipulation typeclasses.

That being said, while this is an angle I want to pursue and ideally we'd pursue it sooner rather than later because of the implications for port freezes, I don't think that possibility has to get in the way of merging this change.

src/comp/APaths.hs
krame505: Yes, there are a number of tests for that added in the previous PR (#849).

src/comp/ASyntaxUtil.hs
krame505: Tuples are implemented as flat bit vectors in Verilog/C++ for now, this is used in the transformation to `VExpr`/`CCExpr`.

src/comp/Parser/BSV/CVParser.lhs
krame505: That was a weridly-phrased comment.  We don't support BVI with multiple ports for an input currently, but `VModInfo.Method` expects a list of ports for each arg.

src/comp/Parser/Classic/CParser.hs
krame505: Yeah, I'm not sure what that comment was about.  There is just one port per arg for now.

src/comp/ACheck.hs
krame505: You could probably make something like that work.  I feel like it probably shouldn't be using tuples any more, though, due to the asymmetry of not having a "nil" element making the nesting a bit weird.  Correlating a flat list of port names with the nested port structure gets more annoying - checking that you e.g. have the right number of port names has more overhead.  Or maybe the port names also shouldn't be a flat list to match the same structure? 

This is maybe a good future direction to consider but needs more careful thought to avoid adding overhead elsewhere.

src/comp/ASyntax.hs
krame505: Hmm, that seems plausible.  

What does it mean for a method with ready/enable signals to take clock/reset arguments?  Is there anything to consider there about validity?

src/comp/AVerilogUtil.hs
krame505: This is only used in `getAVDependDefs`, which I don't think is a hot path.

src/comp/ASyntaxUtil.hs
quark17: I'm also confused. Under what situation is a tuple still existing in the Verilog, that it requires representation as a concat?

If we have an Action method with a split input argument, where each element is written to its own register, will the Verilog connect each register to an input port directly, or will we see a wire assigned to the concat and then bit extraction from that at each register? (Similarly for an ActionValue method returning a tuple constructed from register.)

krame505: Yeah, after further consideration, the only places that ASyntax tuples show up they are immediately selected from or bound to an `ADef`.  I updated the translation to just generate separate Verilog signals for the tuple elements instead of concatenating them and bit-extracting the result.  

This does make me wonder if `ATuple`s should really even exist, after all, if they can only exist at the immediate bounteries of method invocations.  Maybe an approach more along the lines of the one I abandoned in https://github.com/krame505/bsc/tree/input-ports-tuples would be more appropriate - this gives a cleaner path to having Bluesim produce VCDs with the appropriate split ports, for instance.  

But this will take more investigation, and the `ATuple` approach is good enough for now.

krame505:
I've addressed all the comments.  I think that there is maybe more future work to do in reconsidering if tuples in ASyntax are even the right approach, or if port-splitting info should be baked into `AMethCall`/`AMethValue`/`ANoInlineFunCall` more directly.  But I think that this approach is acceptable for now, and I probably won't have bandwidth for more investigation of this until August.

krame505:
I think the failing check is a fluke because the other platforms seem to build fine.

quark17 (approved):

PR #962, opened 2026-05-31
Fix post-inference IAp assert for ATF-bearing class instances

krame505:
ISyntaxCheck.eqTypeNum's satisfy call couldn't prove equality between a fresh metavariable and an ATF application whose parent class dict had been decomposed by an Add-style instance (e.g. Bits (a, b) sab <= (Bits a sa, Bits b sb, Add sa sb sab) loses the SizeOf (a, b) ≡ sab link).  The fundep-derived sub-goal NumEq sa sa' couldn't be discharged because the env's EC --- where sa ≡ sa' is recorded --- was not visible to satisfy.

Fix: inject the env's numeric-kind EC equivalences as explicit NumEq preds before calling satisfy.  A spanning-tree's worth of pairs per class (n-1, not n*(n-1)/2) suffices since satisfy chases equalities transitively.

Also fixes a latent kind bug in iTAdd/iTMax/iTMin/iTMul/iTDiv: IKFun is left-associative as a backtick operator, so the unparenthesized ``IKNum `IKFun` IKNum `IKFun` IKNum`` form represented `(KNum -> KNum) -> KNum` rather than `KNum -> KNum -> KNum`.  The new eqTypeNum code calls kCheck on EC entries, which is the first path to actually query these kinds.

### DIFF
--- a/src/comp/ISyntaxCheck.hs
+++ b/src/comp/ISyntaxCheck.hs
@@ -118,7 +118,8 @@ eqTypeNum flags symt r@(E _ _ ecRaw _) t1 t2
           satisfy (eqs ++ ecEqs) [vp]
     in  case runTI flags False symt satisfyEq of
           (Right ([],_), _, _) -> True
-          _ -> False
+          res -> --trace("eqTypeNum: not satisfied: " ++ ppReadable (t1, t2, res)) $
+                 False
  where isITAp (ITAp _ _) = True
        isITAp _          = False
 eqTypeNum _ _ _ _ _ = False
--- a/src/comp/ISyntaxCheck.hs
+++ b/src/comp/ISyntaxCheck.hs
@@ -3,6 +3,7 @@ module ISyntaxCheck(iGetKind,
                     tCheckIPackage,
                     tCheckIModule) where
 
+import Data.List (mapAccumL)
 import qualified Data.Map as M
 import qualified Data.Set as S
 import qualified EquivalenceClass as EC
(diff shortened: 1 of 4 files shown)


quark17:
Does this conflict with #955 ?  That changes the `satisfy` in `ISyntaxCheck.eqTypeNum` (to be `satisfyFV boundTVs`), where this PR is also changing things.

nanavati:
I think it should be fine. This PR is just adding missing equality constraints and that one is being more accurate about bound type variables. Both changes improve the accuracy of typechecking so they should be compatible.

src/comp/ISyntaxCheck.hs
```diff
@@ -76,27 +77,48 @@ eqType1 _ _ _ (ITNum n) (ITNum n') = n == n'
...
           addBoundTVs (M.elems m)
-          addExplPreds eqs
+          addExplPreds (eqs ++ ecEqs)
           vp <- mkVPredFromPred [] (IsIn numEqCls [t1', t2'])
-          satisfy eqs [vp]
+          satisfy (eqs ++ ecEqs) [vp]
     in  case runTI flags False symt satisfyEq of
           (Right ([],_), _, _) -> True
-          res -> --trace("eqTypeNum: not satisfied: " ++ ppReadable (t1, t2, res)) $
-                 False
```
quark17, line 99: Did you mean to remove this commented-out debug statement?

krame505: Oops, reverted.

quark17 (approved):

Issue #964, opened 2026-06-04
Deriving via support


krame505:
It would be nice if we supported something like `deriving via` in Haskell: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/deriving_via.html

This would avoid a lot of boilerplate instance for port splitting, for instance
```
instance (ShallowSplitPorts (Foo s l) p) =>
    SplitPorts (Foo s l) p where
  splitPorts = shallowSplitPorts
  unsplitPorts = shallowUnsplitPorts
  portNames = shallowSplitPortNames
```
could become just a `derive SplitPorts via ShallowSplit` clause on the declaration of `Foo`.  That would generate the instance
```
instance (SplitPorts (ShallowSplit (Foo s l)) p) =>
    SplitPorts (Foo s l) p where
  splitPorts x = splitPorts (ShallowSplit x)
  unsplitPorts x =
    let ShallowSplit y = unsplitPorts x
    in y
  portNames x y = unsplitPorts (ShallowSplit x) y
```

PR #965, opened 2026-06-06
Deriving Via

jkopanski:
My take on #964 during ZuriHac.

For now it has a commit with my dev env setup that I'll get rid of later.


src/comp/CSyntax.hs
```diff
@@ -255,6 +255,16 @@ instance NFData IdK where
     rnf (IdKind i k) = rnf2 i k
     rnf (IdPKind i pk) = rnf2 i pk
 
+data CDeriving
+        = CStock [CTypeclass]
+        -- TODO: expr?
```
krame505: Not sure what this TODO is about?

testsuite/bsc.typechecker/deriving/DerivingViaHK.bs
```diff
@@ -0,0 +1,12 @@
+package DerivingViaHK() where
```
krame505, line 1: Does HK mean higher-kinded? That kind of threw me at first because `a` is kind `*`.  But I guess `Maybe` is kind `* -> *`?  The issue here is that `a` isn't at the top-level, I guess.

krame505 (commented):
Looks good!  I'll let @quark17 approve merging.

krame505:
Actually it looks like there are some testsuite failures, nothing major: https://github.com/B-Lang-org/bsc/actions/runs/27140760096/job/80181556806?pr=965

jkopanski:
> Actually it looks like there are some testsuite failures, nothing major: https://github.com/B-Lang-org/bsc/actions/runs/27140760096/job/80181556806?pr=965

If I saw it right failing tests were: `EContextReduction1`, `bugID169` and `ETooGeneral1`. All of those used `deriving ()` which I didn't think was valid and thus I replaced `sepBy` with `sepBy1`. Now it's reverted.

src/comp/CSyntax.hs
jkopanski: This was my first change in the code, I wasn't sure if the `Id` is the proper node to hold via target. Actually I'm still not sure if that is the _correct_ choice here, but it worked out.

testsuite/bsc.typechecker/deriving/DerivingViaHK.bs
jkopanski: yeah I wasn't sure how to name it. Indeed that's about for example `Maybe`, I'm not making any effort to substitute `a` for concrete data type we are deriving instance for, when `a` is used as argument to `TAp`. I don't think I could derive that kind of function anyway.

krame505:
Thanks, looks good to me. 

@matx-chris, can we include this in our next internal BSC release?

src/comp/CSyntax.hs
krame505: I guess in theory we could have a partially-applied type constructor instead of the name of a type here? Does Haskell support that?

jkopanski: I don't think so. Haskell expects a type here. Given that DerivingVia works by the way of `coerce`, it can be arbitrarily complex type, see example from ghc user guide:
```haskell
newtype Kleisli m a b = Kleisli (a -> m b)
  deriving (Semigroup, Monoid)
    via (a -> App m b)
```
In here (bsc) I've assumed, that the type and constructor names must be the same. Perhaps this could only be constructor name, if I were not to generate type signature for the typeclass methods. But I've had the hunch the errors would be slightly better if I were to do it.

src/comp/Parser/Classic/CParser.hs
```diff
@@ -531,9 +531,19 @@ pSummand' = pSummandConIds `into` \ constr_names ->
...
+pDeriving :: CParser CDeriving
+pDeriving = l L_deriving ..+
+          ( pDerivingStock
+        ||! pDerivingVia)
+
+pDerivingStock :: CParser CDeriving
+pDerivingStock = lp ..+ sepBy pTypeclass cm +.. rp         >>- CStock
+
+pDerivingVia :: CParser CDeriving
+pDerivingVia = pTypeclass +.+ l L_via ..+ pTyConId          >>> CVia
```
matx-amy, line 543: is there any chance this could be generalized to allow a parenthesized type? `deriving Foo via (Bar Baz)` is useful

jkopanski: I don't think so, at least not using the syntax proposed in the #964.

Right now the generated code uses the via identifier in two places, the class method type signature and data constructor. I guess the type signature isn't really needed here, I've added it because I've had a hunch it could lead to better error messages when something is not right.

What is necessary is data constructor that is used in the patterns in the generated code, and `Bar Baz` is a type not a constructor.

This all comes from the fact that it works quite different from the Haskell `DerivingVia`. Haskell variant just uses `coerce`. Here we generate as described in the issue that relies on the convention (single constructor datatype, with constructor name same as the type name), since there are no newtypes available (right?).

src/comp/CFreeVars.hs
```diff
@@ -560,10 +560,10 @@ getFTCDn (Cforeign { cforg_type = t } ) = getFQTyCons t
 getFTCDn (Ctype _ _ t) = getFTyCons t
 getFTCDn (Cdata { cd_internal_summands = summands,
                   cd_derivings = derivings }) =
-    S.unions (map f summands) `S.union` S.fromList (map typeclassId derivings)
+    S.unions (map f summands) `S.union` mconcat (map getFTDer derivings)
```
quark17: Here and a few lines below, I would probably use `S.unions` instead of `mconcat`

src/comp/Parser/BSV/CVParser.lhs
```diff
@@ -869,15 +869,28 @@ though can it ever be anything other than KStar?)
...
-> pDerivationsIf :: Bool -> SV_Parser [CTypeclass]
+> pDerivationsIf :: Bool -> SV_Parser [CDeriving]
 > pDerivationsIf False = option [] $
 >     do pos <- getPos
 >        pKeyword SV_KW_deriving
 >        failWithErr (pos, EForbiddenDeriving)
-> pDerivationsIf True = option [] $
->     do pKeyword SV_KW_deriving
->        pInParens (pCommaSep1 (pTypeclass <?> "typeclass name"))
+> pDerivationsIf True = many pDeriving
```
quark17: You've changed BH syntax to allow multiple `deriving` clauses.  While you haven't enabled deriving-via for BSV, here you have enabled multiple deriving clauses, which wasn't the syntax before.  If we're not sure about the BSV syntax yet, then maybe we shouldn't change this.

src/comp/CSyntax.hs
quark17: I would either remove the TODO comment or make it more clear.

src/comp/CVPrint.hs
```diff
@@ -438,9 +438,8 @@ pBlockNT _ n nl xs sep =
         (t (replicate n ' ') <>
         foldr1 ($+$) (map (\ x -> x <> if nl then sep $+$ empty else sep) xs))
 
-ppDer :: PDetail -> [CTypeclass] -> Doc
-ppDer d [] = empty
-ppDer d is = text "deriving (" <> sepList (map (pvPrint d 0) is) (text ",") <> text ")"
+ppDer :: PDetail -> [CDeriving] -> Doc
+ppDer d drvs = vcatList (map (pPrint d 0) drvs) empty
```
quark17: I'm not sure we can use `pPrint` this way.  The identifier in "deriving" can be qualified, and BSV and BH print qualifiers differently.  I think we have to duplicate what `pPrint` of `CDeriving` is doing, except use `pvPrint` inside for IDs.

src/comp/Deriving.hs
```diff
@@ -241,6 +245,59 @@ doStructDer _ _ _ i vs cs (CTypeclass di) | isTCId i =
...
+          body = case isCTyVar ret of
+            Just v | v == dv -> Ccase (getPosition di) call [CCaseArm (CPCon tgt [CPVar id_y]) [] (cVar id_y)]
+            _ -> call
+        in CDef fun_name (CQType preds $ uncurry cTApplys tys) [CClause (map (CPVar . snd) vars) [] body]
+      mkFun :: CField -> CDefl
+      mkFun f = CLValueSign (mkCDef f) []
+    in Right $ pure $ Cinstance (CQType ctx appliedClass) (map mkFun fs)
+
+doVia _ _ _ (di, tgt) =
+  Left (getPosition di, ECannotDerive (pfpString di ++ " via " ++ pfpString tgt))
```
quark17, line 334: Do we want to add a test case for this error situation?

src/comp/Deriving.hs
```diff
@@ -241,6 +245,59 @@ doStructDer _ _ _ i vs cs (CTypeclass di) | isTCId i =
 doStructDer _ _ _ i vs cs (CTypeclass di) =
   Left (getPosition di, ECannotDerive (pfpString di))
 
+-- -------------------------
+
+-- | Derive an instance of a typeclass using Deriving via mechanism
+-- for a given data (sum type), and return the instance definitions.
```
quark17: It says "data (sum type)", but `doVia` is called for both `data` and `struct` types.

testsuite/bsc.typechecker/deriving/DerivingViaHK.bs
```diff
@@ -0,0 +1,12 @@
...
+class Weird a where
+  weird :: Maybe a -> Bool
+
+data I a = I a
+
+instance Weird (I a) where
+  weird = const True
+
+data I' a = I' a
+  deriving Weird via I
```
quark17: This file needs a final newline

testsuite/bsc.typechecker/deriving/deriving.exp
```diff
@@ -23,13 +23,20 @@ compare_file NoDerive2.bs.bsc-out
...
 # --------------------
 # more deriving tests (using verilog only for $display and $finish) 
 
 test_veri_only DerivingMod
 
+# --------------------
+# tests for deriving via
+
+compile_pass DerivingVia.bs
+compile_fail_error DerivingViaHK.bs T0030
```
quark17, line 47: This needs a comment explaining what's being tested.  Is this a bug in BSC, that it gives an error or at least gives an unhelpful error?  Or is this the message (and position) that you expect?  (If it's a bug, we'll want to file it and record the number here.)

testsuite/bsc.typechecker/deriving/DerivingVia.bs
```diff
@@ -0,0 +1,13 @@
+package DerivingVia() where
+
+import SplitPorts
+
```
quark17, line 4: I would also add uses of the derived instances, to prove that they exist and were defined properly.  I don't know if that's easy to do with `ShallowSplit` or if we should use a simpler typeclass for this example.

testsuite/bsc.typechecker/deriving/DerivingVia.bs
```diff
@@ -0,0 +1,13 @@
...
+data Foo = Foo (Int 8)
```
quark17, line 5: You should also test on a `data` type with multiple disjuncts.

src/comp/SystemVerilogKeywords.lhs
```diff
@@ -604,6 +605,7 @@ scanner and to prettyprint keywords.
 >      (SV_KW_actionvalue,         "actionvalue",          Bluespec38),
 >      (SV_KW_endactionvalue,      "endactionvalue",       Bluespec38),
 >      (SV_KW_deriving,            "deriving",             Bluespec38),
+>      (SV_KW_deriving,            "deriving",             Bluespec38),
```
quark17: This duplicate line should be removed.

src/comp/SystemVerilogKeywords.lhs
```diff
@@ -256,6 +256,7 @@ Data type declaration for the keywords:
 >     | SV_KW_actionvalue
 >     | SV_KW_endactionvalue
 >     | SV_KW_deriving
+>     | SV_KW_via
```
quark17: I'd probably comment this out, if it's not being used yet.

quark17 (commented):
Thanks for your interest in developing BSC and contributing this!

I have not closely looked at `doVia` yet, but have comments on everything else.  I think we need to decide whether this constructor approach is what we want to implement, or whether we should do something like use `coerce` (should be an easy simplification to `doVia`) or take an arbitrary expression for what to call (would require parsing an expression, also possibly not difficult?).

nanavati:
Adversarial review of a tree carrying this PR found one wrong-results-class gap, confirmed by reproduction: `doVia` generates the unwrap as a single-arm case (`Ccase ... [CCaseArm (CPCon tgt [CPVar y]) ...]`) without validating the target's shape. With a multi-constructor target:

```haskell
data W a = W a | X a
instance (MyC a) => MyC (W a) where def = X def
data T = T (Int 8)
  deriving MyC via W
```

this compiles with zero warnings and the derived method elaborates to an unspecified value in the generated Verilog (`assign r$D_IN = 8'b10101010 /* unspecified value */` in our repro) — the partial case falls through silently. GHC's DerivingVia enforces representational equality; the analogue here is requiring the via target to be a single-constructor, single-field wrapper whose constructor is named like the type (which the generated `CCon tgt`/`CPCon tgt` already assume).

Validated fix (guard clause before the main `doVia` clause):

```haskell
viaTargetShapeOK :: [(Id, CDefn)] -> Id -> Bool
viaTargetShapeOK xs tgt =
    case lookup (unQualId tgt) xs of
      Just (Cdata { cd_original_summands = [summand] }) ->
          length (cos_arg_types summand) == 1 &&
          any (qualEq tgt) (cos_names summand)
      _ -> False
```

rejecting with a positioned `ECannotDerive` ("the via target must be a data type whose single constructor is named ... and takes a single argument"). With that in place the existing `ShallowSplit` uses still derive fine, and the multi-constructor case gets a clean compile-time error. Happy to push the change (plus a regression test asserting the error) to this branch if you'd like.

nanavati:
Follow-up to the above: the fix is now available as a ready-made commit on top of this branch — nanavati/bsc@bd91fcf1 (branch [`965-via-target-shape`](https://github.com/nanavati/bsc/tree/965-via-target-shape)), including the regression test and expected output.

One refinement over the snippet posted earlier: when the via target is written *qualified* (`via Q.W`), the shape check matches the candidate definition's own qualifier rather than taking the first unqualified hit — otherwise a local type shadowing the imported one the user named would be judged instead, which could let a multi-constructor target through. Unqualified targets keep normal locals-first scoping, which is also what the generated `CCon` reference resolves to.

The identical change is validated in our integration tree (multi-constructor target → positioned error; the existing `ShallowSplit` derivations unaffected). @jkopanski feel free to cherry-pick it into this PR, or say the word and I'll push it here directly.

src/comp/CFreeVars.hs
jkopanski: While doing so I've noticed that it might clearer (for me at least) to write it as: `S.unions $ map f summands ++ map getFTDer derivings`

But I've noticed that below there are many cases where there are unions of unions taken instead. So I've just did the simple change.

testsuite/bsc.typechecker/deriving/DerivingVia.bs
jkopanski: done in @nanavati commit

jkopanski:
> @jkopanski feel free to cherry-pick it into this PR

I did, thanks

testsuite/bsc.typechecker/deriving/DerivingViaMultiCon.bs.bsc-out.expected
```diff
@@ -0,0 +1,3 @@
+Error: "DerivingViaMultiCon.bs", line 17, column 11: (T0000)
```
jkopanski, line 3: @nanavati I'm getting a following diff here:

```
--- DerivingViaMultiCon.bs.bsc-out.expected	2026-07-13 14:34:56.121794265 +0200
+++ DerivingViaMultiCon.bs.bsc-out	2026-07-13 21:02:35.052396080 +0200
@@ -1,3 +1,5 @@
+checking package dependencies
+compiling DerivingViaMultiCon.bs
 Error: "DerivingViaMultiCon.bs", line 17, column 11: (T0000)
   Illegal deriving: MyC via W: the via target must be a data type whose single
   constructor is named W and takes a single argument
```

looking at other examples I think those, `checking (...) compiling (...)` lines should be there, right? I'm doubting myself given that you've said that you have pulled those changes out of the internal repo, which I presume is passing tests with this exact expected?

PR #967, opened 2026-06-10
Port splitting part 4: Split tuples by default, make PrimPair no longer be an interface

krame505:
This is based on #959, only the last 2 commits are new.  

This did break a few tests that were relying on synthesizing a module with a tuple as the interface.  I changed most of them to just use interfaces, and a couple of tests that seemed to be specifically testing tuples as interfaces I marked as expected fail for now.


src/comp/AExpr2Yices.hs
```diff
@@ -587,16 +588,42 @@ convAExpr2YExpr mty (AMethCall ty@(ATBit width) modId methId args) = do
...
+    -- will appear logically equivalent
+    smap <- gets stateMap
+    let portId = getMethodOutputPorts smap modId methId `genericIndex` selIdx
         e = (AMethValue ty modId portId)
+    -- XXX This could be an unevaluated function, applied to converted arguments
+    addUnknownExpr mty e width
+-- A select from any other tuple-typed expression (e.g. a noinline result, or a
+-- def holding split ports) is opaque to the solver, like the call itself; model
+-- it as an independent variable (memoized, so identical selects compare equal).
+convAExpr2YExpr mty e@(ATupleSel (ATBit width) _ _) =
```
quark17, line 613: I'm not sure if this is right.  But it would help to see examples of when this arm is used.

testsuite/bsc.bsv_examples/AssertionsDemo/AssertionWires.bsv
```diff
@@ -46,14 +46,17 @@ endinterface
 
 typedef ModuleCollect#(AssertionWire, i) AssertModule#(type i);
 
-typedef Tuple2#(AssertionWires#(n), i) AssertIfc#(type i, type n);
+interface AssertIfc#(type i, type n);
```
quark17, line 49: For synthesized modules, we create a new interface for the boundary, which we lift back to the original type. So then, couldn't we continue to allow users to write tuples, and just translate that to something else (that's an actual interface) in the synth interface?

src/comp/AExpr2Yices.hs
krame505: Good catch, there was an off-by-one error since the indexing is 1-based here.  I added a test.

testsuite/bsc.bsv_examples/AssertionsDemo/AssertionWires.bsv
krame505: You are suggesting making tuples synthesizable as a special case?  That seems like a lot of extra special-case complexity, for what benefit?

quark17: The user writes an interface with their high-level types, and BSC's machinery determines the synth-level interface that corresponds to it and constructs the lowering and lifting functions.  For `Bool` in the user's interface, there's `Bit 1` in the synth-level interface.  I'm not asking for special machinery, any more than `Bool`-to-`Bit 1` is handled by special machinery; what I think I'm asking is: What if a tuple in the user's interface (is that `PrimPair`?) could correlate to a pair interface type (some new `PrimPairIfc`?) in the synth interface?

If that correspondence is done by typeclasses, can we declare some instances for `PrimPair` that would allow a user's interface to synthesize this way?  (Or via some small change to the typeclasses -- nothing hardcoded for tuples.)  I guess the situation I'm after is only methods with no arguments, just a return value -- which is the same as a subinterface.  Are subinterfaces handled by typeclass machinery and, if so, what part checks whether the type is an interface, and could that be loosened to allow non-interface types, as long as an instance exists for defining how to convert from the non-ifc type to an ifc type?

quark17: I'm just thinking that people might like to write BH that uses tuples, and have that correspond to tuple interfaces in places.  Although, I guess, you're just trading "not having to write something special to get a tuple ifc" for "having to write something special when you really just want a tuple value returned, not separate methods for each part".  So maybe I'm coming around to not needing this feature.

If someone wants to return a tuple of interfaces, what's the recommended way to do that -- always write a new interface type declaration?  Or should we predefine `PairIfc` and a `tupleToIfc` function that converts from nested `PrimPair` to nested `PairIfc`.  Although, there's still the problem of how you would declare the type of the module's interface -- the parentheses syntax is still only going to work for `PrimPair` and not for `PairIfc`.  Hm.

quark17: Or, perhaps like we have the type constructors specifying split, we could have `AsIfc` and `AsValue` that could be applied to types, to say whether it should be treated as a subinterface or as a no-arg method returning a value.

krame505: The actual generation of the wrapper interface structure itself, and flattening of sub-interfaces, is not being done via type classes - it is still being done by GenWrap as before.  

There is nothing stopping someone from having a (non-synthesized) function of type `Module (Foo, Bar)` that returns a tuple of interfaces.  But it does not seem to me to be a good design to allow synthesis boundaries to be tuples?  There is nowhere to specify pragmas, for instance.

krame505: > Although, I guess, you're just trading "not having to write something special to get a tuple ifc" for "having to write something special when you really just want a tuple value returned, not separate methods for each part". So maybe I'm coming around to not needing this feature.

I guess I maybe still don't understand what you are proposing.  Are you talking about just having a value method whose type is a tuple?  That still works.  Before it would have been split into separate methods, now it stays as a single method with multiple output ports.

nanavati (changes requested):
Can we get a branch-to-branch comparison against the previous PR so that it is clearer what still needs to be reviewed?

quark17:
> Can we get a branch-to-branch comparison against the previous PR so that it is clearer what still needs to be reviewed?

The last 4 commits are the new ones.  The following link shows a diff with part 3 (confirming that it's just the last 4 commits): https://github.com/MatX-inc/bsc/compare/input-ports-tuples-v2...krame505:split-tuple-ports

nanavati (dismissed):
Seems pretty clean. A few more tests of the new splitting behavior wouldn't be a bad idea, but otherwise it looks good.

krame505:
Rebased and updated, I think this one is also ready.  There already are a few tests that adequately cover tuple splitting that I updated tp exercise the new behavior.

krame505:
I am also now convinced that `Vector` should probably be split by default.  This would match the current interface-flattening behavior.  Additionally it is rather awkward to specify that a vector in a method input or output *should* be split, because it that would require writing an orphaned instance for `Vector`; one could instead use a struct wrapper around a Vector if they specifically don't want splitting.

krame505:
To summarize our discussion today:
* We are open to making this change but will postpone until the next release.  We agree that `PrimPair` should NOT be an interface, but whether it should always be split by default (in inputs, etc.) is an open question.
* `Vector` should not be split by default.  We should add a `SplitVector` alias for `Vector` that does split.

krame505:
I had forgotten that there is also a bug fix on this branch (commit 12ea6fdd09246d43890fe8abab42ccb1db66802d) which should go in the release, even if this PR is deferred.  I will cherry-pick the fix back to #959 on which this is based.

krame505:
Ravi's latest commits here look good to me.  It did seem weird that the `TupleSize` context on `checkPortNames` wasn't being demanded before.

src/comp/ILift.hs
```diff
@@ -33,6 +33,35 @@ iLift errh flags imod@(IModule { imod_local_defs = ds,
         ifc' = map (iLiftIfc errh flags) ifc
 --      itvs'? do we need to go into state vars?
 
+-- Merge one argument pair of two mutually exclusive calls to the same
```
quark17: I would just suggest that this belongs at the end of the file (after `lift1` which uses it, along with `genLiftCond` which is another utility function used by `lift1`).

src/Libraries/Base1/Prelude.bs
```diff
@@ -4852,14 +4855,13 @@ instance PrimSeqTupleBits () where
 -- Compute the list port names for type 'a' given a base name.
 -- Check that the number of port names matches the number of ports.
 -- This error should only occur if there is an error in a SplitPorts instance.
-checkPortNames :: (SplitPorts a p) => a -> String -> List String
+checkPortNames :: (SplitPorts a p, TupleSize p n) => a -> String -> List String
 checkPortNames proxy base =
   let pn = portNames proxy base
-      n = valueOf (TTupleSize p)
```
quark17: There's something unsatisfying about having to surface the proviso, when we know that every type will have an instance (and it's coherent and won't have orphans) and there are no class methods so the dictionary doesn't need to be plumbed for access to its fields.

krame505: I think that our notion of "coherent" needs further refinement - it is actually wrong to reduce `TupleSize p n` using the default instance that matches size-1 tuples, we need to do it "outside" the function when we have a specific `a`.  But marking `TupleSize` as incoherent breaks something else (I don't remember what exactly, off the top of my head.)

quark17:
> Ravi's latest commits here look good to me.

Sadly, they should have not been put in this PR.  I'm going to create a new PR for them, so we can get them merged -- because I accidentally merged the fix for 890 without checking that it compiled and passed the testsuite, and it does not without this fix.

src/comp/AExpr2STP.hs
```diff
@@ -603,6 +603,11 @@ convAExpr2SExpr mty (ATupleSel ty@(ATBit width) (AMethValue _ modId methId) selI
     smap <- gets stateMap
     let e = AMethValue ty modId (getMethodOutputPortAt smap modId methId selIdx)
     addUnknownExpr mty e width
+-- A select from any other tuple-typed expression (e.g. a noinline result, or a
```
quark17, line 606: Are these additions specific to tuple changes in this PR, or is this a fix that we should separately merge?

krame505: I think this was a fix to a bug that I discovered by doing the prim-pair-not-an-interface change?

krame505:
I'm confused what that earlier comment of mine was about - I don't see any commits from Ravi here now.

quark17:
> I'm confused what that earlier comment of mine was about - I don't see any commits from Ravi here now.

I moved the commits to a new PR #1039 , so that they could be merged.

I pushed a rebase of this PR that removed already-merged commits, and added a commit to fix the SplitVector test that had a tuple in it.

PR #968, opened 2026-06-10
Merge the *S size-parameter interfaces in CGetPut/BGetPut

krame505:
The CGet/CGetS (and CPut/CPutS, CClient/CClientS, CServer/CServerS, BGet/BGetS, BPut/BPutS, BClient/BClientS, BServer/BServerS) split existed only as a workaround: BSC historically could not expand the SizeOf type function inside an interface/struct field, so the bit-width was hoisted to an explicit numeric parameter on a `*S` interface, with the value-typed `CGet n a = CGetS n a (SizeOf a)` provided as a type synonym.

Type functions can now deal with this properly: expand `SizeOf a` when `a` is concrete, and defer it as a numeric equality (`sa ~ SizeOf a`, available from the modules' `Bits a sa` provisos) when `a` is abstract. So the `*S` interfaces can be dropped and the value-typed interfaces given `Bit (SizeOf a)` fields directly.

- CGetPut.bs: CGetS/CPutS/CClientS/CServerS removed; CGet/CPut/CClient/ CServer are now real interfaces (CGet/CPut abstract, as before).
- BGetPut.bs: BGetS/BPutS were size-only (`BGetS sa`); merging makes BGet/BPut value-typed (`BGet a`). Same for BClient/BServer.
- Dropped the "For compiler reasons ... Hopefully this will be fixed soon" notes and the SizeOf type synonyms.
- Updated exports, Connectable instances, all mk* module signatures, and the doc comments / libraries_ref_guide BGetPut.tex.

### DIFF
--- a/doc/libraries_ref_guide/LibDoc/BGetPut.tex
+++ b/doc/libraries_ref_guide/LibDoc/BGetPut.tex
@@ -128,11 +128,11 @@ typedef Tuple2 #(Get#(element_type), BPut#(element_type)) GetBPut #(type element
 
 The \te{BGet} and \te{BPut} interface are connectable.
 \begin{libverbatim}
-instance Connectable #(BGet#(sa), BPut#(sa));
+instance Connectable #(BGet#(element_type), BPut#(element_type));
 \end{libverbatim}
 \lineup
 \begin{libverbatim}
-instance Connectable #(BPut#(sa), BGet#(sa));
+instance Connectable #(BPut#(element_type), BGet#(element_type));
 \end{libverbatim}
 
 \item{\te{BClient} and \te{BServer}}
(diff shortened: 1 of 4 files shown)


doc/libraries_ref_guide/LibDoc/BGetPut.tex
```diff
@@ -167,8 +167,8 @@ \subsubsection{BGetPut}
 \te{mkBGetPut} & Create one end of the buffer.  Access to it is via a \te{Put} interface.\\
 \cline{2-2}
 & \begin{libverbatim}
-module mkBGetPut(Tuple2 #(BGet#(sa), Put#(a)))
-  provisos (Bits#(a, sa));
+module mkBGetPut(Tuple2 #(BGet#(element_type), Put#(element_type)))
+  provisos (Bits#(element_type));
```
quark17: Here, and below, I think you need a second argument to `Bits`

quark17: Although I notice that the CGetPut documentation has `Bits` without a second argument like this (that should eventually be corrected).

The CGetPut documentation also omits the bodies of the interfaces, just saying "abstract" -- I don't know if that was motivated by not wanting to get into the type details of the workaround, or the desire to not overload the reader with details they don't need (since in theory you just use the connectable instances etc and don't work with the individual methods).

krame505: Whoops, I asked Claude to update the documentation and didn't review it carefully enough.

quark17 (approved):

PR #978, opened 2026-06-18
RFC: Warn when using interface syntax to construct structs

matx-amy:
Right now, interface expressions get type-checked by translating them to a struct declaration and type-checking that; because of this, there's no check that when `interface` syntax is used, the type actually _is_ an interface rather than a struct. This adds a warning when a struct is used where an interface is expected.

This is an RFC -- I'm not sure if this is actually hugely important, but it seemed worth flagging, and it was a small patch.

T0160 was chosen to not conflict with #923.

@krame505 suggested it'd be nice to continue to have some indentation-sensitive syntax for constructing structs. Maybe we could use the `struct` keyword in an analogous manner to do that? So, this test case would work if the user wrote instead:

```bh
foo :: Module Foo
foo = module
  return $
    struct Foo
      bar = 0
```

### DIFF
--- a/src/comp/Error.hs
+++ b/src/comp/Error.hs
@@ -807,6 +807,7 @@ data ErrMsg =
         | ENotExpr
         | ELocalRec [String]
         | ENotAnInterface
+        | WInterfaceSyntaxUsedForStruct String
         | ENotModule String
         | EPolyField
         | ENotKNum String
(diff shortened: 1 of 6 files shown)


quark17:
I would suggest testing BSV as well.  And also test the other direction, where struct syntax is used for an interface.  I would not put new tests under `bsc.interra` -- I would try to find a place under `bsc.typechecker` or potentially under `bsc.syntax`.  I note that `bsc.typerchecker/constructors` has a test for record update syntax on interfaces.  (We had done some work to make BSC more strict about not mixing syntax for structs and unions in pattern matches, and I know some of the tests are in `bsc.typechecker/constructors` but I was trying to see whether any tests were elsewhere, like `bsc.syntax`.)

PR #979, opened 2026-06-19
More prominent HLS mention

jkopanski:
Things would be much easier for me if I found this README sooner.  I think HLS is a pretty important part of devenv and should be mentioned in the top level document.


quark17:
Thanks!  I don't use HLS, so I didn't think about this, but it's definitely worth including.

I've also added a pointer to the CI, which provides examples of setting up the HLS, as well as GHCi and installing dependencies and compiling.  If you have no objection, I'll squash merge these.  Thanks!

jkopanski:
> (...) CI, which provides examples of setting up the HLS, as well as GHCi

oh cool, I didn't think of looking into CI config for hints on setting up dev tools. Yes lets merge this

PR #980, opened 2026-06-24
typecheck: propagate substitution to sibling preds in satMany' Right branch

krame505:
## Summary

`bsc` rejects some valid instances with a spurious **T0029 "Signature mismatch (given too general)"** when the instance context contains a proviso that is only *partially* solved during context reduction (it grounds a variable but leaves a residual proviso) together with a sibling proviso that is determined, via a functional dependency, by that same variable.

## Root cause

In `satMany'` (`src/comp/TCMisc.hs`), the "still-needed" branch (where reducing predicate `p` produced residual `needed` preds) commits the substitution `s'` it learned to the accumulator (`s' @@ s`), but does **not** apply `s'` to the not-yet-processed sibling predicates `ps`. The fully-solved branch immediately below already does this (`apSub s' ps`).

Because of the asymmetry, a sibling predicate determined by a fundep from a variable that `s'` just grounded is matched against an unbound variable, fails to reduce, and is left deferred — so the instance's fundep-determined parameter stays general in the declared head while the method body grounds it. That mismatch is reported as T0029.

## Fix

Apply `s'` to `ps` in the still-needed branch too, matching the fully-solved branch. This only propagates an already-committed substitution, so it adds no new reduction rounds and cannot affect termination.

## Minimal reproduction

```
package T where
data Port t = Port
data Proxy a = Proxy
class Foo a where {}
class Det c o | c -> o where { det :: c -> o }
instance Det Bool (Port Bool) where det _ = Port
class G x c | x -> c where { g :: x -> c }
instance (Foo a) => G (Proxy a) Bool where g _ = True
class S a o | a -> o where { s :: a -> o }
instance (G (Proxy a) c, Det c o) => S (Proxy a) o where
  s p = det (g p)
```

Reducing `G (Proxy a) c` grounds `c = Bool` but leaves the residual `Foo a`; the sibling `Det c o` (fundep `c -> o`) then determines `o = Port Bool`. Before this change the grounding was not propagated to `Det`, so `o` was left general → T0029. After, it compiles. (Originally hit with associated type functions and the `SplitPorts` library, where `FilterField` grounds the common representation but leaves a residual from a stuck type-function field; the bug is independent of those features.)

## Testing

- Added regression test `testsuite/bsc.typechecker/ctxreduce/SiblingFundepSubst.bs` (fails T0029 before, compiles after; no type-function or library dependencies).
- Rebuilt the full standard library with the patched compiler — no errors.
- Compared patched vs unpatched compiler output across the test suite: **0 behavioral differences over 5438 `.bs`/`.bsv` files**.

https://claude.ai/code/session_01MJteQY8qEsYnU4tcZ2rmA4

### DIFF
--- a/src/comp/TCMisc.hs
+++ b/src/comp/TCMisc.hs
@@ -450,8 +450,20 @@ satMany' dvs es rs_accum sbs s (p:ps) = do
             -- prevented satisfying -- is this apSub just covering for an
             -- issue elsewhere (that should not have returned unsubst'd preds)?
             --
+            -- We also apply "s'" to the not-yet-processed preds "ps".  Even
+            -- though "p" could not be fully satisfied, reducing it may have
+            -- committed improvements (e.g. grounding a variable via a
+            -- functional dependency).  A sibling pred whose own reduction is
+            -- determined by such a variable must see the improvement, otherwise
+            -- it is matched against an unbound variable and left unreduced --
+            -- and its own fundep never fires.  (E.g. FilterField grounds
+            -- "r_common" but leaves a residual "Foo a c"; without this, the
+            -- sibling "DeepSplitPorts'' r_common p" never reduces, so "p" is
+            -- not determined in the instance head.)  The fully-solved ("Left")
+            -- branch below already does this ("apSub s' ps"); doing it here too
+            -- keeps the branches consistent.
             rtrace ("satMany Right: " ++ ppReadable needed) $
-            satMany' dvs es ((apSub s' needed) ++ rs_accum) (sbs' <++ sbs) (s' @@ s) ps
+            satMany' dvs es ((apSub s' needed) ++ rs_accum) (sbs' <++ sbs) (s' @@ s) (apSub s' ps)
         ([], sbs', s') ->
             -- If p is satisfied, we "drop" it, but add its binding and
             -- substitution.
--- a/src/comp/TCMisc.hs
+++ b/src/comp/TCMisc.hs
@@ -450,18 +450,8 @@ satMany' dvs es rs_accum sbs s (p:ps) = do
             -- prevented satisfying -- is this apSub just covering for an
             -- issue elsewhere (that should not have returned unsubst'd preds)?
             --
-            -- We also apply "s'" to the not-yet-processed preds "ps".  Even
-            -- though "p" could not be fully satisfied, reducing it may have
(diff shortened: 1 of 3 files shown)


nanavati (dismissed):
This does seem to be a useful fix in the direction of getting the typechecker to commit to more  instance matches sooner and more completely.

src/comp/TCMisc.hs
```diff
@@ -450,8 +450,20 @@ satMany' dvs es rs_accum sbs s (p:ps) = do
             -- prevented satisfying -- is this apSub just covering for an
             -- issue elsewhere (that should not have returned unsubst'd preds)?
             --
+            -- We also apply "s'" to the not-yet-processed preds "ps".  Even
+            -- though "p" could not be fully satisfied, reducing it may have
+            -- committed improvements (e.g. grounding a variable via a
+            -- functional dependency).  A sibling pred whose own reduction is
+            -- determined by such a variable must see the improvement, otherwise
+            -- it is matched against an unbound variable and left unreduced --
+            -- and its own fundep never fires.  (E.g. FilterField grounds
```
quark17: What is this "e.g." comment referring to?  I don't think it's an example in the repo?

quark17 (commented):
I'm trying to understand.  If someone writes `(Bits Bool sz)`, that should not be an error about `sz` being too general (rather than `1`)?  And I guess that is already allowed, as the following definition typechecks, even before your PR:
```
foo :: (Bits Bool sz) => Bit sz
foo = pack True
```

Given that we already extend the substitution, and have the comment "we keep growing the substitution because anything we have learned we can commit to", it seems obvious that the substitution should be applied to everything -- if we're committing to it, we should fully commit to it -- and the fact that we weren't applying it to `ps` (or `needed`) seems like an oversight.  (Unless we think the substitution should be getting applied elsewhere, and the right fix is to fix that other place?)

I don't know that we need such a huge comment on the change, though; can we express the info more succinctly?

src/comp/TCMisc.hs
krame505: Ah I had a more complicated test case when I was debugging this, and I later replaced it with a simpler one.  I agree this comment isn't needed.

krame505:
Yeah, I think this was an oversight.  I simplified the comment.

quark17 (approved):

PR #981, opened 2026-06-26
typecheck: make isTFun O(log n) via a precomputed Set

matx-amy:
isTFun (and PredTrie.isTFunTyCon) tested membership with
`i `elem` numOpNames ++ strOpNames`, which rebuilds the (10-element)
list and does a linear scan of idEq comparisons on every call. isTFun is
called once per type constructor during type expansion/normalization, so
on wide types it dominates: profiling a 521-port always_ready struct
method (a port-shape boundary module) showed id_fs/idEq/isTFun as the top
cost centres, ~30% of allocation.

Precompute the operator-name set once (isTypeFunOp in TypeOps) and use
Set membership instead. Semantics-preserving. On the struct-form module
above this cut the BscV elaboration ~12% (131.3s -> 115.6s, isolated).

### DIFF
--- a/src/comp/Pred.hs
+++ b/src/comp/Pred.hs
@@ -260,12 +260,9 @@ expandSyn t0 = exp [] f as
         -- Type functions (TIatf) are NOT expanded here; they are handled by
         -- normT via expTFun, which generates the appropriate class constraint.
         exp _ f@(TCon (TyCon i _ _)) as
-          | isTFun i = apTFun f i $ map expandSyn as
+          | isPrimTFunName i = apTFun f i $ map expandSyn as
         -- f does not contain a TAp or a synonym TCon, so it cannot have synonyms to expand
         exp _ f as = foldl TAp f $ map expandSyn as
-
-isTFun :: Id -> Bool
-isTFun i = i `elem` numOpNames ++ strOpNames
 
 apTFun :: Type -> Id -> [Type] -> Type
 apTFun _ i [TCon (TyNum x px), TCon (TyNum y py)] | Just n <- opNumT i [x, y] = TCon (TyNum n p')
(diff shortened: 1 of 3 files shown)


nanavati (commented):
I know they're both small but please separate the two independent changes into two separate PRs. we might care about bisecting in the future.

matx-amy:
> I know they're both small but please separate the two independent changes into two separate PRs. we might care about bisecting in the future.

Done!

nanavati (dismissed):
Wouldn't it be even better if you built the set once rather than once per call?

GHC might take care of it for you, but it is a small change that removes that uncertainty.

Otherwise clean improvement.

matx-amy:
yeah, it does; the original version of this was more circuitous (https://github.com/MatX-inc/bsc/commit/c7033c4e3f664f9c9ddb7aed3a125f023d85b807) and did actually have it separate, I benched this against that to be sure I didn't break anything; floated it back out now.

src/comp/Pred.hs
```diff
@@ -265,7 +266,10 @@ expandSyn t0 = exp [] f as
         exp _ f as = foldl TAp f $ map expandSyn as
 
 isTFun :: Id -> Bool
-isTFun i = i `elem` numOpNames ++ strOpNames
+isTFun i = i `S.member` tfunOpNames
```
quark17: Can I suggest moving both `isTFun` and `tfunOpNames` into `TypeOps.hs`?

And, while doing that, take the opportunity to rename `isTFun` to something like `isTFunName` or, perhaps better, `isPrimTFunName` (since this is only about the built-in type functions that need special handling).

quark17: And I guess, `PredTrie.isTFunTyCon` would then be worth renaming to `isPrimTFunTyCon`.

quark17 (approved):

PR #983, opened 2026-06-27
Port splitting part 5: SplitVector

krame505:
Add a `SplitVector` utility, as discussed on Friday.

This is based on #959 not #967, and should be included in the next release.


src/Libraries/Base1/SplitPorts.bs
```diff
@@ -199,3 +209,60 @@ instance (DeepSplitPorts a p) => DeepSplitPorts'' (Conc a) p where
   deepSplitPorts'' (Conc x) = deepSplitPorts x
   deepUnsplitPorts'' = Conc ∘ deepUnsplitPorts
   deepSplitPortNames'' _ = deepSplitPortNames (_ :: a)
+
+-- A newtype wrapper for a Vector that recursively splits its elements into
+-- the corresponding SplitPorts ports.
+data SplitVector n a = SplitVector (Vector n a)
```
matx-amy: needs a DefaultValue instance

krame505: Good catch, done.

nanavati: Should we use generalized new type deriving here?

I wonder if we should expose the capability more directly now that we have deriving-via.

nanavati (dismissed):
Other than my generalized newtype deriving question, looks good.

src/Libraries/Base1/SplitPorts.bs
```diff
@@ -199,3 +209,60 @@ instance (DeepSplitPorts a p) => DeepSplitPorts'' (Conc a) p where
   deepSplitPorts'' (Conc x) = deepSplitPorts x
   deepUnsplitPorts'' = Conc ∘ deepUnsplitPorts
   deepSplitPortNames'' _ = deepSplitPortNames (_ :: a)
+
+-- A newtype wrapper for a Vector that recursively splits its elements into
```
quark17: I would be tempted to create a separate `SplitVector.bs` package.

src/Libraries/Base1/SplitPorts.bs
krame505: I already am for most of the things that can; I guess the Bounded instance can be, but the other type classes have the wrong shape.

src/Libraries/Base1/SplitPorts.bs
krame505: Yeah, sure.

quark17 (approved):

PR #984, opened 2026-06-29
Add Functor/Traversable/Foldable instances for UAList

krame505:


quark17 (approved):
Looks good.  There's no tests added, but there were no existing tests for this library, so it's not making things worse.  I even didn't recall that this library existed!

krame505:
Yeah, I didn't know it existed either, until a coworker of mine used it in a pull request.

PR #987, opened 2026-07-01
Make Bluesim C++ code ordering more stable

nanavati:
Fix a number of places where Bluesim C++ code ordering depended on the internal ids generated when interning FStrings

### DIFF
--- a/src/comp/SimMakeCBlocks.hs
+++ b/src/comp/SimMakeCBlocks.hs
@@ -292,7 +292,10 @@ onePackageToBlock flags name_map full_meth_map ss pkg =
       meth_rets = [ (rt, n, vn)
                   | (n, (_,_,(Just (rt,vn)),_,_)) <- M.toList meth_map
                   ]
-      ports = meth_ens ++ meth_args ++ meth_rets
+      -- Sort by base name so the order doesn't depend on the AId map order
+      -- (AId's Ord follows run-dependent interned-FString order).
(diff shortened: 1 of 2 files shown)


src/comp/SimMakeCBlocks.hs
```diff
@@ -1430,6 +1436,20 @@ tsortActionsAndDefs modId rId mmap ds acts reset_ids =
         -- Convert the graph to the format expected by tsort.
         g_edges = M.toList g
 
+        -- tsort breaks ties by node Ord (for defs, the run-dependent AId Ord),
+        -- so rank def nodes by id-name before tsort and map back for a stable
+        -- order.  (Actions keep their position; Left<Right keeps defs first.)
+        g_def_ids = nub [ i | (n,ns) <- g_edges, Left i <- n:ns ]
```
quark17: I would tend to put explicit types on some of these definitions -- but I'll leave that up to you.

quark17 (dismissed):
Awesome!  Looks good.

src/comp/SimMakeCBlocks.hs
nanavati: done along with a few performance cleanups

nanavati:
> Awesome! Looks good.

Great! After the cleanup is this ready to go?

quark17 (approved):
> After the cleanup is this ready to go?

I think so.  Did you want to preserve the commits or squash merge?

nanavati:
> > After the cleanup is this ready to go?
> 
> I think so. Did you want to preserve the commits or squash merge?

No preference. Whatever you think is best.

PR #988, opened 2026-07-01
Block-level C++ code generation

nanavati:
Add a mode for block-level C++ code generation where the output is identical to when the block is produced by compiling a simulation top-level. This makes it easier to work with build systems that want to orchestrate the final C++ compilation (like Bazel).

This depends on the C++ code-ordering fixes from #987 and thoroughly tests those orderings by generating modules that would differ if those orderings were not stable.


src/comp/Error.hs
```diff
@@ -4493,6 +4494,11 @@ getErrorText (EMissingVPIWrapperFile fname is_dpi) =
      in  s2par ("Cannot find the " ++ ifctype ++ " file " ++ ishow fname ++
                 " in the Verilog search path."))
 
+getErrorText EBlockCodegenNoSim =
```
quark17: I would avoid having a new message for every flag.  I'd make a general error (assuming we don't already have one) that is parameterized by the flag and the backend.

src/comp/SimBlocksToC.hs
```diff
@@ -95,8 +96,13 @@ simBlocksToC flags time top_block def_clk def_rst
 
     let cvtModBlock = convertModuleBlock flags sb_map ff_map clk_map wdef_mod_map reused top_block
     module_names <- concatMapM (cvtModBlock writeFileC) mod_blocks
-    schedule_names <- convertSchedules flags time top_block def_clk def_rst sb_map ff_map
-                                       wdef_inst_map scheds clk_groups gate_info writeFileC
+    -- With -block-codegen the root is a reusable block, not a runnable top: it
```
quark17: Can you explain what's going on here?  I'm confused because BSC already doesn't generate a runnable top, so I don't understand why the current codegen doesn't do what you want.

src/comp/SimCOpt.hs
```diff
@@ -123,7 +123,19 @@ moveDefsOntoStack flags instmodmap (blocks,scheds) =
                    , let unqual_id = unQualId qual_id
                    , let sbid = sb_id blk
                    ]
-      sched_refs = [ (k, dr) | (k, (dr,_)) <- sched_tups ]
+      -- Drop own-schedule reads of top interface-method CAN_FIREs: a submodule
+      -- reaches that readiness by the RDY call, so this matches its member/local
+      -- classification (see DEVELOP.md).
```
quark17: Can you explain this?  (And what does it mean to see `DEVELOP.md`?)

quark17 (commented):
Ok, I think I'm slowly understanding.  There's two things going on here.  One, you just don't want BSC to run the C++ compiler or linker.  Two, you want BSC to generate the C++ code for modules separate from creating a top-level sim -- similar to how the Verilog flow works: `bsc -g` creates Verilog files, then `bsc -e` does the linking.  Can we treat those as separate features, or are they bound up somehow?

You've added a few module generation changes, that I assume are about part two, which I would need to look at closer to understand, unless you can explain them?

The `-e` link stage is about creating a top-level sim.  So if you want BSC to just generate the individual module code, without a top-level, I'd rather do that in some other mode than the `-e` link mode.  Perhaps the `-g` mode can take a flag that says, "don't just generate .ba files but also give me the C++ code for the modules".  Or else, we create new modes for Bluesim (say, a `-s` mode that goes from `.ba` to top-level agnostic C++ files, and then the `-e` only needs to generate the schedule file and possibly a top model object).

Off hand, I don't recall how much of Bluesim's module codegen is dependent on knowing the full model -- like, do we do any cross module optimizations.  The current process of generate the modules during `-e` link mode allows for that dependency; if we generate the modules separately, we lose that ability, but maybe that's OK.

Anyway, I would really suggest renaming the flag, because I had no idea what "block" meant here (and at first didn't know what the flag had to do with "codegen" either).  Although that may be moot if we separate the two features into separate flags and implement the second one as part of a BSC mode (`-g` or some new mode with a meaningfully named flag).

src/comp/SimCOpt.hs
nanavati: There is a DEVELOP.md at top-level of the bsc repo. For the existing, purely time-based reuse check to be valid the code needs to be identical whether you generate it via linking or via a simulation top-level. The detailed argument for why the output will be identical is made there.

This problem here is the own-schedule reads of the interface-method CAN_FIREs can add extra reads to the CAN_FIREs, which can change C++ code generation. This isn't a top-level so those reads should not be counted and this code corrects that for the block-level case.

quark17: I did not see that there were changes to `DEVELOP.md`, because GitHub's web interface put it last in the list of changed file and I stopped reading when I got to the testsuite changes as I assumed that was all that was left -- sorry!  I'll have to go read what's written in that file.  (Words like "own-schedule" and "block-level" are opaque.)

nanavati:
@quark17 Thanks for pushing back on my initial version of this, which was itself a cleanup of an even hackier attempt: MatX-inc/bsc#21

The core story is that we want to generate C++ files like Verilog files are generated, but this originally did that by conflating compilation stages in a confusing way. I've tried to fix that in the new version.

This PR now implements the new structure for the Bluesim backend and nanavati/bsc#1 and nanavati/bsc#2 extend the same structure to the Verilog backend.

My proposed new structure is to add a "codegen" step, controlled by the `-c` flag, which is intended to be parallel to the `-g` and `-e` flags.

In the new world we have three steps to making a simulation:

1. `-g` for elaboration: which is  Bluespec source -> `.ba` file. 
2. `-c` for the new codegen stage: this is `.ba` -> `.cxx` or `.v`
3. `-e` for the new link stage which is really `.ba` -> executable, using either backend and generating any missing or stale pieces of code.

Right now this PR adds `-c` for Bluesim. nanavati/bsc#1 adds `-c` for Verilog (and flips the default so `.ba` files are produced by default in the Verilog flow) since I am starting to discover reasons why `.ba` is particularly handy in the Verilog flow (like detecting when you are using primitives Verilator can't easily support).

So far the implementation is pretty basic and there are questions I haven't resolved. For example, should bare `-g` just go to a backend-agnostic `.ba` (and error if `genC` or `genVerilog` is used)? Would that mean that `-sim` and `-verilog` beyond specifying a backend are also shorthand for triggering the codegen phase?

I am sure there are plenty of additional issues that come from this ambitious a refactor. I just decided it made sense to lay out the whole structure so we can see it and make decisions based on how things should work holistically.

nanavati:
One more fix: Make sure that top-level and block-level forms of compiled modules are distinguished so that recompilation checking can work properly.

PR #990, opened 2026-07-03
testsuite: reflect/TypeOf use $display so verilator output isn't clobbered

nanavati:
A newline-less $write puts verilator's "$finish" banner on the same line as the value, and clean_verilator_output strips it along with the value. $display adds a trailing newline so all simulators agree; this test checks type reflection, so the newline is immaterial to what it exercises.

## Testing

Verified standalone on a branch cut from `main`: `bsc.typechecker/reflect` runs clean under iverilog with the Bluesim side enabled (20 PASS, 0 FAIL).

---


testsuite/bsc.typechecker/reflect/TypeOf.bsv
```diff
@@ -2,7 +2,12 @@
 module sysTypeOf();
 
    rule test;
-      $write(printType(typeOf(1)));
+      // Use $display (with trailing newline) rather than $write: a newline-less
```
quark17: I don't think this comment is needed.  The tip can be put in a global document of tips/conventions for writing tests -- the testsuite README has a section header "Test suite structure" and maybe a section under that can be a list of other conventions/tips.

quark17 (approved):

quark17 (approved):

PR #992, opened 2026-07-03
testsuite: run backend-independent dump checks regardless of $ctest

nanavati:
vector_interfaces and relax-schedule each generate a compile-time,
backend-independent dump (-dwrapper_typecheck / -dvschedinfo) and then check it,
but generated it with compile_object_pass -- which is guarded by
"if {$ctest == 1}" -- while the consuming find_n_strings / compare_file check is
ungated.  Under a -no-ctest run (e.g. a verilog-only run) the setup is skipped,
so the check runs against a file that was never created and fails spuriously.
(The failure was previously masked by stale dump files left over from earlier
-ctest runs.)

Generate the dump with an ungated compile_pass ... -verilog instead.  Both dumps
are backend-independent -- the .sched is byte-identical under -verilog, and the
wrapper-typecheck dump has the same match count -- so the goldens are unchanged
and both tests now run and pass under -ctest and -no-ctest alike.

## Testing

Verified standalone on a branch cut from `main` (iverilog):
- `bsc.codegen/vector_interfaces`: CTEST=0 7/0 (previously failed on the missing dump file), CTEST=1 7/0
- `bsc.scheduler/relax-schedule`: CTEST=0 34/0 (previously failed), CTEST=1 60/0

---

### DIFF
--- a/testsuite/bsc.codegen/vector_interfaces/vector_interfaces.exp
+++ b/testsuite/bsc.codegen/vector_interfaces/vector_interfaces.exp
@@ -10,8 +10,5 @@ compile_verilog_pass SizeZero.bsv
 # to ensure that theinstance dictionaries can be combined for vector-blasted interfaces.
 # This relies on the -dtypecheck dump from wrapper compilation
 # overwriting the original -dtypecheck dump.
-# The wrapper-typecheck dump is a compile-time (backend-independent) artifact, so
-# use an ungated compile_pass (compile_object_pass is $ctest-gated, which left the
-# ungated find_n_strings below checking a never-created file under -no-ctest).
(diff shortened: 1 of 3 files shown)


quark17:
Normally, I would just add a guard of `$CTEST==1` or `$VTEST==1` on these kinds of tests, depending on which arbitrary backend they chose.  That, though, causes the test to be skipped if the particular backend is off.

I like this idea (of codegen for the backend regardless of the testsuite setting), but the comments in the exp files are confusing and verbose -- what if we defined `compile_backend_pass` in `config/unix.exp` and have that hide all the implementation, so we just simply call it in the test tcl?

There's probably a ton of places that guard the follow-on steps, that we may want to update at some point?  Also, do we want to add a CTEST=0 and VTEST=0 run to the CI?

quark17 (approved):

PR #995, opened 2026-07-03
testsuite: fix MPEG4 (sort ordering, check_vcd arg, bad bitstream sentinel)

nanavati:
Three fixes that together make MPEG4 pass under Bluesim, iverilog and
verilator:

1. The Verilog output differed under verilator only by the order of two
   coincident $displays from separately-synthesized modules (a single swap in
   30k lines; the outputs are sorted-equal).  Enable sort_output on the
   Verilog comparison so all simulators share the golden.

2. The $check_vcd=0 argument (intended to disable the VCD check "because the
   files are too large") was landing in the sort_output slot, so the giant
   VCD-dump run had silently been executing all along.  Plumb it to the right
   position on both the Bluesim and Verilog calls.

3. bitstream.txt ended with a malformed 'ffffffff' sentinel in an 8-bit
   stream: Bluesim's "Malformed value" load error was baked into the .c
   golden, and iverilog 12+ newly warns "Excess hex digits" at load, breaking
   the .v comparison.  Fix the data (-> 'ff', the value Verilog simulators
   truncated to anyway; the address is never read), drop the error from the
   .c golden, and accept Bluesim's now-accurate gap warning (the file has
   always been one entry short of the declared range).  Also update the
   testbench comments that documented the old loader error.

Verified: verilator CTEST=1 8/0 (Bluesim + sorted Verilog), iverilog 4/0,
no VCD-dump runs, 0 XPASS.

## Testing

Verified standalone on a branch cut from `main`: iverilog with the Bluesim side enabled runs clean (8 PASS / 0 FAIL), and `testrun.sum` contains no VCD-dump runs, confirming the `check_vcd` argument now lands in the right slot.

---

### DIFF
--- a/testsuite/bsc.long_tests/MPEG4/Testbench.bsv
+++ b/testsuite/bsc.long_tests/MPEG4/Testbench.bsv
@@ -16,11 +16,9 @@ module mkTestbench ();
    Mpeg4_IFC dut();
    mkMpeg4 the_dut(dut);
 
-   // Error: value processing error at line 208670 of file 'bitstream.txt'
-   //    Malformed value.
-   // Warning: file 'y.txt' does not completely fill the memory 'the_y_io'.
-   // Warning: file 'u.txt' does not completely fill the memory 'the_u_io'.
-   // Warning: file 'v.txt' does not completely fill the memory 'the_v_io'.
+   // The data files do not completely fill their memories, so the loader
+   // warns about a gap at the end of 'bitstream.txt', 'y.txt', 'u.txt'
+   // and 'v.txt'; the unfilled addresses are never read.
 
    RegFile#(Bit#(20),Bit#(8)) stimulus_io();
    mkRegFileLoad#("bitstream.txt",0,208670) the_stimulus_io(stimulus_io);
--- a/testsuite/bsc.long_tests/MPEG4/Testbench.bsv
+++ b/testsuite/bsc.long_tests/MPEG4/Testbench.bsv
@@ -16,10 +16,6 @@ module mkTestbench ();
    Mpeg4_IFC dut();
    mkMpeg4 the_dut(dut);
 
-   // The video data files do not completely fill their memories, so the
-   // loader warns about gaps in 'y.txt', 'u.txt' and 'v.txt'; the unfilled
-   // addresses are never read.
-
    RegFile#(Bit#(20),Bit#(8)) stimulus_io();
    mkRegFileLoad#("bitstream.txt",0,208669) the_stimulus_io(stimulus_io);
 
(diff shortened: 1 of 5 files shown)


testsuite/bsc.long_tests/MPEG4/mpeg.exp.golden
```diff
@@ -7,15 +7,22 @@ set env(GHCRTS) "-K20M"
 # disable VCD check because the files are too large
 set check_vcd 0
 
+# (the trailing args are: cbug/vbug, link_options, sim_options, sort_output,
+# check_vcd -- $check_vcd was previously landing in the sort_output slot,
```
quark17: Do we need to say what the previous code looked like?

quark17 (commented):
Is the gap part of the test?  Or do we think someone just made an off-by-one mistake when writing the instantiations like:
```
mkRegFileLoad#("bitstream.txt",0,208670)
```
and should we just change the third argument in all of them to be one less? (and remove the expected warnings)

I guess it doesn't hurt to leave it minimally changed.

testsuite/bsc.long_tests/MPEG4/Testbench_novideo.bsv
```diff
@@ -16,14 +16,12 @@ module mkTestbench_novideo ();
    Mpeg4_IFC dut();
    mkMpeg4 the_dut(dut);
 
-   // Error: value processing error at line 208670 of file 'bitstream.txt'
-   //    Malformed value.
-   // Warning: file 'y.txt' does not completely fill the memory 'the_y_io'.
-   // Warning: file 'u.txt' does not completely fill the memory 'the_u_io'.
-   // Warning: file 'v.txt' does not completely fill the memory 'the_v_io'.
+   // The video data files do not completely fill their memories, so the
```
quark17: My thought was that this entire comment and all the warnings at the start of the expected output could go away, by subtracting one from the upper index in the instantiations for `the_y_io`, `the_u_io`, and `the_v_io`, too.

testsuite/bsc.long_tests/MPEG4/mpeg.exp.golden
nanavati: No, dropped

testsuite/bsc.long_tests/MPEG4/Testbench_novideo.bsv
nanavati: Done in 26b707a4 — all three video memories are now sized exactly to their data (I verified the line counts: y.txt is 10,644,480 entries, u.txt and v.txt are 2,661,120 each, so every "gap" was just the one-past upper index, same as the bitstream). The warnings, their lines in the expected output, and the comment are all gone. Full mpeg.exp run (Bluesim + Verilog) passes 8/8 with byte-identical output.

testsuite/bsc.long_tests/MPEG4/Testbench.bsv
```diff
@@ -16,14 +16,12 @@ module mkTestbench ();
    Mpeg4_IFC dut();
    mkMpeg4 the_dut(dut);
 
-   // Error: value processing error at line 208670 of file 'bitstream.txt'
-   //    Malformed value.
-   // Warning: file 'y.txt' does not completely fill the memory 'the_y_io'.
-   // Warning: file 'u.txt' does not completely fill the memory 'the_u_io'.
-   // Warning: file 'v.txt' does not completely fill the memory 'the_v_io'.
+   // The video data files do not completely fill their memories, so the
```
quark17: This file isn't being tested.  If we're going to update it, then update it the same as `Testbench_novideo.bsv` (by deleting this comment and fixing the off-by-one in the other instantiations).

I hesitate to suggest deleting the file, as there's probably other files that could be deleted, and maybe we want to preserve this as an example?  The difference between this file and `Testbench_novideo.bsv` is the instantiation of a `mkScreen` module and reducing of the number of frames from 30 down to 3, to be reasonable simulation run-time for the testsuite.  (In theory, the two files could be unified, with preprocessor macros to select inserting the screen; but no need for that effort in this PR.)

nanavati: Done — Testbench.bsv now gets the identical treatment (indices sized to the data, comment removed). Agreed on keeping the file as the with-video example and leaving any unification for another day.

quark17 (approved):

PR #997, opened 2026-07-03
Add -system-verilog-output flag; emit SystemVerilog string parameters

nanavati:
Rename the internal flag systemVerilogTasks -> systemVerilogOutput.  -system-verilog-output is now the canonical spelling and -system-verilog-tasks is kept as a deprecated, backward-compatible spelling of it (both are Toggles setting the same field, so -no-system-verilog-tasks still works).

Under the flag, a String module parameter is emitted as a SystemVerilog 'parameter string' rather than an untyped bit-vector parameter, so SV consumers (VCS, verilator) treat it as a string.  parameter string is IEEE 1800-2005+ (rejected by classic Verilog), hence opt-in.  Verilog task emission ($error/$fatal via the former -system-verilog-tasks) is unchanged.

Update the -print-flags-raw golden for the field rename.

## Testing

Verified standalone on a branch cut from `main`:
- Builds cleanly; `-system-verilog-output` appears in `-help-hidden`; the deprecated `-system-verilog-tasks` spelling is still accepted.
- `bsc.options`: 70 PASS / 0 FAIL (including the `-print-flags-raw` golden)
- `bsc.verilog/parameters`: 60 PASS / 0 FAIL / 1 XFAIL
- `bsc.verilog/parameters/string`: 52 PASS / 0 FAIL

---

### DIFF
--- a/src/comp/AVerilog.hs
+++ b/src/comp/AVerilog.hs
@@ -24,7 +24,8 @@ import Util
 import FileNameUtil(hasSuf)
 import PFPrint
 import Error(internalError, ErrorHandle)
-import Flags(Flags, removeReg, removeCross, removeInoutConnect, removeUnusedMods,
+import Flags(Flags, systemVerilogOutput,
+             removeReg, removeCross, removeInoutConnect, removeUnusedMods,
              useDPI, verilogDeclareAllFirst)
 import Id
 import Pragma(PProp(..))
@@ -322,15 +323,19 @@ aVerilog errh flags pps aspack ffmap =
         -- XXX might be good to allow the user to specify a default
 
         args :: [VArg]
-        args =        [ VAParameter (vId i) r v
+        args =        [ VAParameter (vId i) r v isStr
                      | (i, t) <- ps,
-                       let (r, v) = case t of
+                       let (r, v, isStr) = case t of
                                       ATBit sz -> (Just (VEConst (sz-1),
                                                          VEConst 0),
-                                             VEWConst (mkVId "0") sz 2 0)
-                                      ATString _ -> (Nothing, VEString "")
-                                      ATReal -> (Nothing, VEReal 0.0)
-                                      _ -> (Nothing, VEConst 0) ] ++
+                                             VEWConst (mkVId "0") sz 2 0, False)
+                                      -- under -system-verilog-output, emit a SV
+                                      -- `string` parameter so $display treats it
+                                      -- as a string, not a packed bit-vector
+                                      ATString _ -> (Nothing, VEString "",
+                                                     systemVerilogOutput flags)
+                                      ATReal -> (Nothing, VEReal 0.0, False)
+                                      _ -> (Nothing, VEConst 0, False) ] ++
                 [ VAInput (vId i) (vSize t)
                      | (i, t) <- is ] ++
                 filterSharedInout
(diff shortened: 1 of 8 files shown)


quark17:
How does this relate to the `-v95` flag?  We should eliminate that, since BSC already outputs non-v95-isms.  This is proposed in PR #719 (which continues the discussion from PR #400) -- there was a lingering question of whether the `Either` in `vi_inst_params` that controls positional vs named instantiation arguments could be replaced with just named, but it was unclear if there were any legitimate users remaining.

nanavati:
Two parts to answer — the `-v95` relationship, and the `Either` in `vi_inst_params`.

**On `-v95`:** this PR is orthogonal to it — it only touches the module-definition side (`VAParameter`) under the new flag, and doesn't read or change the v95 paths. On eliminating `-v95` itself: agreed, and I think the evidence is even stronger than "BSC already outputs non-v95-isms." The flag gates exactly three things in generated code: named→positional instance parameters (two sites), and `$signed`/`$unsigned` replaced by a *comment* — i.e. the cast is silently dropped, which is lossy rather than compatible. Meanwhile the shipped primitive library is not gated at all: `SizedFIFO.v` (and the Probe primitives) use `localparam`, and `SyncFIFOLevel.v` uses named parameter instantiation internally — so `-v95` output hasn't been V95-clean for any design with a sized FIFO in years. Making it genuinely clean would mean maintaining V95 variants of the hand-written library, for an audience that no longer exists. So: reland #719; happy to help drive that if useful.

**On the `Either` in `vi_inst_params`:** I chased down every construction site. Two of the three are the `-v95` fallback and die with #719. The third — noinline-function instantiation — splits into two populations: BSV `(* noinline *)` functions are enforced monomorphic (T0111), so they never have parameters at all; Classic `foreign` functions applied at numeric types do pass them (`Fork` is the in-tree example), and those are positional **by necessity**, not style — the declaration syntax (`foreign vfork :: Bit n -> Bit m = "Fork",("i","o")`) names only ports, so the compiler never learns the hand-written module's parameter names (`Fork.v`'s `iw`/`ow`); the values pass in type-argument order matching the parameter declaration order. I've put up #1021 routing the always-empty noinline case through the named side and documenting the Classic-foreign contract at both the emission site and the `ANoInlineFun` type — so after #719, the residual positional arm is exactly that one legacy feature, and full collapse would require either extending the Classic foreign syntax to name parameters or retiring numeric-polymorphic foreign.

(I also checked the inout paths specifically, since I wondered about them: inouts flow entirely through `vi_inst_ports`, which is already named-only, and bsc never emits raw gate primitives — so nothing there needs positional.)

nanavati:
Following up on my last comment: the `Either` turns out to be fully collapsible after all. The "unknowable" parameter names of module-bound Classic foreign functions aren't unknowable — the declaration names each parameter slot with a *type variable*, so the convention is simply that the module's parameter names are the declaration's type variables (Fork.bs conforms to the existing Fork.v by renaming its type variables to `iw`/`ow`). #1022 implements named-parameter emission everywhere, removes `-v95` (relanding the intent of #719), and adds a check (T0159) that module-bound foreign widths are bare variables or literals — a derived width has no name to pass, and relationships belong in provisos once contexts are permitted on foreign declarations. It supersedes #1021.

PR #998, opened 2026-07-03
Emit SystemVerilog-safe identifiers, with a warning (fixes h264 on verilator)

nanavati:
BSV's parser rejects Verilog-2001 keywords as identifiers, but words that are
reserved only in SystemVerilog pass through and are emitted verbatim -- so a
design with a register named "process" (bsc.bsv_examples/h264) produces
Verilog that no SystemVerilog-mode consumer can parse (verilator with DPI
fails with "syntax error ... expecting '{").

Three-part fix:

1. The Verilog printer emits identifiers that collide with SV-reserved words
   as escaped identifiers (\global ), which are legal in both Verilog-2001
   and SystemVerilog, so the output parses under either language level.  The
   set is the IEEE 1800-2017 reserved words minus Verilog-2001, plus the std
   package class names.  (Of these, BSV's parser currently admits process,
   semaphore, mailbox, global, soft, strong and weak; the parser's own
   keyword table stops at SV 3.1.)

2. Escaping cannot help with the std built-in class names (process,
   semaphore, mailbox): verilator resolves even the escaped identifier to
   the std class.  Since generated code never uses the std package, pass
   --no-std to verilator.

3. Verilog generation warns once per colliding identifier (G0129, with the
   identifier's source position), since users may prefer to rename.  The
   warning is suppressible and can be promoted to an error with
   -promote-warnings G0129 for a strict "reserved word" policy; reserving
   the words outright in the parser was rejected as a breaking change
   (h264 itself would stop compiling).

Adds bsc.verilog/svkeywords testing all seven admitted identifiers: compile
(expecting 7 G0129 warnings), link and run identically under Bluesim,
iverilog and verilator.

Verified: h264 now compiles, links and runs under verilator (previously the
last verilator-specific failure besides log2_loop); signal_names, vcd, b925
and Hierarchy regressions unchanged.

## Testing

Verified standalone on a branch cut from `main`:
- Builds cleanly; a register named `process` produces the G0129 warning with position, is emitted as `\process `, and the design links and runs under both iverilog and verilator; `-suppress-warnings G0129` silences it and `-promote-warnings G0129` turns it into an error (an opt-in strict "reserved word" policy).
- New `bsc.verilog/svkeywords` test: 11 PASS / 0 FAIL under both iverilog and verilator.
- `bsc.bsv_examples/h264` now links under verilator (its `mkInverseTrans` has a register named `process`; the link previously failed with `syntax error, unexpected ==, expecting "'{"`).

---


src/comp/Verilog.hs
```diff
@@ -1081,6 +1090,45 @@ vKeywords =
...
 
+-- Words that are reserved in SystemVerilog (IEEE 1800) but not in
+-- Verilog-2001.  BSV's parser rejects Verilog-2001 keywords as identifiers,
+-- but these can appear as user names (e.g. a register named "process") and
+-- would break any SystemVerilog-mode consumer of the generated code (for
+-- example verilator with DPI, or VCS with -sv), so the printer emits them as
+-- escaped identifiers.  The list is the 1800-2017 reserved words minus the
+-- Verilog-2001 set, plus the built-in std package class names, which parsers
+-- treat as type names.
+svKeywordSet :: S.Set String
```
quark17: Can this be unified with `SystemVerilogKeywords.lhs`?

src/comp/Error.hs
```diff
@@ -3119,6 +3121,12 @@ getErrorText (ESplitInsideNoSplit expr) =
 getErrorText (WMissingRule s) =
     (Generate 17, empty, s2par ("Rule not found: " ++ ishow s))
 
+getErrorText (WSVReservedIdent name) =
```
quark17: Put this numerically in order (after `Generate 128`).

I have less concern about where to put it in the type declaration.  I'm not sure it belongs in between the scheduling-related tags, but there's no real order there (except that Generate tags are together).  I don't think we have any user warnings/errors after scheduling, in the Verilog codegen branch, so this would be the first?

src/comp/Verilog.hs
nanavati: I don't think so. These are identifiers `bsc` has accepted for a long time so we don't want to just reject them. If we did, we'd just break designs like `h264`. Instead a specific warning can be promoted by anyone who just wants to ban the identifiers without breaking backwards-compatibility.

tl;dr: Two different policies means two different lists.

src/comp/Error.hs
nanavati: Fixed

src/comp/Verilog.hs
quark17: I think you misunderstand?  All of these keywords should already be in the list in `SystemVerilogKeywords.lhs`.  I'm not proposing anything that would change the parser behavior.  (If there are any keywords in the AVerilog list that are not in SystemVerilogKeywords, though, it would be good to know that!)

Inside `SystemVerilogKeywords.lhs` is defined `svKeywordTable` which is a list of tuples, of words and the language that defines it as a keyword -- such as `SystemVerilog31`, `SystemVerilog30`, `Verilog2001`.  It does also include `Bluespec38` keywords.

Would it not make sense to have a single canonical list of SV keywords?  Either (A) treating `svKeywordTable` as the canonical list and having `AVerilog` filter for the words that are Verilog/SystemVerilog relevant; or (B) have a canonical list in a third location, that is shared by `AVerilog` and by `SystemVerilogKeywords.lhs` (which can create its list from the union of several lists, perhaps -- or verify that all of the keywords in the canonical list are covered in its map).

src/comp/Verilog.hs
```diff
@@ -1081,6 +1090,45 @@ vKeywords =
     "localparam", "medium", "highz1", "initial"
     ]
 
+-- Words that are reserved in SystemVerilog (IEEE 1800) but not in
+-- Verilog-2001.  BSV's parser rejects Verilog-2001 keywords as identifiers,
+-- but these can appear as user names (e.g. a register named "process") and
+-- would break any SystemVerilog-mode consumer of the generated code (for
+-- example verilator with DPI, or VCS with -sv), so the printer emits them as
+-- escaped identifiers.  The list is the 1800-2017 reserved words minus the
+-- Verilog-2001 set, plus the built-in std package class names, which parsers
```
quark17: Why are we subtracting Verilog-2001 keywords?  If the user used escaping to name a signal `always` or `buf`, then we'd want to escape that in the output too?  Does BSC already do that? even for BH source that doesn't go through the BSV parser?  (Incidentally, I don't see `process` in the keywords that BSV knows about.)

src/comp/Verilog.hs
```diff
@@ -757,7 +759,14 @@ getVIdString :: VId -> String
...
 
 instance PPrint VId where
-        pPrint d p (VId s i _) = text s
+        pPrint d p (VId s i _)
+          -- identifiers that are reserved in SystemVerilog (but not in
+          -- Verilog-2001, which BSV's parser already rejects as identifiers)
+          -- are printed as escaped identifiers, so the generated code parses
+          -- under both language levels (escaped identifiers are legal in
+          -- both; the trailing space is part of the escape syntax)
+          | s `S.member` svKeywordSet = text ("\\" ++ s ++ " ")
```
quark17: You defined `isSVReservedWord` below, but then didn't use it here?

quark17:
I'm not sure that it's safe to pass `--no-std` to Verilator.  Users can import their own SV into a design.

nanavati:
Heads-up for whichever of #998 / #999 merges second: #999 moves DPI imports to module scope as a new `VMItem` constructor (`VMDPI`), while this PR's `vModuleDeclVIds` decl scan is deliberately written without a wildcard so new constructors surface. The combination crashes `bsc -verilog -use-dpi` on any BDPI design (non-exhaustive patterns in `go`) until the scan learns the new constructor:

```haskell
go (VMDPI (VDPI n _ _ _ args)) = n : [ a | (a, _, _) <- args ]
```

(the DPI function name is a foreign-fn name per the scan's invariant, and the argument names are printed declaring occurrences). `instance Ord VMItem` needs a matching arm as well — it is likewise exhaustive-without-wildcard, and DPI imports sort first:

```haskell
compare (VMDPI _) (VMDPI _) = EQ
compare (VMDPI _) _         = LT
compare _         (VMDPI _) = GT
```

Neither branch can host the fix alone (each lacks the other's definitions). Found by adversarial review of a combined tree carrying both PRs; the fix above is validated there (`-verilog -use-dpi` compiles and emits the module-scope DPI-C import).

Issue #170, opened 2020-05-05
Add automated linking for Verilator


quark17:
BSC can automate the linking of a Verilog simulation for a number of simulators.  We should add Verilator to that list:
```
bsc -e topMod -verilog -vsim verilator
```

Basic support will require two things: (1) a new script called `bsc_build_vsim_verilator` in `src/exec/` and (2) a `verilator_main.cpp` file (added somewhere in the repo) to perform the function that `src/Verilog/main.v` does for most other Verilog simulators.

Adding support for a new simulator is usually as simple as adding a new script in `src/exec/` with a name starting with `bsc_build_vsim_` and followed by the name of the simulator.  The BSC executable does not contain any information specific to the simulator.  BSC merely calls the appropriate script, using arguments to pass generic information about the design.  The script contains all the knowledge about how to link the simulator, given that generic info.  It should be possible to copy one of the existing scripts and change it for the specifics of Verilator.

The one wrinkle is that Verilog needs a top-level C++ file, which most simulators do not.  For most simulators, we link the user's design with a `main.v` file that instantiates the design, provides a clock and reset, and turns on VCD dumping (if directed by plusargs).  This file is provided in the release and can instantiate any module (by using preprocessor macros to insert the user's specific module name etc).  For Verilator, we will need to a "verilator_main.cpp" that does the same thing -- instantiates a model (as specified by preprocessor macro), provides the clock and reset, enables VCD or tracing, etc.  I'm not sure where this file would live.  Perhaps we create a new directory `Verilog.Verilator` and put it there; or just in `Verilog`; or some new place for simulator files; or something else.  This file doesn't need to be written from scratch but can be adapted from an existing file for the Bluespec RISC-V processors [[link](https://github.com/bluespec/Piccolo/blob/master/builds/Resources/Verilator_resources/sim_main.cpp)].

There's actually a second wrinkle: If a BSV design contains foreign function calls, BSC will generate VPI functions (wrapped around the imported C functions), and the scripts are responsible for linking  those VPI functions into the simulator.  Verilator does not support all of VPI that BSC uses (see issue #169 ), so a first version of the script may have to give an error if VPI files are provided.  Although it may be possible for the script to auto-generate equivalent DPI imports.

jahofmann:
Instead of supporting each simulator directly in BSC the project could use a tool like [Cocotb](https://github.com/cocotb/cocotb) which does the heavy lifting.

Then there would be only one top level Python file for all simulators.

[...]

Issue #169, opened 2020-05-05
Optionally generate DPI instead of VPI


quark17:
Calls to foreign functions (BDPI in BSV) are implemented in the Verilog backend using VPI.  Some simulators may not support VPI, but may support DPI which is an alternative interface introduced in SystemVerilog.  Even when both are supported, DPI may be more efficient.  BSC should be enhanced to allow generating designs that use DPI, in addition to the current ability to generate VPI.  (Any flags that instruct BSC to restrict the Verilog to a standard that doesn't have DPI, such as `-v95`, should conflict with any flags to turn on DPI or should change any implicit choice to use VPI.)

Verilator is an example of a simulator that doesn't fully support VPI, but does support DPI.  BSV designs that use BDPI can be made to simulate with Verilator by replacing the VPI imports with DPI imports.  An example can be seen in Bluespec's RISC-V processors [[link]](https://github.com/bluespec/Piccolo/tree/master/builds/Resources/Verilator_resources).  In those processors, the `DPI-C` declarations have been manually written for each foreign function; having DPI native in BSC would eliminate the need to provide that file.  However, the DPI code is fairly boilerplate, so even if BSC didn't yet generate DPI, I could imagine a workaround that used a post-processing script to automate the VPI-to-DPI conversion.

PR #999, opened 2026-07-03
Support polymorphic import "BDPI" functions over the DPI path (-use-dpi)

nanavati:
Polymorphic BDPI imports previously failed with S0015 ("Polymorphic types are not yet supported with DPI"). SystemVerilog DPI-C imports need concrete widths, so a polymorphic import is now monomorphized: for each concrete width at which it is used, BSC emits one fixed-width DPI import (name-mangled by the polymorphic-operand widths) that links, via a `c_identifier =` alias, to a generated C wrapper which forwards to the single user C function.

A fixed-width packed operand (`bit [W-1:0]`) is passed on the C side as the word-canonical `svBitVecVal*` for every width, which matches the polymorphic C function's word-pointer ABI, so each wrapper just casts and forwards. This avoids DPI open arrays entirely (whose Verilator C runtime is unreliable — `svGetArrayPtr` can return NULL for reg-backed values) and depends only on core DPI features, so it works on old Verilator too (verified on 5.032 and 5.049).

Implementation:
- **ForeignFunctions**: `DPIInst`, width mangling, `getDPIInstantiations` (scans the package for concrete uses), and a per-width `mkDPIDeclarations`.
- **Verilog/GenABin**: `VDPI` carries the `c_identifier` alias and the underlying C function name.
- **AVerilogUtil**: the conversion options carry the foreign-function map and def widths; call names are mangled by concrete width at all three emission sites (`vExpr`, `vForeignCall`, `vDefMpd`).
- **AVerilog**: feeds instantiations to `mkDPIDeclarations`; **DPIWrappers** generates the forwarding wrappers.

The second commit moves `import "DPI-C"` declarations from file scope into module bodies (a new `VMDPI` item): Verilator merges all files into one `$unit`, so file-scope imports collide ("Duplicate declaration of function") when a testbench and DUT both use the same foreign function. This fixes multi-module designs for both monomorphic and polymorphic imports and is inert for the non-DPI flow.

Testsuite integration for the DPI flow (deriving `-use-dpi` for verilator runs and making the foreign-function checks DPI-aware) follows in a separate testsuite-enablement PR; the harness currently exercises these tests via the VPI path, which this change leaves untouched.

## Testing

Built and verified standalone on a branch cut from `main`:
- The S0015 rejection is gone: `battery/ValueImports.bsv` (polymorphic `andN :: Bit#(n)`) compiles under `-verilog -use-dpi`, generating `dpi_wrapper_andN.c` with per-width shims (`andN_w32`/`_w64`/`_w96`) that forward to the single polymorphic C function.
- End to end under verilator: the battery `mkTestValues` design links and runs with `-use-dpi`, and its output matches the test's expected file exactly (modulo the `$finish` banner).
- Module-scope imports: every `import "DPI-C"` in the generated Verilog (DUT, TB, and values tests) appears inside a module body; zero file-scope imports.
- Non-DPI flow inert: the same design compiled without `-use-dpi` contains no DPI imports and keeps its VPI system tasks.
- VPI-path regression: `bsc.codegen/foreign` 196/0 and `bsc.codegen/foreign/battery` 52/0 with only the pre-existing iverilog exemptions as XFAILs — identical to upstream (this change touches no testsuite files).

Known paper cut: listing a generated `dpi_wrapper_*.c` explicitly on the `bsc` link command line fails (bsc pre-compiles explicit user C files without the simulator's `svdpi.h` include path); the intended flow — bsc auto-locates the wrapper and hands the source to verilator — works. A friendlier error or wrapper-aware handling could follow up.

---


quark17:
I haven't read this all yet, but the mentions of concrete widths and need to monomorphize sound wrong to me.  Polymorphic functions just need a wrapper, when implemented with DPI, and PR #439 gives an example of what the wrapper would need to look like.  Unless you've chosen to monomorphize for performance?

PR #439 added the initial DPI-based implementation for BDPI, by implementing all the forms that didn't need wrappers.  The only reason that polymorphic functions weren't supported is that a path for wrapper generation (like we do for VPI-based) hadn't been added.

FYI, the issue for this is #169 and there's an issue for Verilator support #170, that can possibly be closed -- though as I say there, we may want to leave an open issue for thinking through whether we can codegen a deeper integration with Verilator's clocking.  (Verilator is now supporting more timed Verilog, but it may be doing so by adding a second less-efficient codegen.  If we can generate clocking in a way that uses the efficient codegen and doesn't require the new timing approaching, then designs might run faster -- but I don't have a sense of how much difference there is.)

nanavati:
In principle there is a polymorphic rendering, but it does not work with Verilator (I tried, newer Verilator are getting closer but are not there) and Verilator support is the main point. Other simulators also support VPI so a polymorphic DPI rendering adds less to them.

I think a follow-on could be a knob to switch back to the polymorphic rendering (incompatible with Verilator linking) but I don't really want to make that knob because I don't know how to test it without a proprietary simulator.

nanavati:
Interaction note with #998 (same note posted there): this PR's module-scope `VMDPI` items land in `vm_body`, which #998's `vModuleDeclVIds` decl scan enumerates exhaustively-without-wildcard. Whichever merges second must add a `VMDPI` case to that scan (and to `instance Ord VMItem`) or `bsc -verilog -use-dpi` crashes on every BDPI design. Exact clauses and validation details in the #998 comment.

nanavati:
Refreshed: this branch is now stacked on #998 (as #998 is on #997), so it shows the cumulative commits. The commits belonging to this PR are the top three: the two DPI commits plus a small adaptation commit ("Verilog: cover VMDPI in the identifier scans and the VMItem ordering") that only exists because of the stacking on #998. Reviewing just those three gives this PR's own content.

PR #1000, opened 2026-07-03
Add -dump-formats: select waveform dump formats at link time (vcd/fst/fsdb/none)

nanavati:
This adds a link-time flag selecting which waveform dump formats a simulation binary supports, with a uniform runtime selector across simulators.

## The model

- **`-dump-formats <list>`** — consumed when linking; a comma-separated subset of `vcd`, `fst`, `fsdb`, or `none`. The default is `vcd`, so existing behavior is unchanged.
- **`+bscvcd` / `+bscfst` / `+bscfsdb`** — at run time, select which compiled-in format to dump (`main.v` already scaffolded these plusargs; the FST branch was stubbed out and FSDB was `ifdef`'d but never wired).

The format set is a link-time property because that is where the simulators bake it in: verilator compiles in exactly one trace class (`--trace` xor `--trace-fst`), and VCS enables FSDB at elaboration (`-kdb`). For iverilog (formats live in `vvp`) and Bluesim (dumping is the runtime `-V`), the flag records intent and gates the harness.

| | `none` | `vcd` | `fst` | `fsdb` | multiple |
|---|---|---|---|---|---|
| verilator | ✓ | ✓ | ✓ | ✗ | ✗ (at most one; hard verilator constraint) |
| VCS | ✓ | ✓ | ✗ | ✓ (`-kdb`; modern VCS bundles the FSDB writer) | ✓ `{vcd,fsdb}` |
| iverilog | ✓ | ✓ | ✓ (`vvp -fst`) | ✗ | ✓ `{vcd,fst}` |
| Bluesim | ✓ | ✓ | – | ✗ | – |

**No silent failures:** a format the simulator can never produce is a build-time error; a format that is valid but not compiled into *this* binary produces a loud runtime error (previously `+bscfst`/`+bscfsdb` were silent no-ops, and a Bluesim binary asked to dump wrote a headers-only VCD).

**`none` is the performance variant:** verilator builds without tracing entirely, and Bluesim omits the per-signal VCD machinery (~20% less generated C++ on a 1000-register module; `dump_VCD`/`dump_VCD_defs` remain as empty stubs since they are pure-virtual on `Module`).

## Commits

1. `-dump-formats` flag; `main.v` reworked from fsdb-xor-vcd to additive branches with errors for unbuilt formats; iverilog FST wired end to end.
2. VCS: `fsdb` via `-kdb -debug_access+all +define+BSC_FSDB`. *(Verified by construction only — no VCS installation was available; the `-debug_access` level may warrant adjustment.)*
3. Verilator: `fst` via `--trace-fst`, with `sim_main.cpp` selecting `VerilatedFstC`/`VerilatedVcdC` from `VM_TRACE_FST` and matching the plusarg to the built format.
4. Bluesim: `none` omits the per-signal VCD dump code.
5. `none` made honest on every backend (error instead of dumping or writing empty files).

## Testing

Built and verified standalone on a branch cut from `main`:
- **Flag**: `-dump-formats bogus` errors listing the valid values; `none` and `vcd,fst` parse correctly.
- **iverilog FST**: `+bscfst` with `vvp -fst` produces a binary `dump.fst` that `fst2vcd` round-trips.
- **verilator FST**: `-dump-formats fst` + `+bscfst` produces a valid `dump.fst`; the default (vcd) build still produces ASCII `dump.vcd`.
- **Errors, not silence**: an fst-built binary given `+bscvcd` reports "built for dump.fst"; a `none` build reports "no waveform dumping is available" on both verilator and iverilog (no file created); `-dump-formats vcd,fst` under verilator is a build-time error.
- **Bluesim**: `none` omits the per-signal `vcd_write_*` codegen (stubs remain) and `-V` reports the built-with-none error; the default `-V` flow still produces a normal VCD.
- Testsuite: `bsc.options` 70/0, `bsc.options/messages` 33/0, `bsc.verilog/vcd` 4/0, `bsc.bluesim/vcd` 26/0 (iverilog, Bluesim side enabled).

Known minor gap: on a Bluesim `none` build, `-V` prints the error but the kernel still creates the output file with a VCD header and clock toggles before the (stubbed) signal definitions run — unlike the iverilog `none` path, which creates no file. Fixing that requires a small kernel-side change and can follow up.

---


nanavati:
Refreshed: this branch is now stacked on #999 → #998 → #997, so it shows the cumulative commits. This PR's own content is the top five commits (the -dump-formats series); everything below is the PRs beneath it in the stack.

Issue #327, opened 2021-02-25
TriState code generation is broken


thoughtpolice:
```bluespec
import TriState::*;

interface T;
  interface Inout#(bit) v;
endinterface

(* synthesize *)
module t(T);
  let t <- mkTriState(False, 'b0);
  interface Inout v = t.io;
endmodule
```

`bsc -verilog -u T.bsv` gives:

```verilog
module t(CLK,
	RST_N,

	.v(t$IO));
  input  CLK;
  input  RST_N;

  inout  t$IO;

  // ports of submodule t
  wire t$IO;

  // submodule t
  TriState #(.width(32'd1)) t(.I(1'b0), .OE(1'd0), .O(), .IO(t$IO));
endmodule  // t
```

Looks like there's some simple bug that causes the names to be generated improperly.

thoughtpolice:
This is actually a general problem with all code generation invoving `inout` ports (i.e. I believe the `TriState.bs` module has a correct binding; it's actually a flaw in the code generator.) I'm seeing how easy a fix is...

quark17:
Can you explain what you think is wrong with that Verilog?

I notice that you've instantiated a submodule with instance name `t` that is the same name as the module `t`.  These are in different scopes, so I don't believe there is a conflict, but it's confusing as a reader.  I assume that's not related to your concern?

If you're referring to the dot notation in the port list, that is valid Verilog.  However, we have noticed that some tools do not like it.  In the `util` directory, there is a script called `basicinout.pl` which can be called with the Verilog file as argument.  It will overwrite that file with a new version, that changes code like this:
```
module ... ( ... .v(t$IO) ... );
   ...
   inout  t$IO;
   ...
   TriState ... t( ... .IO(t$IO) ... );
   ...
endmodule
```
into this:
```
module ... ( ... v ... );
   ...
   inout  v;
   ...
   TriState ... t( ... .IO(v) ... );
   ...
endmodule
```
The reason we use the first form is because there is no other way to express designs where multiple inout ports share the same signal.  In your case, and probably in most designs, there are no shared inout ports, and so the simpler form can be used.  (We might want to improve BSC to output this simpler form, when possible, so that there'd be no need for the script.  I don't recall if there's a reason why that's hard to do.)

thoughtpolice:
[...]

PR #1001, opened 2026-07-03
Verilog: render one-port-per-net inout ports plainly

nanavati:
## What

bsc currently binds **every** inout port of a generated module to its internal net with a port expression in the module header:

```verilog
module wrapBoth1(CLK, RST_N, .iioo(x$INOUT));   // ifc inout of a wrapped submodule
module mkImpArgConnect(CLK, RST_N, .arg(a$FOO)); // module-argument inout
```

This header form is exotic: some tools cannot parse it at all (Verilator rejects it outright with a syntax error), and it surfaces internal net names (`x$INOUT`) instead of the documented port names in waveforms.

When the port is the **only one on its net**, the expression serves no purpose. This PR adds a small pass over the final `VModule` (`renameInoutPorts`) that renames the net to the port name throughout the module and emits a plain port:

```verilog
module wrapBoth1(CLK, RST_N, iioo);
  inout [31:0] iioo;
  ...
  Both x(.CLK(CLK), .INOUT(iioo));
```

Truly *shorted* boundary inouts — two or more ports on one net, e.g. a module argument re-exported as an interface inout, or two boundary inouts joined by `mkConnection` — have no plain-port rendering and keep the expression form. The alternatives are no more portable: Verilator rejects port aliases, SystemVerilog `alias`, and the V1995 `tran` primitive alike, and iverilog rejects `alias` too, so the port-alias form bsc emits for genuine shorts is already the most portable choice.

## Why it is safe

* **Port names, widths, and directions are unchanged** — the external interface of every generated module is identical, and parent instantiations connect by the same names. The names adopted for the nets are exactly the documented, attribute-controlled port names (`(* port = "..." *)`, interface field names).
* Internal net names are not a documented interface; where they change, the net now carries the port's name (a strict improvement for waveform debugging, and one fewer `$`-name).
* The rename is capture-checked: a port name already in use elsewhere in the module leaves that port in expression form (this should not occur in generated code, but the pass must not capture).
* One-port-per-net inouts whose net already had the port's name (`.arg(arg)`) simply drop the expression form.

## Testsuite

* `testsuite/bsc.verilog/filter/basicinout.pl` — this `-verilog-filter` demo fixture is a user-space rewrite of exactly the rendering this PR fixes. It now passes already-plain inout ports through instead of dying, which also makes it **idempotent**: the `"basicinout Bug"` double-application XFAIL in `filter.exp` (whose own comment says "this should not fail; basicinout should be able to run twice") becomes a plain pass.
* Six golden `.v` files with one-port-per-net inouts are regenerated. The diffs are pure renames; notably, the old filter renamed the header and instance connections but missed the `wire` declaration — the compiler rename covers all of them.

Verified on this branch: `bsc.verilog/inout` and `bsc.verilog/filter` pass under iverilog (171 PASS / 0 FAIL and 10 PASS / 0 FAIL), and full-suite runs show no other golden or behavioral changes (the only generated-Verilog changes in the suite are the six goldens above). Under Verilator, previously-unparseable wrapper designs (e.g. `TbBoth`'s three-driver tristate bus) now build and simulate.


quark17:
This is issue #327.  There's a workaround using `-verilog-filter` but BSC should just output the cleaner code to begin with, so this is great, thanks.

I'm not sure about the changes in `bsc.verilog/filter/`.  That directory should be testing the `-verilog-filter` flag.  If the inout Perl script is no longer doing anything, then it and its tests should be deleted and replaced with tests and a script that actually does something -- even just simple renaming of identifiers, say?

testsuite/bsc.verilog/inout/mkFourInoutBuses.v.expected
```diff
@@ -34,21 +34,21 @@ module mkFourInoutBuses(CLK,
 			RST_N,
 			.a(a),
 			.b(a),
-			.c(o$FOO),
```
quark17, line 37: It's interesting that this case was using a submodule-based name `o$FOO` before -- even with the dot notation, it could have used `c` for the signal name.

nanavati: Right — the internal net simply kept whatever name it already had: for an interface inout fed by a submodule port, that is the instance wire (`o$FOO`), and the header expression bound the port name to it verbatim. There was never a pass that preferred the port name. That is exactly what `renameInoutPorts` adds: whenever the net carries only that one port, the net is renamed to the port name throughout the module, which is why this golden line changed from `.c(o$FOO)` to a plain `c`.

nanavati:
Thanks — good to know this is #327; the timing worked out nicely.

You are right about `bsc.verilog/filter/` — with the compiler doing the rename itself, `basicinout.pl` had nothing meaningful left to do (it only still fired on true shorts, where its rewrite produces broken Verilog anyway). I have replaced it as you suggested: the directory now tests `-verilog-filter` with a simple identifier renamer.

- `renamefire.pl` shortens the `-keep-fires` rule-signal prefixes (`CAN_FIRE_` → `CF_`, `WILL_FIRE_` → `WF_`) in the generated `.v`, using the same call convention (file argument, rewritten in place).
- The inout test designs (`ImpArgConnect*`, `RedoInoutConnect`, `InoutStub`) and their goldens are deleted, replaced by a small one-rule design compiled with `-keep-fires`.
- The flag coverage is unchanged: a single filter with a golden compare, the same filter applied twice (idempotent — the old "basicinout Bug" double-application case, now genuinely passing), a perl + sed filter chain (`simple.sed` still renames `CLK` → `CLOCK`), and a failing filter (missing sed script) failing the compile.

Verified green under both iverilog and verilator.

testsuite/bsc.verilog/inout/mkFourInoutBuses.v.expected
quark17: What's interesting about it is that `.a(a)` and `.d(d)` _did_ use the port name, and not, say, `.a(k$BAR)` or `.d(g$ARG)`.  So something _was_ able to pick a port name over a submodule-based name, when the signal connects to multiple ports, but wasn't doing so for signals that only connected to one port.

src/comp/AVerilog.hs
```diff
@@ -55,10 +56,65 @@ import qualified GraphWrapper as G
 -- Flags are the compiler flags
 -- [PProp] are the Pragmas
 -- XXX this function is too big
+-- Rename one-port-per-net inout ports to plain named ports.  bsc binds
```
quark17: Put `renameInoutPorts` further down in the file, where there are already functions under a comment header "Top-level inout handing ".

Here it's sitting between `aVerilog` and the header comments for `aVerilog`, making it look like those comments are attached to `renameInoutPorts`.  The convention is that the entry point (that shares the name with the file) goes first, so `renameInoutPorts` should later.  There's already a section for Inout functions, so put it there.

testsuite/bsc.verilog/filter/filter.exp
```diff
@@ -8,29 +8,27 @@ proc make_sed_cmd { cmd_file out_file_ext } {
...
-compile_verilog_pass ImpArgConnect.bsv ""  [list -verilog-filter $filter1]
-compare_verilog mkImpArgConnect.v
-compare_verilog mkArgImpConnect.v
+# the filter only renames identifiers it introduces no match for, so
+# applying it twice gives the same result
+erase RenameTest.bo
+compile_verilog_pass RenameTest.bsv "" [list -keep-fires -verilog-filter $filter1 -verilog-filter $filter1]
+compare_verilog mkRenameTest.v mkRenameTest.v.renamed.expected
 
+# multiple scripts compose in order (the sed renames CLK to CLOCK)
```
quark17, line 27: This tests that both scripts applied, but doesn't test that they are applied in the same order as on the command line.

We weren't testing that before, so you've not made things worse.  But thought I'd note it, in case you think it's worth adding while we're here.  (For that, we'd need, say, a first script to change `WILL_FIRE_` to `WF_` and a second script to change `WF_` to `W_F_`.)

testsuite/bsc.verilog/inout/mkFourInoutBuses.v.expected
nanavati: I had Claude dig into this — it turns out nothing was picking port names deliberately; it's an artifact of connection order. The net representative comes from `addInoutPair` in AVerilog.hs (the default `-remove-inout-connect` path), which folds the InoutConnect instances into a union-find where: a new class is named after the *first argument* of the connection that creates it, a merge keeps the *first argument's* representative, and the fold processes connections in reverse elaboration order.

In this test, `d` wins its net because `mkConnection(d,f)` seeds the class; `a` wins because the merging connection `mkConnection(a,k.foo)` has `a` first; and `c` loses because the last connection on its net is `mkConnection(o.foo, c)` — the submodule wire is the first argument, so `o$FOO` names the class (which is also why it beat `n$FOO` and `p$FOO`). Written as `mkConnection(c, o.foo)`, the net would have been named `c`.

So with different argument order these headers could just as well have read `.a(k$FOO)` — the port-named cases were luck, not policy.

src/comp/AVerilog.hs
nanavati: Done

testsuite/bsc.verilog/filter/filter.exp
nanavati: Test added.

quark17 (dismissed):

quark17 (approved):

PR #1003, opened 2026-07-04
INSTALL.md: describe GHC support by the CI matrix

nanavati:
The claim that BSC builds with any GHC from 7.10.3 through 9.12 is stale: the code already requires at least GHC 9.4 (e.g. BinData.hs imports GHC.IsList, base >= 4.17), and nothing older than the CI matrix (9.6.7 primary, plus 9.8/9.10/9.12/9.14) is actually tested. Describe the supported versions by pointing at the CI matrix instead of maintaining an aspirational range.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -52,12 +52,12 @@ symlinks in `/usr/bin/` that point to the executables in
 ## Requirements
 
 To build a complete release of BSC, you will need:
- - The standard Haskell compiler [GHC]. The recommended version is the
-   latest 9.6 point release, which is the primary version built and
-   tested by this project's CI; newer releases through 9.14 are also
-   tested (see the [CI workflow] for the exact set).  Older
-   releases are untested and may not work.  We recommend installing
-   GHC via the popular installer [GHCup].
+ - The standard Haskell compiler [GHC]. The recommended version is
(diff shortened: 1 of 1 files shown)


quark17:
Sounds reasonable.  I would just make `.github/workflows/ci.yml` a link to the file, though.

nanavati:
> Sounds reasonable. I would just make `.github/workflows/ci.yml` a link to the file, though.

Done

kenta2:
My suggested more aggressive edit:

> The recommended version is 9.6.7, which is the primary version built and tested by this project's continuous integration (CI); newer versions are also tested (see the [CI workflow] for the exact set).  Older versions are untested and may not work.

* explicitly include the string "9.6.7" to make find and replace updates easier when `ci.yaml` is updated.
* omit the newest ghc version tested to prevent the document from become stale when `ci.yaml` is updated.
* consistently use the term "version" instead of "release".

nanavati:
Adopted your suggested wording verbatim in fd65646 — literal "9.6.7" for greppability, newest-tested version omitted so the text can't go stale, and "version" used consistently. Thanks!

kenta2 (approved):

PR #1004, opened 2026-07-04
CI: run the Verilog testsuite with Verilator (non-blocking)

nanavati:
## What

Adds a **Verilator testsuite job** to the release build/test matrix on every platform (ubuntu and macos):

```
make -j ${CORES} -C testsuite checkparallel TEST_BSC_VERILOG_SIM=verilator CTEST=0 SYSTEMCTEST=0
```

- **Verilog-only** (`CTEST=0 SYSTEMCTEST=0`): the Bluesim and SystemC backends are already covered by the main testsuite job, so this job swaps only the Verilog simulator (iverilog → Verilator) without duplicating the rest.
- **Non-blocking** (`continue-on-error: true`): failures report Verilator compatibility without gating merges.
- Runs only in the release matrix (`verilator_testsuite: true`); the GHC-version matrices skip it, since Verilator results don't vary by GHC version.
- Own ccache key per platform (Verilator's generated C++ dominates the job's cost; the runtime objects are identical across tests, so ccache carries most of the weight).
- Verilator comes from the distribution (apt/brew), which deliberately brackets the version range: **ubuntu-22.04 ships Verilator 4.038** (oldest, a compatibility canary), ubuntu-24.04+ ship 5.x, and **Homebrew tracks the current release** (newest).

## Why non-blocking, and when it turns green

With the pending Verilator enablement PRs applied (#990–#1001 and two follow-ups), the full testsuite runs clean under Verilator 5.x locally — `checkparallel` verilog-only with **zero unexpected failures** (verified on 5.032; spot-verified 5.049). This job is the visibility layer for keeping it that way; it can be promoted to a required check once the enablement queue lands and the job is green on the supported floor (Verilator ≥ 5).

The 22.04 job will stay red-ish even then: Verilator 4.038 predates several upstream Verilator fixes the suite relies on (fatal-instead-of-warning `$readmem` short-file handling, escaped-identifier lexing, `BLKANDNBLK` support, `$fclose` on concatenations, `VL_SHIFTL_WWQ`, among others — a full census exists). Its job is to answer "what still works on old Verilator" (the compiler flow, DPI including the polymorphic path, and tracing all do), not to be green.

## Notes

- The 360-minute timeout is a deliberate first-run ceiling; runner wall-time with a cold ccache is untested, and the cap can be tightened once real numbers exist.
- `CTEST=0` runs safely since #992 (backend-independent dump checks no longer gated on ctest).

## Update: rebuilt on the enablement queue

This PR now carries the full enablement queue so the Verilator job can demonstrate the result directly in CI:

- The unapproved enablement PRs (#995, #996, #997, #999, #1000, #1001) were rebased onto current main and force-pushed; #998 was already current. The approved ones (#990, #991, #993, #994) were left untouched to preserve their reviews — their commits are included here by cherry-pick.
- On top: the two CI commits, a commit addressing the review comment (Verilator is now installed only in the jobs that test with it, as a second step after the shared dependency script, on both Ubuntu and macOS), and one integration commit adjusting pattern arities where the SV-identifier scan (#998) meets the widened `VDPI` (#999) and `VAParameter` (#997) constructors.
- As the enablement PRs merge, their commits fall out of this diff; what remains at the end is the CI change itself.


.github/workflows/install_dependencies_testsuite_macos.sh
```diff
@@ -8,4 +8,5 @@ brew install \
   ccache \
   deja-gnu \
   icarus-verilog \
+  verilator \
```
quark17: This installs verilator even in runs that don't need it.  I guess that's fine -- eventually we may use Verilator everywhere.  But if you care, an alternative would be to put the command for installing verilator only into the job for testing with verilator (as a second line after running the script, in the install step).

nanavati: As we use more SV constructs iverilog doesn't support, I suspect we will want verilator available by default even if we don't use it for a full testsuite run.

nanavati: Done — Verilator is out of the shared dependency scripts and installed only in the Verilator jobs themselves (second step after the shared script, both Ubuntu and macOS). The PR is also now stacked on the full enablement queue (rebased onto current main) so this job can demonstrate the green run directly; details in the updated description.

PR #1006, opened 2026-07-05
testsuite: fix parallel-make race that silently drops long tests

nanavati:
The `checkparallel` and `fullparallel` targets list their setup target and `run-tests-setup` as sibling prerequisites, which a parallel make (`-j`) runs concurrently. `run-tests-setup` generates the test list with `find . -name '*.exp'`, while `fullparallel-setup`'s `make enablelongtests` is what *creates* the `.exp` files under `bsc.long_tests` — so whether the long tests are included in a `fullparallel` run depends on which prerequisite wins the race.

Observed concretely: two identical back-to-back `make -j128 fullparallel` invocations on the same tree, where one ran 5 fewer test directories than the other (`bsc.long_tests/MPEG4`, `conflict_free_large`, and the three `log2_loop` variants were silently absent from the first run's `ALL_TESTS`, and no failure was reported).

Fix: invoke `run-tests-setup` from the recipe body, after the setup prerequisite has completed. Same structure applied to `checkparallel` (its `make clean` has the analogous ordering exposure against the list generation).

## Testing

`make -n fullparallel` / `make -n checkparallel` show the serialized recipe; a full `make -j128 fullparallel` on the fixed tree generated `all_tests.mk` containing all five long-test directories deterministically.


quark17:
Looks good.

If you feel like doing some cleanup, I noticed that `run-tests-setup` lacks a `PHONY` declaration; and I'd move the `PHONY` declarations for `check` and `localcheck` to be immediately above the targets, not so far away, and doublecheck that all targets are properly declared `PHONY` right above, if needed.

nanavati:
Done — `run-tests-setup` is now declared `PHONY`, and the `check`/`localcheck` declarations moved to sit immediately above their targets. Audited the rest: every target in `suitemake.mk` now has its declaration right above it, and `parallel.mk` already followed the convention.

quark17 (approved):

Issue #659, opened 2023-12-30
Improve the mechanism for testing Bluespec Inc bug 1490


quark17:
With GHC 9.8.1, the following failures occur in `bsc.bugs/bluespec_inc/b1490/`:
```
FAIL: module `' in `Bug1490MyUnion.bsv' should compile to Verilog
FAIL: module `' in `VsortWorkaround.bsv' should compile to Verilog
```
This directory is testing that certain examples will compile within a 256M heap.  The two failures are because, with GHC 9.8.1, they now exceed the heap and abort.  Presumably there has been some change in GHC 9.8.1 that now takes more heap.  Unfortunately, I don't know how to debug this, to decide if this is a significant regression in GHC, or what.

I guess I can test whether it compiles with slightly more heap.  I do see that `MyUnion` compiles with 260M and `VSortWorkaround` with 265M.  So maybe the new GHC just uses slightly more heap and we just need to increase the limit a little (at least for newer GHC versions) -- and not that there's a significant change in the order of growth, say.  However, note that the tests have a parameter that controls the size: the `Bool` test, for example, defines `Ports`, which can be increased to scale up the test (as been increased, from 16 to 21, in the past).  So someone has constructed these as a test of BSC's scaling/optimization; but it's unfortunately also tied to the performance of GHC (which changes from version to version).  It would be nice if there was a better way to write these tests, or some information on what to look for when adjusting them.

However, another measurement we can do is, with earlier GHC, see how low the heap max can be decreased before the test fails.  Increasing the heap from 256 to 265 for GHC 9.8.1 is "insignificant" only if we assume that the earlier GHC was using nearly 256.  But that doesn't seem to be the case!  With BSC built with older GHC (I think 9.4.8 or 9.4.7), I can drop the `-M` flag down to the minimum (`4K`) without any change in the testsuite results!  What's going on there?!  With newer GHC, I can also decrease the flag to `4K` and still only those two tests fail.  Am I misunderstanding how the flag works, or are those tests really using less than 4K heap?  (This was on macOS 11.  I also tested on an old Debian, with GHC 8.8.4, and lowering to 4K also worked fine there.)

The fact that two tests now need more than 256M which used to work fine with 4K seem like a significant change!  If that's true, is there some way to report this example to the GHC developers?

[...]

PR #1007, opened 2026-07-05
evaluator: make the PrimBNot push O(cells) over shared conditionals

nanavati:
Evaluating a dynamic `PrimBNot` pushes the negation through `PrimIf` structure so constant leaves can fold. The push recursed into both branches of every `PrimIf` with **no memoization** — its definition site carries the comment `XXX When the same heap ref appears twice, this does twice the work!` — and its leaf handler re-heaped a fresh application per visit. A negation over a *shared* conditional DAG therefore performed work (and allocated heap) proportional to the number of root-to-leaf **paths** — exponential in the DAG's depth.

Conditional-update pipelines build exactly such DAGs: K stages of conditional swaps over an N-element vector deepen the DAG by two if-levels per stage.

## What this fixes

- **`VsortOriginal.bsv` (bug 1490)** has been documented as exhausting a 256M heap since the initial open-source release, and #659 records it still exhausting 400M and churning indefinitely at 500M. With this change it compiles in ~1 second within the standard limit, so its expected-failure entry is flipped to a positive test.
- The new **`BNotShared.bsv`** test is a starker, fully self-contained case: 16 registers × 64 odd-even conditional-swap stages elaborate to only 4,792 defs, but an exact count over the def graph shows the conditional DAG contains **≈2^129 then/else paths**. A per-path push cannot terminate (at 10^10 visits/second, ~10^21 years); the memoized push compiles the file in ~4 seconds with default flags. Trace counts confirm the fix is linear: `primBNot` conAp entries fit `441·K + 836` exactly across K = 8…64.

## The change

The push (new function `pushBNot`) now:
- memoizes its result **per heap cell** (`heapBNots` in the evaluator state, following the existing `extractWires` wire-set cache);
- heaps intermediate results, so pushed output remains a DAG rather than an inline tree;
- guards against self-referential structure with a visited set;
- starts from the WHNF argument (not the heaped form), so re-dispatch through the evaluator can still collapse all-equal dynamic selects — their residual forms are returned inline, never frozen into cells — preserving the stock behavior exercised by `bsc.arrays/dynamic/PrimBNot_DynArrSelect.bsv`;
- delivers constants bare with the cell predicate lifted, and merges branch predicates via `pIf`, matching the previous `evalStaticOp` behavior for implicit conditions.

## Testing

Full testsuite on this branch (this commit on top of `main`): **18,358 PASS / 0 FAIL / 129 XFAIL / 0 XPASS** — no golden or expectation churn beyond the intentional `b1490.exp` flip. The same commit was also validated on a development branch with an independent full run (18,432 / 0 / 130).

Fixes the residual question in #659.

### DIFF
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -3863,13 +3863,13 @@ conAp' _ (ICPrim _ op) fe@(ICon prim_id _) as | strictPrim op = do
          else
             case (op, as') of
             (PrimBNot, [E e]) | isDyn e ->
-                -- The iTransExpr catch-all will handle PrimIf but not arrays
-                let handler e' =
-                        case (doPrimOp bestPosition op [] [e']) of
-                          Just (Right e_res) -> return (pExpr e_res)
-                          Just (Left errmsg) -> errG (bestPosition, errmsg)
-                          Nothing -> evalAp "Prim PrimBNot" fe [E e']
-                in  addPredG p $ evalStaticOp e itBit1 handler
+                -- Push the negation through conditional structure with
+                -- per-heap-cell memoization (see pushBNot): pushing once
+                -- per PATH instead of once per cell is exponential in the
+                -- depth of a shared conditional DAG.  The WHNF arg is used
+                -- (not the heaped form) so that re-dispatch through the
+                -- evaluator can collapse all-equal dynamic selects.
+                addPredG p $ pushBNot bestPosition fe e
             -- name primitives
             (PrimJoinNames, [E (ICon _ (ICName { iName = n1 })),
                              E (ICon _ (ICName { iName = n2 }))]) ->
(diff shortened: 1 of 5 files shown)


src/comp/IExpand.hs
```diff
@@ -3112,6 +3112,113 @@ bldApUH' s f as = do
         when doConAp $ traceM ("bldApUH' " ++ s ++ " " ++ ppReadable e)
         return (P pTrue e)
 
+-- Push PrimBNot through a (possibly shared) conditional DAG once per
```
quark17: I would move this elsewhere in the file, rather than among the conAp/bldAp functions.  Below, we start to get into functions like "doArray*", "doIf"/"improveIf*", "doAnd"/"doOr", and "doIs"/"doSel".  I'd maybe put it after "doOr" (before the delimiter for "doIs")?

quark17 (commented):
Awesome!  Great to have issue b1490 diagnosed and fixed!

Looks good, aside from moving the code.  With the caveat that I haven't thought too deeply about the change.  Is there something that needs to be fixed in `evalStaticOp` generally? or is it just that BNot is special (being unary and common and appearing in these kinds of constructs)?

src/comp/IExpand.hs
nanavati: Moved — it now sits after `doOr`, just before the `doIs` delimiter.

nanavati:
Good question — the exposure is general in principle, but BNot is where
it detonates in practice.

The general part: `evalStaticOp'`'s `PrimIf` case distributes *any*
handler into both branches with no per-cell memoization (the
"XXX When the same heap ref appears twice, this does twice the work!"
comment sits on that shared machinery), so any static op pushed through
a shared conditional DAG does O(paths) work rather than O(cells).

Three things made BNot the one that actually explodes:

1. Its special-case arm not only recursed per path, it **rebuilt the
   leaf residual from the unheaped expression at every visit** (the
   handler's fallback was `evalAp` on the WHNF), so each of the 2^k
   paths also allocated fresh heap cells for the same leaf — the heap
   exhaustion in bug 1490 is that allocation, not just the walk.
2. Boolean negation is ubiquitous in exactly the constructs that build
   deep *shared* conditional structure: comparator/guard idioms like
   `!v[j-1] && v[j]` inside iterative conditional updates, where each
   stage's conditions reference the previous stage's muxes. Depth grows
   per stage, so the DAG gets deep in a way type-shaped structures
   don't.
3. The other common demands routed through `evalStaticOp` (doSel/doIs
   on decode trees) hit structures whose depth is bounded by a type's
   nesting rather than accumulated per pipeline stage, and their
   branch results re-merge (the ICTuple/ICCon cases of improveIf) as
   they return, which keeps the effective structure shallow.

So I'd describe a general per-(op, cell) memoization inside
`evalStaticOp'` as plausible follow-up work rather than something this
PR should attempt: the handlers it serves are heterogeneous (predicate
lifting, undet handling, the doBK bookkeeping variants, array element
distribution), so a shared cache needs per-handler care. This PR takes
the narrow fix for the demonstrated detonation. Happy to file an issue
for the general version, citing that XXX comment.

One empirical data point in support of "general in principle, BNot in
practice": on a development branch where deep shared conditional DAGs
are considerably more common (#1008 keeps values symbolic across
pipeline stages), the full testsuite plus the pathological workloads
(200-element conditional-update loops, 32- and 64-wide odd-even sort
networks, the PAClib DFT pipelines) surface no other op with this
signature.

quark17 (approved):

PR #1010, opened 2026-07-05
Improve wording about installing profiling libraries from the distro

kenta2:
package manager.


INSTALL.md
quark17: Note that the same wording is here, which I assume you'll also want to change

kenta2: I have force pushed a combined commit that updates both Debian and Fedora.

quark17 (approved):

Issue #963, opened 2026-06-02
Build failure on Ubuntu 22.04 / Ubuntu GHC 8.8.4


tmarkettos:
A clean Ubuntu 22.04 container, with the minimal dependencies installed, including Ubuntu's GHC 8.8.4.  The build fails with

```
[126 of 227] Compiling ISyntaxSubst     ( ISyntaxSubst.hs, /home/me/cheri/bsc/src/comp/../../build/comp/ISyntaxSubst.o )

ISyntaxSubst.hs:58:1: error:
    • Non type-variable argument
        in the constraint: SubstContext ctx (IExpr a)
      (Use FlexibleContexts to permit this)
    • In the type synonym declaration for ‘ExprSubstCtx’
   |
58 | type ExprSubstCtx ctx a = SubstContext ctx (IExpr a)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

ISyntaxSubst.hs:59:1: error:
    • Non type-variable argument
        in the constraint: SubstContext ctx IType
      (Use FlexibleContexts to permit this)
    • In the type synonym declaration for ‘TypeSubstCtx’
   |
59 | type TypeSubstCtx ctx = SubstContext ctx IType
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
```
[145 of 227] Compiling BinData          ( BinData.hs, /home/me/cheri/bsc/src/comp/../../build/comp/BinData.o )

BinData.hs:61:1: error:
    Could not find module ‘GHC.IsList’
    Perhaps you meant
      GHC.List (from base-4.13.0.0)
      GHC.OldList (from base-4.13.0.0)
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
61 | import GHC.IsList(IsList(..))
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
```
[165 of 227] Compiling RSchedule        ( RSchedule.hs, /home/me/cheri/bsc/src/comp/../../build/comp/RSchedule.o )

RSchedule.hs:51:10: error:
    • Illegal instance declaration for ‘PPrint RAT’
        (All instance types must be of the form (T t1 ... tn)
         where T is not a synonym.
         Use TypeSynonymInstances if you want to disable this.)
    • In the instance declaration for ‘PPrint RAT’
   |
[...]

PR #1012, opened 2026-07-05
doc: Caveat of older ghc in system packages (#963)

kenta2:


INSTALL.md
quark17: The same wording is down here, that I assume you would also want to update.

kenta2: I have forced pushed to my branch the update also to the Fedora section.

quark17 (approved):

PR #1021, opened 2026-07-05
Verilog: confine positional instance parameters to Classic foreign calls

nanavati:
Follow-up to the discussion on #997 about whether the `Either` in `vi_inst_params` could become named-only.

Analysis of all construction sites: two of the three are the `-v95` fallback (they disappear if #719 relands), and the third — the noinline-function instantiation path in `AVerilogUtil` — was positional unconditionally. That site serves two populations:

- **BSV `(* noinline *)` functions** are enforced monomorphic (T0111), so their `ANoInlineFun` never carries numeric type arguments and no parameters are emitted at all. This PR routes that always-empty case through the named side.
- **Classic `foreign` functions applied at numeric types** (in-tree example: `Fork` — `foreign vfork :: Bit n -> Bit m = "Fork",("i","o")`) pass their `ITNum` type arguments as instance parameters, and these are positional **by necessity**: the declaration syntax names only ports, so the compiler never learns the hand-written module's parameter names (`Fork.v`'s `iw`/`ow`); the values pass in type-argument order matching the parameter declaration order.

So after #719, the residual `Left` is exactly the Classic-foreign contract, now documented at the emission site and on the `ANoInlineFun` type. Collapsing the `Either` entirely would require extending the Classic foreign declaration syntax to name parameters (or retiring numeric-polymorphic foreign) — likely not worth it for one legacy feature.

No change to generated output: empty parameter lists render identically on either side (verified: `bsc.verilog/noinline` 62 PASS, `bsc.lib/fork` 5 PASS, goldens unchanged).

Related: #997, #719, #400.

---
**Update:** superseded by #1022, which collapses the `Either` entirely (the parameter names turn out to exist: the declaration's type variables). If #1022 is preferred, this can close; if this lands first, #1022 rebases trivially.

### DIFF
--- a/src/comp/ASyntax.hs
+++ b/src/comp/ASyntax.hs
@@ -1243,7 +1243,12 @@ data ANoInlineFun =
     ANoInlineFun
          -- function nam
          String
-         -- numeric types
+         -- numeric types (the ITNum type arguments of the call).
+         -- Only Classic foreign functions can have these; BSV noinline
+         -- functions are enforced monomorphic (T0111).  They are
+         -- emitted as positional Verilog instance parameters in
+         -- type-argument order, because the foreign declaration gives
+         -- no parameter names (see vDefMpd in AVerilogUtil).
          [Integer]
          -- port list (inputs,outputs), each is port name and size
          -- XXX sizes all seem to be generated as 0.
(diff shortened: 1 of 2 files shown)


nanavati:
Closing in favor of #1022 which is a more complete solution.

PR #1022, opened 2026-07-06
Verilog: pass instance parameters by name everywhere; remove -v95

nanavati:
Instance parameters are now always passed by name; the `Either` in `vi_inst_params` is gone, along with the `-v95` flag that was its main client. Follows the #997 discussion; supersedes #1021 (if this is preferred, #1021 can close — if #1021 lands first, this rebases trivially).

## The insight

The one non-`-v95` client of positional parameters was the Classic foreign-function-bound-to-a-module flavor (in-tree: `Fork.bs`), where the hand-written module's parameter names were "unknowable." But the declaration *does* name each parameter slot: the type variable. So the convention becomes: **the module's parameter names are the declaration's type variable names**, and bsc emits

```verilog
Fork #(.iw(32), .ow(64)) instance_vfork_0(.i(...), .o(...));
```

`Fork.bs` conforms to the existing `Fork.v` by renaming its own type variables (`foreign vfork :: Bit iw -> Bit ow = "Fork",("i","o")`) — no Verilog changes. This is a strict upgrade in failure mode: the old positional contract (type-argument order must match the `.v` parameter declaration order) failed *silently* as a miswired netlist; a name mismatch fails *loudly* at Verilog elaboration.

## What's in it

- **Name threading**: `MakeSymTab` captures the declaration's type variable names in quantification order (the same order the type arguments arrive at applications) into `VarForg`; they ride `ICForeign.fTyVarNames` and pair with the concrete values in `AConv`, so `ANoInlineFun` now carries `[(String, Integer)]`.
- **Width discipline (new error T0159)**: for a foreign function bound to a module, every bit width must be a bare type variable or a literal. A type-function width (`Bit (TAdd n 1)`) has no name to pass and would force the `.v` to reimplement bsc's type arithmetic (with its exact ceiling/rounding semantics) to size its ports — rejected, with guidance to give the derived width its own variable. "True" foreign functions (no port list) are unaffected, as are BSV `(* noinline *)` functions (enforced monomorphic by T0111, hence parameterless).
- **`Either` collapse**: `vi_inst_params :: [(VId, Maybe VExpr)]`, matching how ports have always worked; printer simplified; `pv95params` deleted.
- **`-v95` removal**: the flag's only remaining substance was positional parameters plus *dropping* `$signed`/`$unsigned` (replacing the cast with a comment — lossy, not compatibility). It also hasn't delivered V95-clean output in years: the shipped primitive library uses `localparam` (SizedFIFO, Probe) and named parameter instantiation (SyncFIFOLevel) ungated. Relands the flag-removal intent of #719/#400 (cc @thoughtpolice); `$signed` is now always emitted. User guide updated.
- **Format bumps**: `.bo` (`ICForeign` gains the names field) and `.ba` (`ANoInlineFun` shape) headers bumped to `20260706-1`.

## Migration

For out-of-tree Classic code using `foreign ... = "mod",(ports)` with numeric type arguments (believed to be nearly nil): name the declaration's type variables after the module's parameter names (or vice versa). Everything else is unaffected.

## Testing

- `bsc.lib/fork`: passes, plus a new `find_regexp` pinning the named emission (`.iw(...)`, `.ow(...)`) in the generated Verilog.
- `bsc.verilog/noinline`: passes, plus a new negative test (`BadForeignWidth.bs` → T0159).
- `bsc.verilog/v95` → renamed `bsc.verilog/verilog_params`; keeps testing parameter ordering and `$signed` output without the flag.
- `bsc.options`: help/print-flags goldens updated.
- Full testsuite: **18,345 PASS / 0 unexpected FAIL / 129 XFAIL / 0 XPASS** (fullparallel). Correction: the long tests were silently dropped from this particular run by the parallel-make race that #1006 fixes. A later run on this branch's continuation (#1023) with long tests enabled shows the only long-test failures are pre-existing on stock main and unrelated to these changes (the MPEG4 output diff that #995 fixes, and a stack overflow in `log2_loop`, cf. #957) — verified by reproducing both on an unmodified main-based build.

Related: #997, #1021, #719, #400.


matx-amy:
reviewed, :+1:

nanavati:
Rebased onto current main (941eecfe) and force-pushed (cf7dc26f). What the rebase resolved: Re-applied named-parameter emission across upstream's port-splitting part 2: foreign-call `VMInst` port construction is now grouped per argument (`iports ++ oports`, instance list form) — plus the `GenABin` header refactor (`headerBS`).

PR #1023, opened 2026-07-06
Permit numeric provisos on foreign functions, checked then erased

nanavati:
Follow-up to #1022 (stacked on its branch; only the last commit is new — the rest is #1022 and merges out). Completes the modernization of module-bound Classic foreign functions: with #1022 the widths pass as *named* parameters; with this PR the *relationships between* the widths are finally checkable.

## Classic foreign: numeric provisos, checked then erased

`Fork.bs` has carried this comment since the beginning: `XXX m == 2*n, but we can't have contexts on foreign.` The declared type was more general than the implementation, and applying `vfork` at inconsistent widths typechecked anyway — the wrongness surfaced only as a miswired netlist. Now:

```haskell
foreign vfork :: (Add iw iw ow) => Bit iw -> Bit ow = "Fork",("i","o")
```

The typechecker enforces the proviso at every application (a violating use fails with T0031: `no instances of the form Add#(8, 8, 15)`), and the proviso is then **erased**: the numeric classes have no methods, so their dictionaries carry no content, and `IConv` drops the dictionary arrows from the foreign's type and the dictionary arguments from its applications — mirroring the existing `dropDicts` treatment of primitives. Nothing downstream changes: the type arguments still pass as (named, per #1022) instance parameters, and the generated Verilog is identical.

Only the numeric classes (`Add`, `Mul`, `Div`, `Log`, `Max`, `Min`, `NumEq`) are permitted — anything else (e.g. `Bits`) would need real dictionary content or coercion insertion, which a foreign body cannot provide; rejected with a new error, **T0160**. Noinline-created foreign functions are exempt (their `WrapField` proviso is checked in typecheck, exactly as before), and the erasure only applies when *every* proviso is numeric, so their conversion is unchanged.

This also pairs with #1022's bare-width rule: widths stay bare variables in the type, and the *relationships* between them live in provisos — checked statically, invisible to the Verilog.

## import "BDPI": already works — now pinned by tests

The same question for BDPI turns out to need **no compiler change**: numeric provisos on `import "BDPI"` declarations already flow through the standard signature machinery — enforcement at applications verified (T0031), and both Bluesim and Verilog-VPI codegen handle the polymorphic import (DPI-mode polymorphism is #999's scope — S0015 today). The C side receives the sizes it needs as explicit arguments, per the usual polymorphic-BDPI convention. New tests pin this with a worked example:

```bsv
import "BDPI" function Bit#(m) bit_dup (Bit#(n) x, Bit#(32) nsz)
   provisos(Add#(n, n, m));
```

with a C implementation, run end-to-end under both backends with checked output, plus a violated-application negative test.

## Testing

- `bsc.verilog/noinline`: positive Classic case (including a `find_regexp` that the named parameters still emit), violated application → T0031, non-numeric proviso → T0160.
- `bsc.codegen/foreign`: the BDPI example above (Bluesim + Verilog runs) and its violation pin. Directory: 207 PASS.
- `bsc.lib/fork`: Fork.bs now exercises the proviso in every use.
- Full testsuite (fullparallel, long tests enabled): **18,397 PASS / 129 XFAIL / 0 XPASS**, with 6 FAILs in two long-test directories that are **pre-existing on stock main** and unrelated to this change (verified by reproducing both on an unmodified main-based build): the MPEG4 output diff that #995 fixes, and a `bsc: Stack space overflow` in `bsc.long_tests/log2_loop/actionvalue` (the territory of #957).

Related: #1022, #997, #999.


matx-amy:
reviewed; it looks like this is missing the hunk:

https://github.com/B-Lang-org/bsc/pull/1022/changes#diff-87430d9b698d4b9082f008d87af33dcaed839250ace9c82b6cae264c0e65cf66R385-R391

presumably just a rebase artifact?

otherwise, :+1: 

---

**EDIT** oh, it's just missing https://github.com/B-Lang-org/bsc/pull/1022/changes/3098e479070133b266a5fa07c5a527082ec9b551 altogether; probably fine

nanavati:
Rebased onto current main (941eecfe) and force-pushed (a86b1a2b). What the rebase resolved: Same port-splitting integration on the foreign-function port-size path: the removed `addSizes` hack's surroundings became per-argument grouped singleton port lists. Still stacks on #1022's commits.

PR #1024, opened 2026-07-06
Comment to keep ghc version being tested in sync with install doc

kenta2:


quark17 (approved):

quark17:
FYI, I would have added this to PR #1003, rather than open a new PR.  (Particularly to reduce the number of CI flows that we create, since there's a queue still waiting to run.)

PR #1026, opened 2026-07-07
A tree-sitter grammar for Bluespec Classic

mieszko:
A tree-sitter grammar for Bluespec Classic, suitable for use as syntax highlighting.

### DIFF
--- a/util/tree-sitter-bluespec/.editorconfig
+++ b/util/tree-sitter-bluespec/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
--- a/util/tree-sitter-bluespec/.gitattributes
+++ b/util/tree-sitter-bluespec/.gitattributes
@@ -1,0 +1,6 @@
+* text=auto eol=lf
+
+# Generated source files
+src/*.json linguist-generated
+src/parser.c linguist-generated
+src/tree_sitter/* linguist-generated
--- a/util/tree-sitter-bluespec/.gitignore
+++ b/util/tree-sitter-bluespec/.gitignore
@@ -1,0 +1,4 @@
+*.o
+*.so
+*.dylib
+*.dll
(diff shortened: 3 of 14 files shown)


quark17:
We set up the repo [B-Lang-org/language-bh](https://github.com/B-Lang-org/language-bh) to hold a grammar for BH, and [registered](https://github.com/github-linguist/linguist/blob/e9fe3c9f230cd9220afcd057f75702de4d7700c9/.gitmodules#L664) that URL with GitHub's Linguist repo (so that GitHub highlighting uses it).  The `language-bh` repo is also [published](https://marketplace.visualstudio.com/items?itemName=B-Lang-org.language-bh) as a VSCode plugin, that people can install from the VS Marketplace, to do highlighting.  (Although it might need updating to newer dependency versions.)

That grammar is just a Haskell Textmate grammar that's been adapted, with the idea that people could contribute to improving it.  If you've got something better, we could replace it with your grammar? (rather than put it here in `util`)

mieszko:
I'm not sure I follow. This is a tree-sitter grammar, not a textmate grammar, though?

mieszko:
Also, I only put it in util b/c there are already highlighter syntaxes there. But yeah ideally it would have its own repo under b-lang-org, otherwise the whole repo gets pulled in just to extract and compile the grammar. I can do that instead if you like?

matx-amy:
I think it'd be better to put it in its own repo if the plan is to commit the very large generated files.

I'd also be happier if this had some tests to ensure it correctly parses everything in the repo; I'm doing that like this ([flake.nix#L47](https://git.sr.ht/~remexre/sylvan/tree/d09e26350f11a9c15d704ce6f72bf9ac39f90f94/item/flake.nix#L47), [misc/tree-sitter-check-parse/main.c](https://git.sr.ht/~remexre/sylvan/tree/d09e26350f11a9c15d704ce6f72bf9ac39f90f94/item/misc/tree-sitter-check-parse/main.c)) in a different compiler.

(I'm not committing generated files there, but that's because I'm pinning the version of tree-sitter; tree-sitter has breaking changes often enough that it's probably a good idea to do one of the two if this is "first-class" instead of "best-effort.")

PR #1027, opened 2026-07-07
Add FST support to Bluesim

nanavati:
Bluesim can now dump waveforms in FST (GTKWave's Fast Signal Trace format) as well as VCD. FST files are compressed and typically 10-100x smaller than the equivalent VCD, which matters for long simulations of large designs, and they load incrementally in GTKWave.

**Stacked on #1000 and #956.** The first 5 commits are #1000 (`-dump-formats`) and the next 11 are #956's wiretypemap work (which the FST correlation test builds on); the remaining commits are new here. Merging those PRs first will shrink this one to the FST commits.

## What's here

- `src/vendor/libfst`: GTKWave's standalone libfst vendored as a submodule, pinned at a revision with the LZ4 1.10.0 CVE fixes. Licenses: fstapi and FastLZ are MIT, LZ4 is BSD-2-Clause — see `LICENSES/LICENSE.fst` and the COPYING update. Compiling it requires zlib headers (`zlib1g-dev`; the macOS SDK provides zlib).
- Kernel rework (`src/bluesim/`): the VCD-specific dump path in vcd.cxx is refactored behind a small `WaveWriter` interface (`bs_wave_writer.h`), with a byte-identical `VcdWriter` and a new `FstWriter` (fst.cxx, over libfst). Per-signal dump code in generated models is format-independent and shared. The forward-buffering policy for backdated combinational changes is unchanged, so file writes stay time-monotonic in both formats. (The dead never-executed file-resume path — `previous_files` was find/clear-only — is deleted rather than ported.)
- Link-time selection: `-dump-formats fst` (or `vcd,fst`) compiles the chosen writers into the model; `fst` pulls in `-lz`; models without FST carry neither the code nor the dependency. Bluesim rejects `fsdb` (a Verilog-simulator-only format) at link time. `-dump-formats none` behavior from #1000 is preserved, and the per-module object-reuse fingerprint now accounts for whether dump code was generated (fixing stale reuse when relinking across `none` and `vcd`/`fst`).
- Runtime selection UX: `+bscvcd[=file]` / `+bscfst[=file]` plusargs (uniform with the Verilog simulators), the existing `-V` flag (VCD, unchanged), and bluetcl `sim fst` alongside `sim vcd` — both queries are format-aware and report only a file dumped in their own format. Requesting a format the model wasn't built with is a loud runtime error; a mid-run format switch cleanly ends the current dump. Models built by older bsc load fine in newer bluetcl (the loader falls back to VCD-only when the format API is absent).
- Tools: `fstscopes` (dump an FST hierarchy including the scope module types FST records) and `fstcheck` (vcdcheck for FST); vcdcheck's checker core is factored into a shared `WaveCheck.hs`. Built by `make install-extra`.
- Docs: user guide (`-dump-formats`, plusargs, bluetcl `sim fst`), INSTALL.md (submodule + zlib), CI dependency scripts.

## How it's tested

- Every Bluesim test that does the standard VCD cross-run now also links with `-dump-formats vcd,fst` and runs a third time with `+bscfst`, comparing output against the plain run (~1,100 tests in the suite exercise the FST writer).
- `testsuite/bsc.bluesim/fst/`: dedicated tests for dumping, `$dumpon`/`$dumpoff` tasks, format switching, X-states, and fst2vcd round-trips (skipped when GTKWave's fst2vcd is absent).
- `testsuite/bsc.bluetcl/fst_correlation/`: FST twin of the vcd_correlation test — wiretypemap correlation against both iverilog (`vvp -fst`) and Bluesim FST dumps, additionally checking the scope module types that FST records and VCD cannot.


src/bluesim/vcd.cxx
```diff
@@ -31,39 +28,391 @@ static const unsigned int minor_rev = 1;
...
+      return false;
+    unsigned long long sz = (unsigned long long) ftello(file);
+    return (sz > size_limit);
+  }
+};
+
+static WaveWriter* new_vcd_writer()
+{
+  return new VcdWriter;
+}
```
matx-amy, line 242: why wrap this instead of just using `new VcdWriter` below?

src/Verilog/main.v
```diff
@@ -70,28 +72,45 @@ module main();
       else if (do_vcd)
 	filename = "dump.vcd";
 
-      if ($value$plusargs("bscfsdb=%s", filename))
+      if ($value$plusargs("bscfst=%s", fst_filename))
+	do_fst = 1;
+      else if (do_fst)
+	fst_filename = "dump.fst";
```
matx-amy, line 78: weird indentation?

src/comp/WaveCheck.hs
```diff
@@ -0,0 +1,660 @@
...
+               timeok = timeOK (sc_nextTime cmd) time
+               valueok = changed_val == (sc_crcfunction cmd) stable_val
+               vnow = to_VCDValue (sc_width cmd) changed_val
+               vexpected = to_VCDValue (sc_width cmd) $ (sc_crcfunction cmd) stable_val
+               --
+               doWaiting =
+                   let prop = show cmd ++ " Started"
+                       ok =  (sc_startVal cmd) == changed_val
+                       cmd' = cmd {sc_state = Running
+                                          ,sc_nextTime = incrTime time (sc_timeIncr cmd) }
```
matx-amy, line 465: weird indentation?

src/comp/SimBlocksToC.hs
```diff
@@ -187,12 +187,18 @@ convertModuleBlock flags sb_map ff_map clk_map wdef_mod_map reused top_blk write
         -- list of subblocks that need to be included
         include_ids = nub (map (\(id,_,_)->id) (sb_state sb))
 
+        -- whether to emit the per-signal waveform dumping code, which is
+        -- shared by the VCD and FST writers (-dump-formats vcd/fst);
+        -- with -dump-formats none it is stubbed out, dropping a lot of
+        -- generated code in large/replicated designs.
+        genVCD = any (`elem` ["vcd", "fst"]) (dumpFormats flags)
```
matx-amy, line 194: should this be called like, genWaveform or something?

testsuite/bsc.bluetcl/commands/WireTypes.bsv
```diff
@@ -0,0 +1,256 @@
+package WireTypes;
```
matx-amy, line 1: can one of the duplicate files be a symlink to the other? otherwise there's risk of divergence (and i think the matx fork actually did diverge in https://github.com/MatX-inc/bsc/commit/c05cd68e)

matx-amy (approved):

nanavati:
Rebased onto current main (941eecfe) and force-pushed (bf4ec131). What the rebase resolved: Mechanical compile fixup only: `aIfaceName` → `aif_name` (upstream removed the helper in ddcd1d3f, "Port splitting part 2"); carried as an explicit rebase-fixup commit at the head.

PR #1028, opened 2026-07-08
Fix spurious T0029 when a field reaches an ATF through a type synonym

nanavati:
## Summary

Fixes two holes where a type function (ATF) hidden behind a type synonym behaved differently from the directly-written form. The branch is based on upstream PR B-Lang-org/bsc#916 ("Fix #890 by expanding type functions in tiExpl") — the first 9 commits are that PR unchanged; the last two commits are the fixes.

### 1. `Unify.hs` — spurious T0029 on auto-derived Generic (`7a4c248`)

With PR 916's instance-head expansion, a data/struct field whose type reaches an ATF through a synonym (`type ElemSyn f = Elem f`) made the auto-derived Generic instance fail with a spurious "signature too general" (T0029), while the direct `Elem f` spelling compiled. Code with this shape is common enough that the regression breaks substantial existing designs.

The instance head (given scheme) expands synonyms and generalizes the exposed ATF into a fresh class-constrained variable, but during typecheck of the derived method bodies, `mgu` did not recognize the synonym-hidden ATF: `isATFAp` sees only a literal `TIatf` head, so unification structurally bound the head's fundep-determined variable to the unexpanded `ElemSyn c` — pinning it. The deduced scheme then quantified one fewer variable than the given scheme and `tiExpl'''` reported T0029. The direct spelling works because `atfUnify` defers `boundVar ~ Elem c` into an equality that becomes the satisfiable class predicate `Container c e`.

Fix: in `mgu`, expand a saturated type-synonym application when (and only when) its expansion is a fully-applied ATF application, routing it into the existing `atfUnify` deferral. Synonyms whose expansion is not ATF-headed — notably the parameter-dropping TLM-style synonyms that forced earlier expansion attempts to be reverted (`CExtendSynonym`) — keep the existing structural behavior. This also makes the missing-context error for a synonym-hidden ATF identical to the direct case (T0030 naming the needed context), while genuinely-too-general signatures still fail with T0029.

### 2. `MakeSymTab.hs` — instance-head check bypass (`dcad864`)

`checkNoTypeFunInHead` scanned instance-head arguments syntactically, so an ATF reaching a non-determined head position through a synonym escaped the T0156 check: depending on the shape it either failed with a confusing T0079 or was silently accepted with the head rewritten by instance-head expansion — while the identical head written directly was rejected.

Fix: resolve constructor sorts and expand each saturated synonym application in non-determined positions, rejecting ATF applications that mention a type variable or are under-applied (they can neither reduce away nor participate in instance matching), reported at the synonym-use position. Ground applications remain accepted — that is the Bug 1729 / #311 behavior pinned by `ExpSizeOf_InstancesBaseSyn` (`typedef SizeOf#(T) ST` in a head). Determined positions remain exempt, synonym or not. Note the directly-written ground form is still rejected (as pinned by `ATFInInstHead`); relaxing that for reducible applications could be considered separately.

## Tests

- `ATFSynonymField.bs` — reproducer: data and struct fields reaching an ATF through a synonym compile (was 2× T0029)
- `ATFSynonymFieldUse.bs` — same shapes through Bits derivation, explicit Generic from/to, and Verilog elaboration
- `ATFInInstHeadSynonym.bs`, `ATFInInstHeadSynonymElsewhere.bs` — variable ATF applications behind synonyms in heads, rejected with T0156 like the direct forms
- `ATFInInstHeadSynonymGround.bs` — ground application behind a synonym, still accepted
- `ATFInInstHeadDeterminedSynonym.bs` — determined position behind a synonym, still accepted

## Validation

- `bsc.typechecker/typeclasses`: 107/107, `bsc.typechecker/primtcons`: 31/31 (includes all PR 916 tests: `ExpSizeOf_*`, `CExtendSynonym`, `CExtendATF`)
- Full `bsc.typechecker` recursively: 1184 passes, 0 unexpected failures; `bsc.bugs`: 940 passes, 1 environment-only failure (permission-denied test running as root); `bsc.bluetcl`: 189/189
- Full compiler + library rebuild clean with both changes

https://claude.ai/code/session_01M8cZoDh1jk24UJbB51y5tU

### DIFF
--- a/src/comp/MakeSymTab.hs
+++ b/src/comp/MakeSymTab.hs
@@ -491,9 +491,11 @@ checkNoTypeFunInHead errh r mi clsId args =
         -- Only check non-determined positions
         nonDetArgs = [ arg | (idx, arg) <- zip [0..] args
                      , not (S.member idx determinedIdxs) ]
-        checkArg a = case findTypeFun a of
-                       [] -> findSynTypeFun a
-                       ds -> ds
+        -- Report both the directly-written type functions and the ones
+        -- hidden behind synonyms, deduplicated (a directly-written type
+        -- function inside a synonym's argument can also appear in the
+        -- synonym's expansion).
+        checkArg a = nub (findTypeFun a ++ findSynTypeFun a)
         found = concatMap checkArg nonDetArgs
     in if null found then ()
        else bsErrorUnsafe errh
--- a/src/comp/MakeSymTab.hs
+++ b/src/comp/MakeSymTab.hs
@@ -464,10 +464,37 @@ checkNoTypeFunInHead errh r mi clsId args =
             | otherwise = []
         findTypeFun (TAp f a) = findTypeFun f ++ findTypeFun a
         findTypeFun _ = []
+        -- A type function can also be hidden behind a type synonym.
+        -- Expand each saturated synonym application and reject any type
+        -- function application in it that mentions a type variable (or
+        -- is not fully applied): such an application can neither be
+        -- reduced away nor used for instance matching.  Ground
+        -- applications are left alone; context reduction expands and
+        -- reduces them to a concrete type (Bug 1729, GitHub issue #311;
+        -- see ExpSizeOf_InstancesBaseSyn in the testsuite).  The error
+        -- is reported at the position of the synonym use.
+        synArity i | Just (TypeInfo { ti_sort = TItype n _ }) <- findType r i = Just n
+                   | otherwise = Nothing
+        findSynTypeFun t =
+            case splitTAp t of
+              (TCon (TyCon i _ _), as)
+                | Just n <- synArity i, toInteger (length as) >= n ->
+                    [ (getPosition i, tf)
+                    | tf <- varTypeFuns (expandSyn (updTypes r t)) ]
+              (_, as) -> concatMap findSynTypeFun as
+        varTypeFuns t =
+            case splitTAp t of
+              (TCon (TyCon i _ (TIatf { atf_param_idxs = pIdxs })), as)
+                | length as /= length pIdxs || not (null (tv as)) ->
+                    i : concatMap varTypeFuns as
+              (_, as) -> concatMap varTypeFuns as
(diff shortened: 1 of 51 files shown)


src/comp/Unify.hs
```diff
@@ -24,8 +25,15 @@ class Unify t where
...
     -- different types generate a deferred equality constraint.
+    -- A type synonym is expanded when (and only when) that exposes an ATF
+    -- application, so that a synonym-hidden ATF unifies exactly like the
+    -- direct ATF application, instead of a variable being structurally
+    -- bound to the unexpanded synonym (which pins a fundep-determined
+    -- variable and makes derived-instance schemes spuriously mismatch).
     mgu bound_tyvars t1 t2
-        | isATFAp t1 || isATFAp t2 = atfUnify bound_tyvars t1 t2
+        | isATFAp t1' || isATFAp t2' = atfUnify bound_tyvars t1' t2'
+      where t1' = expandSynATF t1
```
quark17: Why not just call `expandSyn`?  Did you define `expandSynATF` to avoid calling `expandSyn` in some circumstances, for efficiency?  If so, why not merge the call here to `isATFAp` into the function you defined?

src/comp/MakeSymTab.hs
```diff
@@ -464,10 +464,37 @@ checkNoTypeFunInHead errh r mi clsId args =
...
+                | length as /= length pIdxs || not (null (tv as)) ->
+                    i : concatMap varTypeFuns as
+              (_, as) -> concatMap varTypeFuns as
         -- Only check non-determined positions
         nonDetArgs = [ arg | (idx, arg) <- zip [0..] args
                      , not (S.member idx determinedIdxs) ]
-        found = concatMap findTypeFun nonDetArgs
+        checkArg a = case findTypeFun a of
+                       [] -> findSynTypeFun a
+                       ds -> ds
```
quark17: This omits the syn typefuns from the error if there are non-syn typefuns.  Is that intentional?  Would it not be preferable for BSC to report as many errors as it can?

PR #1032, opened 2026-07-09
Corner-case fixes to instance trie lookup

nanavati:
The PredTrie-based instance index (05865c6f) can select or admit instances that the legacy all-pairs index would not. This PR fixes three such defects, each of which undermines the coherence certificate that instance resolution reports (and that any future coherent-dictionary work would depend on), and adds a testsuite directory that pins the behavior under both index implementations.

### 1. Leaf ordering: sorting a partial order with `sortBy`

`cmpQInsts` yields only a *partial* order — non-overlapping instances are incomparable — and `sortBy` with the EQ-collapsed comparator is not a sound way to linearize a partial order: an incomparable instance sitting between two comparable ones can keep the sort from ever comparing them. A trie leaf can then keep a catch-all ahead of a strictly-more-specific instance, and a fully concrete pred selects the catch-all with **no incoherence flagged** (and no T0128, since the pair is orderable).

**Fix:** order each leaf with a topological sort of the strict specificity edges, the same way `getQInstsLegacy` orders the whole list. To keep this from adding pairwise comparisons, the overlap check and the leaf sort now share a lazily memoized `cmpQInsts` table keyed on the canonical index pair — same-leaf pairs were already being compared by the overlap probes, so the sort costs no additional `cmpQInsts` calls (library build time is unchanged).

### 2. Primitive type functions in instance heads keyed invisibly

An instance head like `NP2 (TAdd a b)` was keyed under `Con TAdd`, where probes for the numeric-literal instances it can evaluate to never look. The overlap check therefore silently admitted instance pairs the legacy index rejects with T0128.

**Fix:** key unstable heads as `Nothing` (the catch-all branch) so such instances are visible to every query and every overlap probe. The legacy index's `canMatch` pre-filter gets the matching correction: prim-type-function heads are treated as inconclusive (`TAdd a b` vs a `TyNum` head is unifiable, not a constructor clash).

### 3. Stuck associated type functions in queries probing too narrowly

`normT` leaves a fully applied ATF in place when its class predicate is not yet satisfiable, and `mkQueryElem` keyed such a pred argument as `Con (ATF)` — probing only the ATF's own (empty) branch plus the catch-all branch. Concrete instances the pred could match after the ATF resolves were neither tried nor visible to the incoherence check, so a catch-all match was committed as if it were coherent.

**Fix:** treat ATF heads as `Free` (probe all branches), like prim type functions.

## Tests

New directory `testsuite/bsc.typechecker/instances/order`, with every test run under both the trie index and `-legacy-no-instance-trie`:

- **LeafOrder / LeafOrderChain** — same-leaf specificity traps (including a four-instance chain with an incomparable instance declared in an order that defeats a comparison sort); the most specific matching instance must be selected under both indexes.
- **TFunHead** — a `TAdd`-headed instance pair that must fail with T0128 under both indexes.
- **StuckATF** — a stuck-`SizeOf` pred whose catch-all match must be deferred and reported in the weak-context error rather than silently committed. Surface syntax always expands ATF applications into class contexts at conversion, so this is pinned as an invariant guard (protecting internal paths that skip an expansion gate) rather than a reproduction of a live bug.
- **FdStability** — pins the resolution behavior that makes fundep improvements from a matched instance unsafe to commit before the instance's context is discharged: a specific instance matches first and would improve the fundep output, but a sibling pred later fixes that output differently and the catch-all is selected coherently on retry. Identical output required under both indexes.

Against the pre-fix compiler the suite fails in exactly the two places that expose the reachable bugs (LeafOrder's simulation output under the default index, and TFunHead compiling when it should not), with the legacy variants passing; against the fixed compiler all checks pass.

## Validation

- The standard library builds **byte-identically** before and after this change.
- Full `bsc.typechecker` sweep plus the new `instances/order` directory pass against the fixed compiler.

**The stack**: **#1032 (instance-trie fixes, this PR)** and #1033 (coherence keywords + coverage warning), independent → #1035 (ordered-clause commitment) → #1036 (SAT settlement) → #1037 (bound-variable discipline) → #1038 (solved-dictionary pool). Each PR contains the ones below it.

https://claude.ai/code/session_01YMPXjaiFCjeVJv1TbhYfd5


src/comp/MakeSymTab.hs
```diff
@@ -980,17 +983,54 @@ getCls errh mi src_pkg iks r incoh ps ik vs fds ifs qts =
...
+                -- every pure-input position, and the leaf's pairs have
+                -- already been compared by the overlap check, so this
+                -- costs no additional cmpQInsts calls.
+                sortLeaf items@(_:_:_) =
+                    let g = [ (i, [ j | (j, _, _) <- items, i /= j,
+                                        pairCmp j i == Right (Just LT) ])
+                            | (i, _, _) <- items ]
+                        im = M.fromList [ (idx, item) | item@(idx, _, _) <- items ]
+                    in  case tsort g of
+                          Right is -> [ im M.! i | i <- is ]
```
quark17: The function `!` reports a GHC error if not found.  Do we want to replace it with something that calls `internalError`?  (`Util.hs` has things like `headOrErr`, but I don't see that it has anything for Map lookup.)

testsuite/bsc.typechecker/instances/order/order.exp
```diff
@@ -0,0 +1,64 @@
...
+# A pred with a stuck associated-type-function head must probe every
+# trie branch, so that unifiable instances (NPick 1 vs NPick (SizeOf a))
+# remain visible to incoherence detection.  This pins the invariant
+# rather than exposing a live bug: every current surface-syntax inlet
+# (signatures, annotations, data fields, instance heads, synonyms)
+# expands ATF applications into class contexts at conversion, so the
+# pred reaches the probe as NPick a__ with a fresh variable.  The pin
+# guards the probe behavior for any future path that skips one of
+# those expansion gates (as mkVPredFromPred already does).
+compile_fail StuckATF.bs
```
quark17, line 63: Does it not make sense to also test this with `-legacy-inst-index`?

quark17 (commented):
Looks good, though I haven't thought about it too deeply

quark17 (approved):

PR #1035, opened 2026-07-10
Ordered-clause fundep semantics: commit coherent instance matches

nanavati:
A functional dependency declares that a class's determined positions are a function of its input positions. BSC's resolution has historically violated this in two ways: an instance whose inputs matched but whose determined positions conflicted with the predicate was treated as a non-match (falling through to a later, less specific instance — letting the demanded *output* choose the instance), and even a coherently matched instance was backed out and retried rather than committed. This PR makes the ordered-clause semantics explicit and commits coherent matches by default.

Stacked on #1032 and #1033: the first 8 commits here are those two PRs linearized; **the 7 commits from "Give the constraint solver's results proper types" onward are this PR's content**, and each is written to be reviewable on its own.

## Ordered clauses: outputs never bar a match

The old fall-through made instance selection unstable under refinement (a matched instance could be lost when its output position was refined from elsewhere) and produced order-dependent acceptance: with a specific instance `TC (Maybe a) 5` and a catch-all `TC x 6`, the same two-pred program compiled or failed depending on argument order. It is also the behavior a long-standing comment in `matchTopIsReducible` worked around ("we'd like to return Fails, but to support overlapping instances which disagree on the fundeps...") and that `pipeline.exp` asked about ("the typeclass instances break the stated dependency requirement; why doesn't BSC catch this?").

Under the new semantics, a class's instances form ordered clauses of a type-level function keyed on input positions: the first clause whose inputs match is selected, the fundep improvement always runs, and if the determined positions cannot unify the predicate is unsatisfiable as stated — the walk stops instead of falling through. A class annotated `incoherent` (#1033) is the user's declaration that resolution is *not* a function of its inputs (the `Has_tpl_n` tuple machinery is provably not one), so such classes keep the fall-through behavior. Given-discharge is unchanged: a conflict against a given just means that given does not discharge the predicate.

## Commit coherent matches by default

When the selected clause is coherent — no earlier clause could still capture the goal under any refinement of its free variables (the `earlierInstanceMayCapture` modal check) — the solver keeps the reduction's bindings and continues with its subgoals, instead of putting the original predicate aside to retry from scratch on a later pass. Predicate ancestry records what each residual was reduced from, so context errors are reported in terms of the user-written predicate, anchored at the offending expression rather than the enclosing binder.

`-legacy-defer-instances` is the escape hatch: it disables the ordered-clause semantics and commitment together.

## Fundep conflicts get a dedicated error (T0159)

A committed clause whose determined positions cannot unify with the demanded type was previously reported with the generic "there are no instances of the form" (T0031) — actively misleading, since an instance matching the inputs exists. The new T0159 shows the predicate and the selected instance side by side and explains that instances are chosen by the fundep input positions; the check fails fast when the conflict is final.

## Commit-by-commit

1. **Give the constraint solver's results proper types** — pure mechanical refactor: records `SolveResult`, `Reduction`, `InstMatch` replacing the solver's positional tuples. No behavior change.
2. **Ordered-clause fundep semantics: outputs never bar a match** — the clause-selection change above, carried by new result types: `Match a` (NoMatch / Conflict / Match), `firstMatch`, the `matchFDRow` row core, `SatResult`'s `Commitment` certificate, and the `earlierInstanceMayCapture` modal check.
3. **Commit coherent instance matches by default, with predicate ancestry** — the commitment machinery, acting on the certificate and modal check the previous commit introduced, and ancestry-based (root-anchored) error reporting.
4. **Report ordered-clause fundep conflicts (T0159), failing fast when final.**
5. **Keep diagnostics rooted, positioned, and complete under commitment** — the four reporting-quality holes rooted reporting exposed: rooted reports carry the residual's accumulated source positions; synthesis-failure reports (T0043) fall back to the predicate's carried positions instead of "Unknown position"; the weak-context "could also be deduced from" hint reduces self-contained so it lists the complete sub-proviso set; and fail-fast probes in generated wrapper code defer to the definition-level report when nothing carries a source position. Regenerated expectations where messages improved: gh221, gh894, BuildList, BuildVector (errors anchor at the offending expression, dropping one-entry "implied by" indirection lists; gh894's T0033 ambiguity dump becomes a T0032 naming the missing instance's concrete obligation), and the noinline and `bsc.codegen/signature` synthesis-failure tests (two T0043s: the definition-level report naming the module, and the rooted one at the method position, upgraded from a positionless T0031).
6. **testsuite: ordered-clause commitment pins** — new directory `testsuite/bsc.typechecker/instances/commit`: legacy-hatch behavior, commitment ancestry messages, synonym inputs, given/commit interaction, incoherent fall-through, output determination through full and partial commits, cross-branch clause walks, weak-context rooting.
7. **Document coherent instance selection in the user-facing guides** — BSV and BH reference guides and the user guide.

## Validation

- Full `bsc.typechecker` sweep (1,246 tests, including the new pins) passes clean at this tip.
- Every directory whose expectations this PR touches passes clean.

**The stack**: #1032 (instance-trie fixes) and #1033 (coherence keywords + coverage warning), independent → **#1035 (ordered-clause commitment, this PR)** → #1036 (SAT settlement) → #1037 (bound-variable discipline) → #1038 (solved-dictionary pool). Each PR contains the ones below it.


nanavati:
Full-testsuite result at this PR's head (ff278307), as promised in the description:

- **18,489 PASS / 129 XFAIL / 6 FAIL** across 868 test directories (`make fullparallel`, Verilog backend via iverilog 12.0, C backend, and SystemC tests enabled).
- All 6 FAILs are environment artifacts of the test machine, reproduced identically on near-main builds (2026.01-141) without this PR: 5 are the cascade from a deterministic GHC RTS stack overflow in `bsc.long_tests/log2_loop/actionvalue` under that test's `+RTS -M300M` cap, and 1 is iverilog 12's "Excess hex digits" `$readmemh` warning polluting `bsc.long_tests/MPEG4`'s expected simulation output.

No failures attributable to this PR. The same suite run at the next branch stacked on this one (SAT-batching, this head + 2 commits) shows the identical 6 artifacts and nothing else.

krame505 (commented):
The idea makes sense here and I don't see any obvious issues with the implementation.

nanavati:
Rebased onto current main (941eecfe) and force-pushed (fe137df6). What the rebase resolved: Re-integrated instance commitment with main's new ATF resolution cache (`recordATFs` only on coherent, final matches; `mergeCATFCaches` threading) and the transitive-incoherent warning bookkeeping; goldens follow main's SplitPorts-aware synthesis errors (T0029 → T0043 in the noinline test).

PR #1036, opened 2026-07-10
Settle numeric provisos once per definition, in one SAT session

nanavati:
Numeric provisos that instance reduction cannot discharge were sent to the proviso SAT solver from inside `reducePred`, one predicate at a time, with a fresh solver spun up (and torn down) per query. A definition with N numeric residuals paid N solver initializations per satisfy pass, and the same predicate could be re-posed by every pass that revisited it. This PR removes the per-predicate fallback: numeric residuals now stream out of every satisfy pass and settle **once, at their owner**, in a single solver session per batch.

Stacked on #1035 (which carries #1032 and #1033): **the 2 commits from “Settle numeric provisos in one SAT session at their owner” onward are this PR's content**.

## One session per batch, at the settlement owner

`batchSolveNumericPreds` proves a definition's whole residual set in one solver session: one initialization, the definition's numeric givens asserted once, each residual posed as a query against them — with queries already decided in this definition answered from a cache (`tsNumProven` / `tsNumRefuted`) instead of re-posed. `reducePred`'s givens parameter, which existed only to feed the removed fallback, is renamed `_eps` and remains as the seam where a given-aware walk would plug in.

The owners: `tiExpl'''` proves this definition's retained and deferred residuals in one batch, after the ds/rs split and before defaulting (a deferred pred provable only at an enclosing binding survives its batch untouched and defers as before; ground preds must settle here or be misreported as unsatisfiable). The exported `satisfy`/`satisfyFV` keep their verdict-complete contract — `normT`'s expand-or-not decision, the foreign-import proviso check, `ISyntaxCheck`'s `eqTypeNum`, and the error reporters are unchanged — while new `satisfyStream`/`satisfyFVStream` variants serve the callers that own a later settlement point (`tiExpl'''`, and `tiImpls` whose residuals settle in `reducePredsAggressive`).

Two verdict sites must not adopt a match while its numeric debt is unproven, and settle locally: an incoherent-class match is deliberately revocable until its context is fully discharged, so its full-vs-partial verdict gets a local batch; and `tryDefault`'s candidate validation settles under each trial substitution (the speculative pass is already quarantined — only an accepted candidate's substitution is merged).

Deferred debt belongs to its owner: a non-ground deferred pred rides upward unqueried and settles once, at the binding that owns its variables — except through a binding with numeric-class givens (the only assumptions the solver can assert), whose local proof attempt may be load-bearing. Ground deferred preds settle where they stand.

## Measured impact

On FloatingPoint.bsv (same host, same flags), before vs after: solver sessions **501 → 24**, queries **1418 → 115**, wall-clock **12.9s → 9.6s (−26%)**. The dominant collapse comes from the desugared local bindings, which previously re-queried inherited debt at every nesting level.

## Soundness fix (the ICE pinned by `bsc.bugs/bluespec_inc/b675`)

`byInst`'s dictionary binding references only the instance's declared proviso dictionaries; the numeric equalities a fundep match defers (`num_eqs`) become goals but not arguments, so the binding's free variables do not reveal the obligation — a consumer walking a binding's dependencies would treat the closure as complete even when the equality was never established. `SolvedBinds.addBindDeps` now records the equality dictionary ids as semantic dependencies of the bind: an unresolved equality shows up in the walk, a solved one is walked through.

## Tests

New directory `testsuite/bsc.typechecker/numeric-settle`, where both obligations are chosen to be provable *only* by the SAT solver — the cross-class entailment `Mul#(a,2,b) ⊢ Add#(a,a,b)`, which no direct instance can synthesize (the output position is pinned) and no commuted given can supply. Each test is certified by a dual variant: it must **fail** with the solver disabled (`-no-use-proviso-sat`) and pass with it enabled — a test that still passed without the solver would not be exercising settlement.

- **DebtRidesUp** — a given-free local's obligation rides up (unqueried) and is proven by the enclosing definition's `Mul` proviso, in that binding's batch.
- **IncoherentNumBatch** — an incoherent-class match takes a local numeric verdict: its instance's `Add` proviso is proven in a local batch, so the specific instance wins; without the verdict the match looks partial and selection falls to the catch-all, which cannot satisfy the signature.

## Validation

- Full `bsc.typechecker` sweep including the new dual-variant pins passes clean.

**The stack**: #1032 (instance-trie fixes) and #1033 (coherence keywords + coverage warning), independent → #1035 (ordered-clause commitment) → **#1036 (SAT settlement, this PR)** → #1037 (bound-variable discipline) → #1038 (solved-dictionary pool). Each PR contains the ones below it.


nanavati:
Full-testsuite result at this PR's head (d8a5cab9), as promised in the description:

- **18,493 PASS / 129 XFAIL / 6 FAIL** across 869 test directories (`make fullparallel`, Verilog backend via iverilog 12.0, C backend, and SystemC tests enabled), from a from-scratch build.
- All 6 FAILs are environment artifacts of the test machine, reproduced identically on near-main builds without this stack: 5 are the cascade from a deterministic GHC RTS stack overflow in `bsc.long_tests/log2_loop/actionvalue` under that test's `+RTS -M300M` cap, and 1 is iverilog 12's "Excess hex digits" `$readmemh` warning polluting `bsc.long_tests/MPEG4`'s expected simulation output. The same two artifacts (and nothing else) appear in #1035's run at this branch's base.

The count reconciles with the base exactly: #1035's 18,489 passes + the 4 new dual-variant checks in `bsc.typechecker/numeric-settle` = 18,493.

krame505 (commented):
Relative diff at https://github.com/nanavati/bsc/compare/claude/ordered-clause-commitment...nanavati:bsc:claude/sat-batching-settlement

Seems fine to me.

nanavati:
Rebased onto current main (941eecfe) and force-pushed (85c37e12). What the rebase resolved: The settle-once SAT-session commit re-expressed over main's `TIResult` record (was a `runTI` tuple) and the `DirectIncoherence`/`isIncoherent` bookkeeping (`markIncoherent` now also sets the flag).

PR #1038, opened 2026-07-10
Pool solved dictionaries under a closed certificate

nanavati:
`propagateFunDeps` runs a full satisfy per apply node purely for its fundep-improvement substitution and throws the dictionary bindings away; the same predicates are then re-derived at the next apply node and again at generalization time. Now that coherent instance matches are committed (#1035), those results are final and can be kept. This PR makes solved dictionaries shareable — first within a satisfy pass, then across passes and definitions through a TI-state pool — with an explicit, audited certificate for when a predicate may leave the flow.

Stacked on #1037 (which carries #1032, #1033, #1035, and #1036): **the 6 commits from "Order SolvedBinds at emission; share solved dictionaries within satisfy" onward are this PR's content**.

## Emission-time ordering unblocks sharing

`SolvedBinds`' merge-order invariant ("new must not depend on old") was an artifact of maintaining `Cletseq` dependency order by construction. The actual requirement is only that non-recursive bindings be in dependency order when emitted, and the dependency graph stays acyclic even when later bindings reference earlier dictionaries. The merge is now plain accumulation and emission normalizes instead: non-recursive bindings that transitively reference recursive ones are promoted into the `Cletrec` group, and the rest get a stable topological repair that returns already-ordered input unchanged.

That unblocks intra-pass sharing, which the invariant previously rejected with an internal error: `sat`, `satMany`, `satMany'`, and `satisfy`'s `sMany` thread a "solved" list recording the dictionary of every fully solved or committed non-numeric predicate. The list is consulted after the givens, discharging an equal predicate against the existing dictionary instead of re-deriving its reduction subtree. Adoption follows the bindings: branches that discard a reduction's `SolvedBinds` also discard its solved entries. The committable guard deliberately ignores the solved list — a given blocks commitment because it is evidence not derivable from instances, whereas a solved dictionary is itself instance evidence and cannot contradict the instance choice.

## The cross-pass pool

Fully solved predicates are deposited into a TI-state pool (`tsSolvedPool`/`tsPoolSbs`): the entry's transitive binding closure — extracted via the new `SolvedBinds.extractClosures` — is held for emission with the enclosing binding group's letseq, the predicate is dropped from the flow (returning it would double-bind its dictionary), and `sat` consults the pool just before walking instances. Pool frames are pushed/popped around `tiExpl'''`, `tiImpls`, and `topExpr`, whose emitted letseqs scope over the whole group body; entries die with their frame, and thrown errors restore the state wholesale.

Pool entries must be pure instance derivations: a closure that references any in-scope given's dictionary is refused. `propagateFunDeps` solves against givens from all enclosing levels, but an explicitly-typed definition's final solve may use only its own context plus instances; a given-derived entry would carry that evidence into nested definitions not entitled to it, making acceptance depend on whether an earlier apply node happened to solve the same predicate.

## Non-ground entries: each hazard met by an explicit mechanism

Ground-only pooling dodged three hazards implicitly; making them explicit lets fully solved predicates with free type variables pool too:

- **Coherence**: a non-ground predicate *can* match incoherently, so `markIncoherent` fires on the incoherent path in `sat`, and `extractClosures` refuses any closure passing through a marked binding — information-dependent choices are never pooled. The marking earns its keep: a partial application matching an incoherent class's catch-all is marked, refused, kept flowing, and re-resolves to the specific instance once its second argument is known; pooling it would have frozen the catch-all.
- **Substitution currency**: entries carry live unification variables, so the pool is snapshotted per satisfy pass and threaded with the same incremental-substitution discipline as the intra-pass solved list.
- **Generalization**: a nested definition could generalize over an entry's free type variable, so non-ground entries are frame-local — deposited in `tsSolvedPoolNG` and hidden from nested frames by `pushSolvedPool`.

A fourth interaction was exposed by validation: `tiRecoveringFromError` — wrapping every `action..endaction` block and rule — trims tyvar mappings created inside the block on success, so a pool binding deposited inside kept a raw variable whose mapping was destroyed. The recovery's success path now forces the pre-trim substitution into the pool state (`apSubSolvedPool`), symmetric with the returned value.

## The closed certificate

The full suite surfaced two regressions in the deposit-and-drop policy, both instances of one principle — a predicate may leave the flow only when nothing about it can still change:

- **Stuck type functions** (an acceptance flip on b1326): `propagateFunDeps` solves predicates in raw form, treating a stuck `SizeOf t` as an opaque constant — a strictly weaker goal than the definition-level solve faces after `expTConPred` splits it into the rewritten predicate plus its defining obligation. Deposit-and-drop meant the obligation was never created: a definition missing its `Bits` proviso compiled, then ICE'd in codegen.
- **Metavariables** (an ICE replacing a deliberate G0070 on ERulesMux2a): dropping a predicate with free unification variables removes it from every future join and defaulting set; a variable whose only carrier predicates were dropped is never improved, and the evidence carries the raw variable into emitted code.

The certificate for leaving the flow is now **closed**: no free unification variables (rigid signature-bound variables are pinned by their binder and do not count, which preserves the measured win on polymorphic library code), no type-function applications, and not already pooled. Symmetrically, a consult may only discharge a closed predicate. Qualifier checking hides the pool entirely, and error-reporting reductions run with the pool hidden (`withoutSolvedPool`), so messages show the complete residual story independent of solve history. Two more aggressive variants — open-predicate alias deposits and alias-dropping closed duplicates — caused emission-scope violations in the full suite and are disabled, with their mechanisms retained for a follow-up that solves cross-group evidence scoping.

## The pool-reuse guard judges modally

The pool-reuse guard's defer-to-given check mirrors the committable gate, but judged with the guarded unifier where the gate judges modally. A given from an enclosing frame quantifies over distinct rigid variables; the guarded judgment would hide it, letting a pooled dictionary discharge a predicate the given was meant to claim. It now judges modally — unifying with an empty bound set, the same modal judgment #1037 introduced for the committable gate.

## Measured impact

`newInst` calls on two representative designs, measured at the hardening commit (stdlib build time stays flat where measured): FloatingPoint.bsv **19,191 → 13,244** (−31.0%), AlignedFIFOs.bsv **2,323 → 2,173** (−6.5%). Unrestricted non-ground pooling measured −31.6% on FloatingPoint; the 0.6-point give-back from the closed-certificate hardening is entirely discharges that were unsound.

## Commit-by-commit

1. **Order SolvedBinds at emission; share solved dictionaries within satisfy** — the ordering change and intra-pass solved list (enabling substrate; −0.5% on its own).
2. **Pool solved dictionaries: propagateFunDeps no longer discards its work** — the ground-only pool, frames, closures, given-independence.
3. **Extend the solved-dictionary pool to non-ground predicates** — the three explicit mechanisms above plus the error-recovery interaction.
4. **Harden the solved-dictionary pool: closed certificate, plus diagnostics** — b1326/ERulesMux2a fixes, the closed certificate, pool-hidden error reporting.
5. **testsuite: solved-dictionary pool invariant pins** — `bsc.typechecker/dictpool`: frame scoping and emission scope stressed by nested and sibling bindings sharing closed dictionaries; given-independence pinned by a given-discharged and an instance-discharged use of the same class in one package.
6. **Judge the pool-reuse guard modally, like the committable gate it mirrors.**

## Validation

- Full `bsc.typechecker` sweep including the new pins passes clean.

**The stack**: #1032 (instance-trie fixes) and #1033 (coherence keywords + coverage warning), independent → #1035 (ordered-clause commitment) → #1036 (SAT settlement) → #1037 (bound-variable discipline) → **#1038 (solved-dictionary pool, this PR)**. Each PR contains the ones below it.


nanavati:
Heads-up for the eventual rebase onto current main: now that #923 is merged, `SolvedBinds` upstream has four additional fields (`incoherentIds`, `bindClasses`, `bindTypes`, `directIncoherences`). Two record constructions this PR adds build `SolvedBinds` explicitly and will need those fields, or the compiler dies at the first Prelude compile with

```
bsc: SolvedBinds.hs: Missing field in record construction incoherentIds
```

(GHC compiles incomplete record construction as a warning and crashes at evaluation.)

The fix that worked in a combined tree containing both this PR and #923's fields:

- `renameBinds` maps the new maps' keys through the rename:
  ```haskell
  incoherentIds = S.map ren (incoherentIds sbs),
  bindClasses = M.mapKeys ren (bindClasses sbs),
  bindTypes = M.mapKeys ren (bindTypes sbs),
  directIncoherences = M.mapKeys ren (directIncoherences sbs)
  ```
- `extractClosures` restricts the maps to the collected ids and is coherent by construction (its walk refuses closures through incoherent-marked bindings), so its incoherence fields are empty:
  ```haskell
  collected_ids = S.fromList [ i | ((i, _, _), _, _) <- collected ]
  ...
  incoherentIds = S.empty,
  directIncoherences = M.empty,
  bindClasses = M.restrictKeys (bindClasses sbs) collected_ids,
  bindTypes = M.restrictKeys (bindTypes sbs) collected_ids
  ```

nanavati:
Status update: we've dropped this PR's six pool commits from the MatX 2026.07 release line after a closer look at the evidence, and I'd like to park it here pending a refactor. It might be a win given that refactoring, but it isn't obviously one now. The reasoning, so it's on the record:

**The measured win is a proxy metric.** The `newInst` reductions (−31% FloatingPoint, −6.5% AlignedFIFOs) count calls, not cost, and aren't attributed — we don't know *which* derivations are being saved. The direct measurement in the PR description is "stdlib build time stays flat where measured," and the enabling ordering commit alone was −0.5%. That's consistent with the eliminated calls being concentrated in cheap leaf derivations.

**The implementation taxes exactly the designs it should help.** The pool is a pair of plain lists threaded through `satisfy'`; every `satisfyEx` entry applies the current substitution to the *entire* pool before the pass starts — including closed entries that are substitution-proof by the certificate's own definition — and deposit-side dedup is a linear scan with full structural equality after another whole-pool `apSub`. That's O(pool) per satisfy entry, growing monotonically across the reduction fixed point, i.e. superlinear overhead concentrated on proviso-heavy designs. It also retains every solved dictionary (extended to non-ground predicates) for the duration, and repeated lazy `apSub` over a growing list is a classic thunk-retention shape: our release-to-release measurements show ctxreduce-territory stress tests (b1490's MyBool, log2_loop) needing 288M→320M-class heap-cap raises with this stack in.

**The output-level benefit is subsumed downstream.** The artifact of re-derivation is duplicate non-recursive dictionary bindings, which #925's LiftDicts/FixupDefs lift and deduplicate by evidence fingerprint anyway. What the pool uniquely saves is typecheck-time CPU on the re-derivations — the quantity measured flat.

**The risk surface is real.** Two more-aggressive variants were disabled after emission-scope violations in the full suite, and the closed-certificate hardening gave back part of the win because those discharges were unsound. The safety story rests entirely on the reuse guard's correctness, and the failure mode on designs outside the suite's shapes is an internal error.

**What a viable revival looks like:** a ground-only pool keyed by class + canonical head in a persistent map (the closed certificate freezes keys, so incremental insert and frame snapshots come free), strict entry forcing, no substitution threading for closed entries, with the non-ground alias pool staying frame-local and small. Acceptance criteria should be wall-clock and max-residency on proviso-heavy designs, not `newInst` counts.

Everything beneath the pool boundary — the trie fixes, coherence commitment, single-session SAT settlement at the owner, and the skolem hardening (#1033–#1037) — is unaffected and stays in.

PR #1040, opened 2026-07-12
Add -remap-path-prefix: reproducible .bo/.ba output across build directories

nanavati:
## Problem

bsc bakes the invocation directory into every output file, so identical inputs compiled in different directories produce different bytes. Under remote execution (Bazel), every sandbox has a different absolute path, so nothing ever caches and outputs are non-deterministic by construction.

An audit of where machine-specific bytes enter the formats found:

- **Source positions**: every `Position` carries a pwd-encoded file name (`createEncodedFullFilePath` embeds the pwd *even for relative invocations*), shared once per file in the `.bo`/`.ba` stream.
- **Transitive cascade**: an importer's `.bo` embeds the importee's source path (via positions on imported `Id`s) *and* a content hash of the imported `.bo` (`ipkg_depends`), so one path difference at the leaves ripples through the entire package graph even where the raw string doesn't appear.
- **`.ba` extras**: `abmi_path` / `abmi_src_name` (and the schedule-error variants) store the elaboration directory directly, and the serialized `Flags` record stores `bdir`, `bluespecDir`, the search paths, output paths, and dump targets.
- Nothing else: there are no timestamps in either format (`abmi_time` is long gone), and the read-time hash is computed over payload bytes, so it becomes deterministic exactly when the bytes do. (Designs that use `date`/`epochTime`/`Environment` values remain inherently non-reproducible; that is orthogonal to paths.)

## Fix

A repeatable flag, `-remap-path-prefix FROM=TO` (after gcc's `-ffile-prefix-map` and rustc's `--remap-path-prefix`), applied at **serialization time**:

- Stored paths whose decoded form starts with `FROM` have it replaced by `TO`; the longest matching `FROM` wins; unmatched paths are stored untouched. Decoding the internal `///` pwd marker before matching means a match also normalizes away the relative-vs-absolute invocation difference.
- Positions are remapped at the single sharing chokepoint in `BinData` (`share`), before sharing, so the cache is keyed on the stored value: positions that remap equal share one payload and the stream is canonical.
- `.ba` files additionally remap `abmi_path`/`abmi_src_name` (+ `abmsei_*`) and normalize the stored `Flags` copy (which is consumed only by bluetcl introspection) — including clearing `remapPathPrefix` itself, whose `FROM` values are machine-specific by construction.
- Diagnostics that print positions read back from imported `.bo` files show the remapped (e.g. workspace-relative) form, which keeps them readable.

**Guarantees** (all verified empirically; `testsuite/bsc.options/remap-path` pins them):

- With `-remap-path-prefix "$PWD=."`, the same sources compiled from two different directories produce **byte-identical** `.bo` and `.ba`, with no residual build-directory strings, including through an import (`-u`) and an elaboration (`-g`).
- Re-running a compile in place is byte-identical (no hidden nondeterminism).
- With **no** flag given, output bytes are **unchanged** from the previous compiler (verified by byte-comparison), so existing flows and caches are unaffected.

The `.ba` format tag advances for the new `Flags` field; the `.bo` format is unchanged.

## Usage under Bazel

Pass `-remap-path-prefix "$PWD=."` (or `=<workspace-relative package dir>`) on every compile; if the toolchain itself lives inside the sandbox, add a second mapping for its prefix. Generated Verilog headers additionally want the existing `-no-show-timestamps`.

### DIFF
--- a/src/comp/Flags.hs
+++ b/src/comp/Flags.hs
@@ -363,5 +363,6 @@ remapFlagsPaths remap flags =
             vdir = rM (vdir flags),
             vPathRaw = rL (vPathRaw flags),
             vPath = rL (vPath flags),
-            dumps = [ (d, rM mf) | (d, mf) <- dumps flags ]
+            dumps = [ (d, rM mf) | (d, mf) <- dumps flags ],
+            dumpAll = fmap rM (dumpAll flags)
         }
--- a/src/comp/Flags.hs
+++ b/src/comp/Flags.hs
@@ -1,5 +1,6 @@
 module Flags(
              Flags(..),
+             remapFlagsPaths,
              redSteps,
              ResourceFlag(..), SATFlag(..), MsgListFlag(..),
 
@@ -44,6 +45,11 @@ data Flags = Flags {
         doICheck :: Bool,
         dumpAll :: Maybe (Maybe FilePath), -- maybe dump to file or stdout
         dumps :: [(DumpFlag, Maybe FilePath)], -- dump to file or stdout
+        -- prefix remappings applied to file paths stored in .bo/.ba
+        -- files (source positions, the .ba module path fields, and the
+        -- path-valued fields of this record as stored in .ba), so that
+        -- outputs do not depend on the build machine's directory layout
+        remapPathPrefix :: [(String, String)],
         enablePoisonPills :: Bool,
         entry :: Maybe String,
         expandATSlimit :: Int,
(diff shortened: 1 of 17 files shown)


src/comp/Flags.hs
```diff
@@ -326,3 +332,37 @@ dumpInfo :: Flags -> DumpFlag -> Maybe (Maybe FilePath)
 dumpInfo f d = lookup d (dumps f) <|> dumpAll f
 
 -- -------------------------
+
+-- Apply the -remap-path-prefix mappings to the path-valued fields of
+-- the record itself.  Used on the copy of Flags stored in .ba files
+-- (which is consumed only by bluetcl introspection), so that .ba
+-- bytes do not depend on the build machine's directory layout.  The
+-- remapping helper lives in FileNameUtil; it is passed in here to
+-- avoid an import cycle.
```
matx-amy, line 341: i don't see an import cycle?

matx-amy:
- Should the `///` syntax be documented somewhere?

other than that, reviewed, :+1:

don't love the need to manually traverse things to find the paths, but i don't love taking on a dep on syb or something either

PR #1041, opened 2026-07-12
CI: gate .bo/.ba format changes on their header-tag bumps

nanavati:
The `.bo` and `.ba` readers decide file compatibility solely by the header tag strings in `src/comp/GenBin.hs` and `src/comp/GenABin.hs`. When a PR changes the serialized format but forgets the tag bump, a same-tag reader misparses stale files and dies with an `internalError` ("please report") instead of the intended `EBinFileVerMismatch` "recompile" error. Recent history shows this slip is easy to make — several open PRs changed `Bin` instances with the tag untouched, and I made the same mistake myself last week.

This adds a cheap, build-free CI job (only triggered when the three files change):

- a PR touching `src/comp/GenBin.hs` must change the `.bo` tag;
- touching `src/comp/GenABin.hs` must change the `.ba` tag;
- touching `src/comp/BinData.hs` (the `Bin` instances shared by both formats) must change **both**.

Not every change to these files alters serialized bytes (comments, hash plumbing, dump code), so there is an auditable escape hatch: a commit trailer `Bin-format: bo-unchanged` / `Bin-format: ba-unchanged` asserting the change is byte-neutral. The trailer lives in history, so a reviewer can hold the commit to its claim.

Validated against real ranges before opening this PR:
- a PR range that changed `GenABin.hs` without a bump → **fails** with the actionable message;
- a PR range that bumped both tags → **passes**;
- a byte-neutral `BinData.hs` change (empirically verified bit-identical output) with a bumped `.ba` but unchanged `.bo` tag → fails on `.bo` until the trailer is added — the intended behavior for that case.

The gate deliberately does not touch the gated files itself. Note for existing open PRs: a few will trip this on their next push until they bump tags (correctly) or add trailers — that is the gate doing its job.

### DIFF
--- a/.github/scripts/check_bin_format_tags.sh
+++ b/.github/scripts/check_bin_format_tags.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/env bash
+# Gate: changes to the binary-format code must bump the format tags.
+#
+# The .bo and .ba readers decide compatibility solely by the header tag
+# strings in src/comp/GenBin.hs and src/comp/GenABin.hs.  A format
+# change without a tag bump makes a same-tag reader misparse stale
+# files and crash with an internalError instead of the intended
+# "recompile" error -- a mistake that has now been made independently
+# by several PRs.  This check makes the tag bump hard to forget:
+#
+#   - a change to src/comp/GenBin.hs  requires the .bo tag to change
+#   - a change to src/comp/GenABin.hs requires the .ba tag to change
+#   - a change to src/comp/BinData.hs (Bin instances shared by both
+#     formats) requires BOTH tags to change
+#
+# Not every change to these files alters serialized bytes (comments,
+# the hash plumbing, dump/trace code).  For those, add a commit
+# trailer to any commit in the range:
+#
+#   Bin-format: bo-unchanged
+#   Bin-format: ba-unchanged
+#
+# asserting that the change provably does not alter the bytes of that
+# format.  The trailer is part of the audit trail: a reviewer can hold
+# the commit to its claim.
+#
+# Usage: check_bin_format_tags.sh <base-sha> <head-sha>
+
+set -eu
+
+BASE=$1
+HEAD=$2
+
+changed() {
+    ! git diff --quiet "$BASE" "$HEAD" -- "$1"
+}
+
+tag_in() {
+    # first bsc-bo-* / bsc-ba-* string in the file at the given commit
+    git show "$1:$2" 2>/dev/null \
+        | sed -n 's/.*T\.pack "\(bsc-b[oa]-[^"]*\)".*/\1/p' \
+        | head -1
+}
+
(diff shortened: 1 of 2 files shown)


.github/scripts/check_bin_format_tags.sh
```diff
@@ -0,0 +1,90 @@
...
+# by several PRs.  This check makes the tag bump hard to forget:
+#
+#   - a change to src/comp/GenBin.hs  requires the .bo tag to change
+#   - a change to src/comp/GenABin.hs requires the .ba tag to change
+#   - a change to src/comp/BinData.hs (Bin instances shared by both
+#     formats) requires BOTH tags to change
+#
+# Not every change to these files alters serialized bytes (comments,
+# the hash plumbing, dump/trace code).  For those, add a commit
+# trailer to any commit in the range:
```
quark17, line 18: It seems dangerous for one commit to be able to exempt other commits from the check.  I thought maybe you were doing this to allow someone to just push a commit to a PR, rather than force-push a change to an earlier commit -- but they couldn't push an empty commit just for its message, so I think that's OK.

.github/scripts/check_bin_format_tags.sh
```diff
@@ -0,0 +1,90 @@
...
+# files and crash with an internalError instead of the intended
+# "recompile" error -- a mistake that has now been made independently
+# by several PRs.  This check makes the tag bump hard to forget:
+#
+#   - a change to src/comp/GenBin.hs  requires the .bo tag to change
+#   - a change to src/comp/GenABin.hs requires the .ba tag to change
+#   - a change to src/comp/BinData.hs (Bin instances shared by both
+#     formats) requires BOTH tags to change
+#
+# Not every change to these files alters serialized bytes (comments,
```
quark17, line 16: These are also not the only files that would require a bump, right?  I think some types have their format generated using generics, and so merely changing the ISyntax or ASyntax could require a bump?  But this at least catches a lot of differences.

.github/workflows/bin-format-gate.yml
```diff
@@ -0,0 +1,26 @@
...
+# override trailer for byte-neutral changes.
+
+on:
+  pull_request:
+    paths:
+      - 'src/comp/BinData.hs'
+      - 'src/comp/GenBin.hs'
+      - 'src/comp/GenABin.hs'
+
+jobs:
```
quark17, line 15: My preference would be to include this in `ci.yml` alongside the cleanliness checks, so that all of the checks are on one page and don't require looking elsewhere (assuming you know to even look for another page) (and it's particularly bad on the GitHub phone app).

quark17 (commented):
This is good to check!  Just a few comments.

(I also wouldn't mind a check for `Error.getErrorText` that there are no tag clashes and that the clauses are in tag order.)

PR #1043, opened 2026-07-13
testsuite: working time-ordered parallel scheduling, with guards against empty test runs

nanavati:
Two related pieces:

**1. Make `RUN_TESTCASES_IN_ORDER_OF_TIME=1` actually work.** The plumbing existed but had rotted: the tool prefix wasn't defaulted, `timing.txt` was never produced or refreshed, and `sort-by-time.pl` fell over on the file it was supposed to consume. With this series, the suite schedules directories longest-first (LPT), which materially cuts wall-clock on parallel runs (`make -jN checkparallel/fullparallel`) — the win grows with core count, since the long tail no longer lands last. `timing.txt` is refreshed at the end of each run (new times win, old entries are preserved, never committed).

**2. Never produce an empty green test run.** While making (1) work we found and closed three paths where scheduling failures silently ran ZERO tests and exited green — the same silent-coverage-loss class as the parallel-make race fixed in #1006:
- an empty or malformed `timing.txt` died inside an unguarded pipeline (awk's exit status wins), yielding an empty `ALL_TESTS`;
- `generate-stats` could install an empty `timing.txt` when a run produced no `time.out` files (deterministic on macOS, where the time wrapper never writes one) — self-arming the failure for the next run;
- from a test subdirectory, the tool-prefix filter matched nothing, so subdir `checkparallel` became a green no-op.

The fixes: `sort-by-time.pl` warns and falls back to the unsorted (or unfiltered-minus-harness-files) list instead of dying; `run-tests-setup` hard-fails if `all_tests.mk` ends up with no `.exp` entries — that one check closes the whole class regardless of cause; `generate-stats` only installs a non-empty timing file.

Motivation for upstreaming now: faster regression turnaround benefits everyone, and it compounds with the Verilator testsuite legs (#1004) where each leg is a full suite run. Scheduling details (LPT vs alternatives, median placement for unknown tests) are very much open to debate — the guards in (2) are the part I'd argue is unconditionally wanted.

https://claude.ai/code/session_019cqQdWKp7jr73GHQYnvhiq

### DIFF
--- a/testsuite/scripts/sort-by-time.pl
+++ b/testsuite/scripts/sort-by-time.pl
@@ -48,7 +48,8 @@ $median_time=$alltimes[int((scalar@alltimes)/2)];
 
 
 #print STDERR "median_time $median_time\n";
-open FI,"find . -name '*.exp' | grep '^\\./$ARGV[0]\\.' | "
+my $tool_pat = "^\\./$ARGV[0]" . "[/.]";
+open FI,"find . -name '*.exp' | grep '$tool_pat' | "
   or die "finding exp files failed";
 
 #print STDERR "reading lines\n";
@@ -56,17 +57,22 @@ while(<FI>){
   push @lines,$_
 }
 
-# From a test subdirectory the tool prefix matches nothing (paths are
-# ./<dir>/... only at the top level); list everything under the current
-# directory instead of silently producing an empty schedule.
+# From a test subdirectory the DEFAULT tool prefix matches nothing
+# (paths are ./<dir>/... only at the top level); list everything under
+# the current directory instead of silently producing an empty
+# schedule.  Only the default tool falls back: an explicit tool that
+# matches nothing is a caller error, and running every test instead
+# would be far worse than the loud empty-list failure downstream.
 unless (@lines) {
-  print STDERR "no .exp files match ./$ARGV[0].*; listing all .exp files instead\n";
-  # exclude the harness's own .exp files (site.exp, config/, lib/), which
-  # live at the testsuite top level and are not tests
-  open FI,"find . -name '*.exp' | grep -v -e '^\\./site\\.exp\$' -e '^\\./config/' -e '^\\./lib/' |"
-    or die "finding exp files failed";
-  while(<FI>){
-    push @lines,$_
+  if ($ARGV[0] eq 'bsc') {
+    print STDERR "no .exp files match ./$ARGV[0].*; listing all .exp files instead\n";
+    # exclude the harness's own .exp files (site.exp, config/, lib/), which
+    # live at the testsuite top level and are not tests
+    open FI,"find . -name '*.exp' | grep -v -e '^\\./site\\.exp\$' -e '^\\./config/' -e '^\\./lib/' |"
+      or die "finding exp files failed";
+    while(<FI>){
+      push @lines,$_
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


testsuite/scripts/sort-by-time.pl
```diff
@@ -25,13 +24,22 @@
 #find . -name time.out -exec cat '{}' \; | perl scripts/times.by-directory.pl
 
 #print STDERR "reading time\n";
+# A bad timing file must never abort list generation: an aborted (or
```
quark17, line 27: I'm confused by this comment, as it seems to be mixing two things.  Further down, the new `unless (@alltimes)` does in fact fall back to the simple output if the file is empty.  However, the `unless` that deals with malformed lines (just below this comment) has been changed, so it no longer dies and leads to an empty file and the fallback to simple output, instead the line is now silently ignored!  That means that a malformed file will indeed lead to tests being not run -- though I guess it's not "silent" in that a warning message is reported (if anyone notices it among all the testsuite output?)

testsuite/suitemake.mk
```diff
@@ -119,6 +119,13 @@ LOCALCHKCMD ?= $(RUNTESTENV) $(RUNTEST) $(RUNTESTFLAGS) *.exp
 localcheck: $(LOCALCHECKPREREQUISITES)
 	$(LOCALCHKCMD)
 
+# The tool whose tests are collected by 'run-tests-setup'.  This must be
+# non-empty so that the generated 'all_tests.mk' is limited to test
+# directories (and can be sorted by time); with an empty value, stray
+# non-test .exp files (config/unix.exp, site.exp, lib/bsc.exp) are also
+# treated as tests.
+tool ?= bsc
```
quark17, line 127: One: you added new code to `sort_by_time.pl` that excludes `config/` and `site/` and `lib/`, so is this assigning to `tool` still needed?

Two: Does this break the `check` target when called in a subdirectory?

Earlier in the file is this line:
```
## do not put --tool bsc here, since that will limit recursion into local directories
RUNTESTFLAGS ?= --tool ""
```
This is for the `check` target, which can be called in a subdirectory and will recurse into its subdirectories.

I assume that `tool ?= bsc` doesn't break that behavior?

(I notice that `RUNTESTFLAGS` is also used by the `localcheck` target, which doesn't recurse, so technically doesn't need `-tool ""`.)

quark17: Also, I notice that the `fullparallel` target (but not `checkparallel`) hardcodes `tool=bsc` (and the top `Makefile` also does in places) -- this is all kind of a mess.

I assume that you need to set `tool` here only because it's referenced in the `run-tests-setup` target, in the argument to `sort-by-time.pl`?  Which I guess could be set to `bsc` some other way, if needed.

testsuite/suitemake.mk
```diff
@@ -148,6 +160,24 @@ generate-stats:
...
 	@echo "=== Brief list of results ==="
 	@find . -name '*.sum' | sort | perl $(CONFDIR)/scripts/process-summary-file.pl
+	@echo ""
+	@echo "Refreshing timing.txt (used for time-ordered scheduling when"
+	@echo "RUN_TESTCASES_IN_ORDER_OF_TIME=1 is set)"
+	@find . -name time.out -exec cat '{}' \; \
+		| perl $(CONFDIR)/scripts/times-by-directory.pl \
+		> $(CONFDIR)/timing.txt.new
+	@if [ -f $(CONFDIR)/timing.txt ]; then \
+		awk '!seen[$$2]++' $(CONFDIR)/timing.txt.new $(CONFDIR)/timing.txt \
```
quark17, line 170: Why are we merging and not overwriting the stale file?

testsuite/scripts/sort-by-time.pl
```diff
@@ -40,14 +48,34 @@
...
 
 #print STDERR "reading lines\n";
 while(<FI>){
   push @lines,$_
 }
 
+# From a test subdirectory the DEFAULT tool prefix matches nothing
+# (paths are ./<dir>/... only at the top level); list everything under
+# the current directory instead of silently producing an empty
+# schedule.  Only the default tool falls back: an explicit tool that
```
quark17, line 63: Why do you need to handle the case of a subdirectory?  Can `checkparallel` be called in a subdirectory?  (That'd be cool -- I'd only ever considered it a target from the top-level and never tried in a subdirectory.)

quark17 (commented):
This is cool.  I don't recall even knowing about this feature, so good find!

PR #1045, opened 2026-07-14
ISyntax: replace SYB traversals with explicit ones

nanavati:
Replaces the two `Data.Generics` (syb) traversals over ISyntax with explicit ones, and drops the now-unused `Data`/`Typeable` derivings from ISyntax and the evaluator heap types.

## Motivation

Generic traversals over a compiler IR fail in the worst way available: when a constructor or field is added, `everywhere`/`listify` silently skip it rather than failing to compile. The two ISyntax users are rewritten as explicit traversals with exhaustive constructor matches, so any future ISyntax change surfaces at the traversal site as an incomplete-pattern warning. This also removes the obligation to keep `Data` instances over types embedding IORef-backed heap references, and leaves ISyntax free to evolve its representation without generic-traversal compatibility as a constraint.

## Changes

- **`IExpand.removeInlinedPositions`**: the `everywhere` rewrite becomes an explicit traversal. It must remain lazy — clock/reset/inout wires are cyclic through `IStateVar`, and only lazy unrolling terminates (now documented at the site). One behavior made explicit rather than incidental: the `Pred` set is rebuilt with `S.map`, since `PTerm`'s `Ord` consults inlined positions and removing them can reorder or collapse elements (this matches the old syb-era rebuild through `fromList`).
- **`bluetcl.find_vmodinfo`**: `listify` becomes a direct query, preserving left-to-right, duplicate-keeping order. The site now documents the invariant that makes the direct query complete: VModInfo reaches an `IPackage` only through `ICVerilog` — the synthesis boundary re-abstracts every synthesized module as a Verilog wrapper (the same representation as a hand-written `import "BVI"`), and the richer elaboration products never re-enter package vocabulary.
- **`ISyntax.hs` / `IExpandUtils.hs`**: drop all `Data`/`Typeable` deriving sites (16 in ISyntax, plus `PExpr`/`HeapCell`/`HeapData`), the now-dead `DeriveDataTypeable` pragmas, and a stale commented-out `constrIndex` alternative. The remaining syb users in `src/comp` traverse Verilog syntax only.

## Validation

- Full build clean (`install-src`, `install-extra`), typecheck closures for both `bsc` and `bluetcl`.
- Testsuite, clean + localcheck on the exercising directories: **515 PASS / 0 FAIL**, including `bsc.misc/method_conditions` 134/0 (the end-to-end exerciser of `-keep-method-conds`, the flag gating `removeInlinedPositions`), `bsc.showrules` 48/0, and the `bsc.bluetcl` directories.
- Behavioral identity spot-check: `sysMethodConds_RegWrite.bluetcl-out` byte-identical to its golden — position tuples are exactly the data this transformation shapes.

https://claude.ai/code/session_01LVL9ornS6uf9hVxAuL7GhK

### DIFF
--- a/src/comp/IExpand.hs
+++ b/src/comp/IExpand.hs
@@ -28,7 +28,6 @@ import Control.Monad.Fix(mfix)
 --import Control.Monad.Fix
 import Control.Monad.State(State, evalState, liftIO, get, put)
 import Data.Graph
-import qualified Data.Generics as Generic
 import System.IO(Handle, BufferMode(..), IOMode(..), stdout, stderr,
                  hSetBuffering, hIsOpen, hIsClosed)
 import System.FilePath(isRelative)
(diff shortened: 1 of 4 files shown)


quark17 (dismissed):
This looks good.  I wonder if deriving can be removed from more files, then?  For example, the deriving in `IStateLoc.hs`, `IType.hs`, `VModInfo.hs`, `Wires.hs`, etc may only have been to support the deriving in ISyntax?

Also, tangentially, PR #722 was interested in removing the `syb` dependency by at least changing the imports to `Data.Data`.

PR #1058, opened 2026-07-18
Semantic port properties: derive from the APackage, feed every backend

nanavati:
## Summary

Port properties (`clock`, `reset`, `reg`, `const`, `unused`, `inhigh`,
...) are now derived by `getIOPropsA`, a semantic analysis of the
`APackage` and the schedule, instead of being measured off the optimized
netlist at the end of the Verilog backend. A property is asserted only
when it is entailed by the design's structure, dataflow, and schedule —
so the answers are stable across optimization settings, inlining
settings, and backends. The results feed the import-BVI wrapper
attributes in the `.bo` for **both** backends (Bluesim compiles gain
port properties, closing the long-standing `XXX` in `bsc.hs`) and the
Verilog "Ports:" comment. The old `getIOProps` remains available for
comparison behind `-dIOproperties`, computed only when that dump is
requested.

A design document (`doc/proposals/port-properties.md`) records the
property definitions, the soundness/stability contracts, the validation
evidence, and the migration plan (this PR is steps 1–3; steps 4–5 retire
`getIOProps` after a release of coexistence).

**Stacked on #1057**, whose two commits this branch contains; the
commits under review here begin at "Add getIOPropsA". This PR can be
rebased to drop them once #1057 merges.

## Why

The measured properties are functions of a particular compile, not of
the design: an unused input clock loses its `clock` label because no
surviving netlist connection witnesses it; `-no-inline-rwire` makes a
value read through a bypass wire lose its `reg` label although the
connectivity is unchanged; a ready can be `const` only because of a CSE
rewrite. The properties are recorded in the `.bo` and consumed by
parent compiles (`inhigh` legality, readiness reasoning), so the
instability propagates up the hierarchy.

## How it works

- Ports are enumerated exactly as `AState` constructs them.
- Structural roles (clock/gate/reset/inout, declared props) come from
  the wire/field info and are always present — an unused input clock is
  `clock unused`.
- `reg`/`const`/`unused` are deduced by dataflow: wire and CReg
  instances are looked through regardless of the inlining flags; a
  memoized evaluator folds schedule-time constants (never/always-firing
  WILL_FIREs, wire validity, complementary conditions); and the argument
  muxes of state-instance methods are modeled with AState's own port
  allocation, exclusivity test, and order — for action methods and value
  methods (RDY-based selectors for interface value-method users). The
  references AState itself creates for a caller's WILL_FIRE are folded
  semantically: enables absorbed by constant or complementary conjuncts,
  selectors of direct connections, losing arms, and last mux arms
  (absorbed into don't-care defaults).
- Merges are crossed by agreement, not enumeration: equal expressions
  collapse; value sets are never enumerated. This boundary is what makes
  the answers stable (see the design document).

## Rebase notes (onto current `main` + #1057)

- The two let-generalization commits the branch previously carried
  landed upstream as PR 902 and were dropped from the series.
- Port splitting (PR 849) landed while this branch was in flight; the
  series ends with an adaptation commit ("Adapt getIOPropsA to split
  method ports") migrating the pass to per-port method outputs
  (`vf_outputs`), per-argument input port groups (`vf_inputs`), and the
  `ATuple`/`ATupleSel` expression forms, mirroring `AState`'s own
  handling. It includes a regression test (`APkgProps_SplitSel`) that a
  selected result port takes the selected port's declared property
  (`ATupleSel` indices are 1-based). One known conservative case: a
  split argument reached through a CSE'd tuple-typed definition falls
  back to an opaque use (live, no properties asserted) rather than
  pairing per-port declared properties through the definition.
- `bsc.bugs/bluespec_inc/b302/mkDesign.v.expected` is regenerated: its
  comparison is dormant without a Verilog simulator, so it was not
  regenerated upstream when port splitting changed the `__h` name
  numbering (verified against a pre-split compiler).

## Validation

- Sweep of all 2124 code-generating testsuite designs (~18,300 port
  lines) comparing both analyses: identical port enumeration everywhere;
  1,686 lines differ, of which 1,638 are the richer structural labels,
  15 are strictly more accurate (dead-logic `unused`, CReg port-0 `reg`,
  a register repacked through an identity case), and 33 are the
  documented boolean-minimization non-goal. No line asserts a property
  `getIOProps` contradicts.
- Stability demonstrated under `-no-inline-rwire`, `-no-inline-creg`,
  `-keep-fires`, where the netlist measurement weakens.
- `testsuite/bsc.verilog/portprops`: golden tests dumping both analyses
  side by side cover each deduction mechanism, including new tests for
  value-method argument muxes (`APkgProps_VMux`) and enable/selector
  folding (`APkgProps_EnFold`).
- Full dejagnu testsuite after the rebase, with a Verilog simulator
  configured (which also enables the otherwise-dormant `compare_verilog`
  golden checks): **20313 PASS / 0 FAIL / 134 XFAIL / 0 XPASS**,
  including `bsc.verilog/portprops` 85/85.
- Cross-backend: a child compiled with `-sim` and a parent with
  `-verilog` — the parent deduces its input `unused` from the child's
  `.bo`.
- Tractability: memoized per definition; no measurable overhead on the
  largest testsuite design (h264 deblocking filter).

## Questions for reviewers

1. Is `doc/proposals/` an acceptable home for the design document?
2. Naming: `-dAPackageIOproperties` / `getIOPropsA`.
3. Timeline for migration steps 4–5 (retiring `getIOProps`).

https://claude.ai/code/session_016sDiGGpNpEup2huNG7LYW7


nanavati:
Superseded by #1060 — same commits, but stacked on the inout-fix branch (#1059) so the diff shows only the semantic-port-properties work.

PR #1076, opened 2026-08-07
Release notes 2026.07

quark17:
Draft of release notes for pending release.  We're still waiting on a PR to add documentation for SplitPorts, and that PR number will need to be added in the notes.

### DIFF
--- a/release/ReleaseNotes.adoc
+++ b/release/ReleaseNotes.adoc
@@ -4,6 +4,258 @@ Bluespec Compiler (BSC) Release Notes
 :last-update-label!:
 :nofooter:
 
+2026.07 Release
+---------------
+
+Blurb here lorem ipsum
+... mention backward incompatible changes (flags on by default,
+let-gen, bluetcl ports, etc?) ...
+... mention new features (ATF, port splitting) ...
+... mention performance improvements, including in particular PR#1007,
+which has been a long-standing elaboration issue that now dramatically
+compiles faster; as well as #919 (loop test that needs less memory);
+and #957 that improves scaling for large modules ...
+
+Changes since release 2026.01:
+
+Documentation
+~~~~~~~~~~~~~
+
+* Updates to `README.md`
+  ** Add Fedora and Homebrew to the list of systems with packaged BSC
+     (GitHub PR#971, PR#972)
+
+* Updates to `INSTALL.md`
+  ** Mention the third-party Terra repository as a source for the
+     `ghc-strict-concurrency` dependency when building on Fedora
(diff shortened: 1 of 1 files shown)


release/ReleaseNotes.adoc
```diff
@@ -4,6 +4,284 @@ Bluespec Compiler (BSC) Release Notes
...
+  (GitHub PR#931)
+
+* Add an `Ord` instance for `String` comparison (GitHub PR#931)
+
+* Add `Rep` associated type function for `Generic` (GitHub PR#908)
+
+* Add `PrintType` library for constructing type-level string
+  representations for types (GitHub PR#958)
+
+* Add `fromChucks` (inverse of `toChunks`) to the `Vector` library
```
rsnikhil:
```suggestion
* Add `fromChunks` (inverse of `toChunks`) to the `Vector` library
```

rsnikhil:
The Release Notes look ok to me (but I did not cross-check that the PR references were accurate).
This is a very impressive release!

kenta2:
> This release enables the `-aggressive-conditions` and `-sched-conditions` flags by default, each of which can be disabled by providing the flag with `-no` prefixed.

Based on the comment at https://github.com/B-Lang-org/bsc/pull/966#issue-4606295867 , it seems the flag is `-no-aggressive-conditions` and not `-noaggressive-conditions`, so the prefix is `-no-` not `-no`.   I recommend explicitly spelling it out explicitly for users who want to deploy a quick fix without thinking too hard what the prefix is:

This release enables the `-aggressive-conditions` and `-sched-conditions` flags by default, each of which can be disabled by `-no-aggressive-conditions` and `-no-sched-conditions` respectively.

quark17:
Spelling it out is a good suggestion, I'll make that change.

As an aside, I notice that the `-help` message words it like this, which is less error prone:
```
Most flags may be preceded by a `no-' to reverse the effect.
```
(My sense was that literally adding `-no` on the front of `-aggressive-conditions` would give you `-no-agressive-conditions`, but I can see how the wording is confusing if one views the "flag" as only the part after the initial hyphen.  Using `no-` makes more sense from that viewpoint, and may be what you meant to say; `-no-` seems wrong either way, as it has too many hyphens.  But, anyway, spelling it out makes the most sense.  I was trying to hard to be concise.)

kenta2:
It's interesting how I considered the initial dash of a flag as semantically or mentally different from the hyphens separating words inside a flag even though they are the same ASCII character, something I never thought about until now; maybe something Noam Chomsky and friends think about all the time.

PR #1079, opened 2026-08-08
Updated libraries ref guide for SplitPorts

rsnikhil:
Summary of updates to libraries doc for SplitPorts:

                2 The Standard Prelude package
                2.1 Type classes
                    NEW: added "SplitPorts" line to table in opening para
                    NEW: added section: "2.1.16 SplitPorts"
                2.1 Data Types
                    NEW: added section "2.2.21 Port"
                3 Standard Libraries
                    NEW: added section: "3.11 SplitPorts"
                Index: NEW:   added entries for SplitPorts and relateds
                Index of Typeclasses:  NEW:  added entries for SplitPorts and relateds

### DIFF
--- a/doc/libraries_ref_guide/LibDoc/Prelude.tex
+++ b/doc/libraries_ref_guide/LibDoc/Prelude.tex
@@ -4398,7 +4398,7 @@ function Rules rJoinConflictFree(Rules x, Rules y);
 \label{sec:SplitPorts_Port_type}
 
 \index{Port@\te{Port} (related type for class \te{SplitPorts})}
-\index{SplitPorts@\te{SplitPorts}!\te{Port} (related type for class \te{SplitPorts})}
+\index{SplitPorts@\te{SplitPorts}!Port@\te{Port} (related type for class \te{SplitPorts})}
 
 The \te{Port} type is used in definitions of instances of the
 \te{SplitPorts} type class (Section~\ref{sec:SplitPorts_typeclass}).
--- a/doc/libraries_ref_guide/LibDoc/Prelude.tex
+++ b/doc/libraries_ref_guide/LibDoc/Prelude.tex
@@ -1770,12 +1770,41 @@ element of  data type \te{data\_t}\\
 
 \subsubsection{SplitPorts}
 
-\label{sec-SplitPorts-typeclass}
+\label{sec:SplitPorts_typeclass}
 
-\index{SplitPorts@\te{SplitPorts}!\te{SplitPorts} type class}
+% Indexing plan
+%                                           Main Index         Fn Index  TCs Index
+%                                      --------------------    --------  ---------
+%                                      Direct  Under  Under
+%                                              Pkg    Class
+% 1 SplitPorts       class                Y                                  Y
+% 2   PortsOf          class type         Y             Y
+% 3   splitPorts       class function     Y             Y          Y
+% 4   unsplitPorts     class function     Y             Y          Y
+% 5   portNames        class function     Y             Y          Y
+
+% 1
+\index{SplitPorts@\te{SplitPorts}!type class in Prelude}
 \index[typeclass]{SplitPorts}
+
+% 2
+\index{PortsOf@\te{PortsOf}!type in class \te{SplitPorts}}
+\index{PortsOf@\te{PortsOf}|seealso{\te{SplitPorts}}}
+\index{SplitPorts@\te{SplitPorts}!PortsOf@\te{PortsOf} (type in class \te{SplitPorts})}
+
+% 3
+\index{splitPorts@\te{splitPorts}!function in class \te{SplitPorts}}
+\index{SplitPorts@\te{SplitPorts}!splitPorts@\te{splitPorts} (function in class \te{SplitPorts})}
 \index[function]{Prelude!splitPorts}
+
+% 4
+\index{unsplitPorts@\te{unsplitPorts}!function in class \te{SplitPorts}}
+\index{SplitPorts@\te{SplitPorts}!unsplitPorts@\te{unsplitPorts} (function in class \te{SplitPorts})}
 \index[function]{Prelude!unsplitPorts}
+
+% 5
+\index{portNames@\te{portNames}!function in class \te{SplitPorts}}
+\index{SplitPorts@\te{SplitPorts}!portNames@\te{portNames} (function in class \te{SplitPorts})}
 \index[function]{Prelude!portNames}
 
 NOTE: \hfill \fbox{\begin{minipage}{0.9\textwidth}\footnotesize
@@ -1787,7 +1816,7 @@ The {\tt SplitPorts} type class is used to improve the
 readability/visibility of \emph{bsc}-generated Verilog code
 (especially the corresponding waveforms).  Additionally, it can make
 it easier to interface \emph{bsc}-generated Verilog to external
-Verilog IP.
+Verilog IP, and may be useful for other downstream Verilog tools.
 
 Consider an interface method argument or result whose type is a
 user-defined type (a struct, or tagged union, or enum).  These
@@ -1798,7 +1827,7 @@ collection of ports (for example, one port per struct field).
 
 NOTE: \hfill \fbox{\begin{minipage}{0.9\textwidth}\footnotesize
 
-Please see Section~\ref{sec-SplitPorts-package} which describes the
+Please see Section~\ref{sec:SplitPorts_package} which describes the
 use of the {\tt SplitPorts} type class in detail, including several
 concrete examples. That section also describes additional facilities
 available by importing the {\tt SplitPorts} and {\tt
@@ -1833,19 +1862,19 @@ It can be used by programs in either syntax.
 
 {\BSVb} syntax:
 \begin{libverbatim}
-   typeclass SplitPorts #(a, p)
+   typeclass SplitPorts #(type a, type p)
      dependencies (a determines p);
-     typedef p (PortsOf #(a));
+     type PortsOf#(type a) = p;
 
-     -- Convert a value to a tuple of values corresponding to ports.
-     function p splitPorts (a);
+     // Convert a value to a tuple of values corresponding to ports.
+     function p splitPorts (a x);
 
-     -- Combine a tuple of values corresponding to ports into a value.
-     function a unsplitPorts (p);
+     // Combine a tuple of values corresponding to ports into a value.
+     function a unsplitPorts (p x);
 
-     -- Compute the list of port names for a type, given a base name.
-     -- This must be the same length as the tuple of values.
-     fun List #(String) portNames (a x, String s);
+     // Compute the list of port names for a type, given a base name.
+     // This must be the same length as the tuple of values.
+     function List #(String) portNames (a x, String s);
    endtypeclass
 \end{libverbatim}
 
@@ -1860,13 +1889,6 @@ This is the central type class for port splitting.
 \end{itemize}
 
 \begin{center}
-
-\index{SplitPorts@\te{SplitPorts}!\te{splitPorts}   (function in type class)}
-\index{SplitPorts@\te{SplitPorts}!\te{unsplitPorts} (function in type class)}
-\index{SplitPorts@\te{SplitPorts}!\te{portNames}    (function in type class)}
-
-\index{unplitPorts@\te{unsplitPorts} (function in \te{SplitPorts} type class)}
-\index{portNames@\te{portNames} (function in \te{SplitPorts} type class)}
 
 \begin{tabular}{|p{1 in}|p{4in}|}
 \hline
@@ -4373,13 +4395,13 @@ function Rules rJoinConflictFree(Rules x, Rules y);
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \com{Port}
 \subsubsection{Port}
-\label{sec-SplitPorts-Port-type}
+\label{sec:SplitPorts_Port_type}
 
-\index{Port@\te{Port} (type, for \te{SplitPorts})}
-\index{SplitPorts@\te{SplitPorts}!\te{Port} type}
+\index{Port@\te{Port} (related type for class \te{SplitPorts})}
+\index{SplitPorts@\te{SplitPorts}!\te{Port} (related type for class \te{SplitPorts})}
 
 The \te{Port} type is used in definitions of instances of the
-\te{SplitPorts} type class (Section~\ref{sec-SplitPorts-typeclass}).
+\te{SplitPorts} type class (Section~\ref{sec:SplitPorts_typeclass}).
 
 \begin{quote}
   {\BH} syntax:
@@ -4397,10 +4419,10 @@ The \te{Port} type is used in definitions of instances of the
   \end{tabular}
 \end{quote}
 
-Specifically, it is used in argument types of the {\tt splitPorts}
-function and the result type of the \te{unsplitPorts} function of the
-type class.  With respect to the \te{SplitPorts} type class
-declaration shown in Section~\ref{sec-SplitPorts-typeclass}, in any
+Specifically, it is used in result type of the {\tt splitPorts}
+function and the argument types of the \te{unsplitPorts} function of
+the type class.  With respect to the \te{SplitPorts} type class
+declaration shown in Section~\ref{sec:SplitPorts_typeclass}, in any
 declaration of an instance of the type class, the type ``{\tt p}'' has
 the tuple-shape:
 
@@ -4410,7 +4432,7 @@ the tuple-shape:
 
 where each tuple component describes one of the ports in the split.
 
-Please see Section~\ref{sec-SplitPorts-package} which describes the
+Please see Section~\ref{sec:SplitPorts_package} which describes the
 use of the {\tt SplitPorts} type class and the type {\tt Port} in
 detail, including several concrete examples. That section also
 describes additional facilities available by importing the {\tt
--- a/doc/libraries_ref_guide/LibDoc/Prelude.tex
+++ b/doc/libraries_ref_guide/LibDoc/Prelude.tex
@@ -55,6 +55,9 @@ representation for use with \te{\$display} system tasks.\\
 \hline
 \te{StringLiteral} &Types which can be created around strings.\\
 \hline
+\te{SplitPorts} & Types that are split into multiple ports (buses) in generated Verilog
+(instead of a single, flat, bitvector port) \\
+\hline
 \end{tabular}
 \end{center}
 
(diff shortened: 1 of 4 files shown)


rsnikhil:
FYI: I just assumed the following, and haven't actually checked. I can remove it, if false:

    Although aimed at Verilog, BlueSim also treats split ports in the
    same way. In particular, it will generate waveform files with the
    same ports/buses seen in Verilog.

quark17:
I thought the plan was to create a section 2.9 containing the majority of the documentation, similar to how Generics is self-contained in section 2.8.  That is, the SplitPorts class would not be in 2.1, the Port type would not be in 2.2, and all the examples etc from 3.11 would be in 2.9 (leaving 3.11 to just have simple subsections 3.11.1 for SplitPorts and 3.11.2 for SplitVector, that briefly mention the types and functions there but refer back to 2.9).

I don't have a huge objection to the current approach, if you think it's the right way to go -- although should the classes and types in 2.9 Generics also be distributed to 2.1 and 2.2?

Bluesim does not yet support split ports, in that sense.  I believe that split types are rejoined into a single bit vector for Bluesim -- which we should improve at some point.  We can open an issue to record that.  (This also exposes how this kind of documentation doesn't fit neatly into one place -- libraries guide, user guide, it's own note? -- as Bluesim support is not really a libraries issue.)

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -1763,7 +1766,127 @@ \subsubsection{StringLiteral}
...
+     -- This must be the same length as the tuple of values.
+     portNames :: a -> String -> List String
+      typeclass StringLiteral #(type data_t);
+          function data_t fromString(String x);
+      endtypeclass
+\end{libverbatim}
+
+{\BSVb} syntax:
+\begin{libverbatim}
+   typeclass SplitPorts #(a, p)
```
quark17, line 1836: This code doesn't parse for a few reasons: the declaration needs `type` in front of variables, `fun` should be `function`, the syntax for associated type functions (`PortsOf`) is not like that, function arguments need names, and comments in BSV use `//`.  Try this:

```
typeclass SplitPorts #(type a, type p)
  dependencies (a determines p);
  type PortsOf#(type a) = p;

  // Convert a value to a tuple of values corresponding to ports.
  function p splitPorts (a x);

  // Combine a tuple of values corresponding to ports into a value.
  function a unsplitPorts (p x);

  // Compute the list of port names for a type, given a base name.
  // This must be the same length as the tuple of values.
  function List #(String) portNames (a x, String s);
endtypeclass
```

Although, I don't recall if I've written an ATF in BSV before, so I don't know if that actually works.  I'd also prefer that `type` not be required for the argument of `PortsOf`, because it's not introducing a new type, it can only refer to the types of the class.  But that's a separate feature to be documented at some point.  (Although, I'm now realizing that the `Bits` class can be updated to show `SizeOf` as an ATF member, in the libraries guide.)

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -1763,7 +1766,127 @@ \subsubsection{StringLiteral}
...
+  \item ``{\tt a}'' is a user-defined type to be split into separate
+        ports. It is typically a structured type (struct, vector,
+        tuple, algebraic type (tagged union)).
+
+  \item ``{\tt p}'' is an $n$-tuple type whose components describe each desired RTL port.
+\end{itemize}
+
+\begin{center}
+
+\index{SplitPorts@\te{SplitPorts}!\te{splitPorts}   (function in type class)}
```
quark17, line 1864: It might be helpful to keep all the directives in one place.  Note that at the top of this section is directives like this:
```
\index[function]{Prelude!splitPorts}
```
which adds the item to a separate "Function and Module by Package" index.

The convention in this document (both Prelude and other libraries) is to add it to both the functions index and the regular index.  See for example:
```
\index{SizedLiteral@\te{SizedLiteral} (type class)}
\index{fromSizedInteger@\te{fromSizedInteger} (\te{SizedLiteral} class method)}
\index[function]{Prelude!fromSizedInteger}
\index[typeclass]{SizedLiteral}
```

What you've written here doesn't follow the same textual convention for the class methods, FYI.  But also, maybe we should keep all these directives together?  Probably by moving the function directives down to here, which seems to be the convention used elsewere (in Prelude and, say, GetPut as an example of a library).

However, you have two sets of indexing here -- that's not a convention that's used anywhere else.  I would delete the first set here (the ones that next the members under the typeclass name), since that's what the "function" index does.

Also note, in the second set, you left out `splitPorts`.

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -4247,8 +4370,51 @@ \subsubsection{Rules}
...
+\begin{quote}
+  {\BSV} syntax:
+  \hmm\begin{tabular}[t]{l}
+    {\tt typedef union tagged \{} \\
+    {\tt \hmm  t Port;} \\
+    {\tt \} Port \#(t);}
+  \end{tabular}
+\end{quote}
+
+Specifically, it is used in argument types of the {\tt splitPorts}
```
quark17, line 4400: You have these backwards.  It's the return type of `splitPorts` and the argument type of `unsplitPorts`.

quark17:
This is the first feature that we've documented that spans multiple library packages.  Each leaf under Section 3, for example, is documenting one package.  But now we've added something under Section 3 that documents `SplitPorts`, `SplitVector`, and even no-package (as presumably you can define a port-splitting instance without importing either of those).

I wonder if we should start a Section 4, for cross-library features.  We could briefly document the Prelude pieces in Sec 2 (as you've done) and briefly document the package contents in Sec 3, but put the feature documentation and examples in Sec 4 -- under 4.1, say, and maybe someday Generics can be moved to 4.2.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1478 @@
...
+
+Consider a {\BHb} or {\BSVb} module compiled by \emph{bsc} to
+Verilog. Suppose an interface method has a ``struct'' argument or
+result, becoming a Verilog input or output port, respectively.  These
+appear in Verilog as flat bit-vectors (usually the concatenation of
+the bit representations of the fields).  These are difficult to use in
+several RTL-level situations:
+
+\begin{itemize}
+  \item Viewing waveforms (whether generated from Verilog simulation or from Bluesim).
```
quark17, line 36: Though if the waveform viewer has hooks for expanding vectors, it can be split without needing support in the language.  Also, there may be other downstream tools, not just waveform viewers, that require split ports or are easier to work with when split.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1478 @@
...
+\index{SplitPorts@\te{SplitPorts}!Shallow port splitting}
+\index{Shallow port splitting}
+\index{Shallow port splitting|seealso{\te{SplitPorts}}}
+
+\index{SplitPorts@\te{SplitPorts}!Deep port splitting}
+\index{Deep port splitting}
+\index{Deep port splitting|seealso{\te{SplitPorts}}}
+
+In general, a type can have \emph{nested} structure.  For example, in
+a struct type $t_1$ a field may have struct type $t_2$, or a vector
```
quark17, line 112: Using math notation is fine here, though the use of lowercase letters for concrete types threw me at first.  An option would be to use `T1` and `T2` (perhaps even in monospace type).  But it's also fine how it is.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1478 @@
...
+
+\subsubsection{Typeclasses}
+
+\label{sec:splitports_shallow_deep}
+
+\index{ShallowSplitPorts@\te{ShallowSplitPorts} (package for \te{SplitPorts})}
+\index{SplitPorts@\te{SplitPorts}!Related type class \te{ShallowSplitPorts}}
+\index[typeclass]{ShallowSplitPorts}
+
+\index{shallowSplitPorts@\te{shallowSplitPorts}
```
quark17, line 194: See my earlier comment above (on the Prelude typeclass) about how indexing of member functions works.

You need to add a directive like this to add each member to the "function" index:
```
\index[function]{GetPut!toGet}
```

Also, the wording you've used on the entry for the ordinary index doesn't exactly match what's used elsewhere, just FYI.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1478 @@
...
+            tagged Port .w1, tagged Port .w2,
+            tagged Port .z}                     : tuple6 (v0,v1,v2,w1,w2,z);
+        endcase;
+      Vector #(3, Bool) v = ?;
+      v[0] = v0; v[1] = v1; v[2] = v2;
+      return Bar { v:v, w:tuple2 (w1,w2), z:z };
+   endfunction
+
+   function portNames (_, base);
+      return Cons ((base + "_v_0"),
```
quark17, line 905: I seem to recall that BSV has a syntax for specifying a constant list more concisely, though maybe it's only for a constant Vector (which would need to be converted to a List).

It also surprises me that BSV doesn't required `tagged` (here and when using `SplitVector`).

doc/libraries_ref_guide/version.tex
```diff
@@ -1,4 +1,4 @@
-\author{Revision: 17 February 2024}
+\author{Revision: 8 August 2026}
```
quark17, line 1: We may want to automate the date here, since we've been forgetting to update it.  Or add a CI check that it's been updated when needed.

quark17 (commented):
I have not looked carefully at the examples, I apologize.  But here's my notes from skimming over.

By the way, I had not realized that the port names are specified with a separate member function.  I don't know how I would have combined it with the `Port` type or the `splitPorts` function, but anyway that's a question for the future, not for your or this documentation.

krame505:
I'll try to find some time to review this tomorrow.  

> Bluesim does not yet support split ports, in that sense. I believe that split types are rejoined into a single bit vector for Bluesim -- which we should improve at some point. 

That is correct.  Properly supporting port splitting in the waveform files produced by port splitting would need a more substantial change in the backend to preserve tuple-ness later. I think there maybe should actually be a deeper re-think of how port-splitting is handled in ASyntax.  Perhaps introducing tuples in ASyntax were a mistake, because they only immediately selected from or constructed as the final result of a method.

doc/libraries_ref_guide/LibDoc/Prelude.tex
krame505: Yeah, Julie's version does work in BSV.  There wasn't much thought put into the BSV ATF syntax, I am open to changing it.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1478 @@
...
+\begin{quote}
+  {\BSVb} syntax:
+  \hspace*{2em}\begin{tabular}[t]{l}
+    {\tt typedef union tagged \{} \\
+    {\tt \hspace*{2em}  Vector \#(n, a) SplitVector;} \\
+    {\tt \} SplitVector \#(n, a);}
+  \end{tabular}
+\end{quote}
+
+The {\tt SplitVector} type is a ``drop-in'' replacement for the {\tt Vector} type.
```
krame505: Perhaps explain what is meant by thIs - most of the type classes for which `Vector` has instances also have instances for `SplitVector`, so functions like `fmap`, `foldr`, `length`, `==`, etc. are defined automatically.  Maybe also worth moving the mention of `unSplitVector` from the example below to here?

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1478 @@
...
+
+% ----------------------------------------------------------------
+
+\subsubsection{Types for Splitting Vectors}
+
+\label{sec:splitports_vector_types}
+
+{\footnotesize (This example needs import of {\tt SplitVectors} package.)}
+
+The following types are in the {\tt SplitVector} package:
```
krame505: Perhaps add a brief explanation of why `SplitVector` exists - `Vector` is not split by default, and one can't define a `SplitPorts` instance for `Vector` without it being orphaned.

krame505 (commented):
Thank you Nikhil, this looks pretty good!  

I think it might also be helpful to add an example of how to override port splitting behavior in a specific case for a type that already has a `SplitPorts` instance, or for a third-party type where defining a `SplitPorts` instance would be orphaned - just define a new type wrapper around the existing type, and define whatever custom `SplitPorts` instance for the new type.

rsnikhil:
> FYI: I just assumed the following, and haven't actually checked. I can remove it, if false:
> 
> ```
> Although aimed at Verilog, BlueSim also treats split ports in the
> same way. In particular, it will generate waveform files with the
> same ports/buses seen in Verilog.
> ```

I’ve removed references to Bluesim

rsnikhil:
> I thought the plan was to create a section 2.9 containing the majority of the documentation, similar to how Generics is self-contained in section 2.8. That is, the SplitPorts class would not be in 2.1, the Port type would not be in 2.2, and all the examples etc from 3.11 would be in 2.9 (leaving 3.11 to just have simple subsections 3.11.1 for SplitPorts and 3.11.2 for SplitVector, that briefly mention the types and functions there but refer back to 2.9).
> 
> I don't have a huge objection to the current approach, if you think it's the right way to go -- although should the classes and types in 2.9 Generics also be distributed to 2.1 and 2.2?
> 
> Bluesim does not yet support split ports, in that sense. I believe that split types are rejoined into a single bit vector for Bluesim -- which we should improve at some point. We can open an issue to record that. (This also exposes how this kind of documentation doesn't fit neatly into one place -- libraries guide, user guide, it's own note? -- as Bluesim support is not really a libraries issue.)

I wasn't aware that we had settled on any plan.

The basic problem is that the organization of the libraries doc does
not suit the SplitPorts situation which contains:
1. A typeclass SplitPorts in the Prelude
2. A type Port in the Prelude
3. Packages SplitPorts and SplitVector that need explicit imports

I have split it one way, following the above structure, but any way we
split it (if we split it) is not ideal.

In the lib doc,
* Prelude typeclasses are described Sec.2.1
* Prelude types are described in Sec.2.2
* Packages to be imported are described in Sec.3

I have basically followed that structure, with cross-references
between the three parts.

I have placed a majority of the material in the third part, including
all the examples. Many of the examples don't need the packages, only
the Prelude, and I have said that in each such example.

Personally, given the copious cross refs, I'm ok with the current
organization, but I'm happy to reorganize it differently, if we feel
that would improve it.

doc/libraries_ref_guide/LibDoc/Prelude.tex
rsnikhil: I have updated the example per @quark17 ’s sugestion

doc/libraries_ref_guide/LibDoc/Prelude.tex
rsnikhil: I have revamped the indexing substantially in the Prelude/SplitPorts section, in the Libraries/SplitPorts section and in the Libraries/SplitVector section

doc/libraries_ref_guide/LibDoc/Prelude.tex
rsnikhil: Fixed

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: Added a note to this effect

rsnikhil:
> This is the first feature that we've documented that spans multiple library packages. Each leaf under Section 3, for example, is documenting one package. But now we've added something under Section 3 that documents `SplitPorts`, `SplitVector`, and even no-package (as presumably you can define a port-splitting instance without importing either of those).
> 
> I wonder if we should start a Section 4, for cross-library features. We could briefly document the Prelude pieces in Sec 2 (as you've done) and briefly document the package contents in Sec 3, but put the feature documentation and examples in Sec 4 -- under 4.1, say, and maybe someday Generics can be moved to 4.2.

I think this would be a good idea, particularly for SplitPorts, and
possibly also for Generics, as you suggest, but I suggest we should
postpone that possibility for the next release.

Another issue is whether we should generate "parallel" documents for
BH and BSV, as opposed to mixing them in a single document (in
SplitPorts section 3.11, the same examples are repeated in BH and in
BSV, adding bulk to the document).

rsnikhil:
> I have not looked carefully at the examples, I apologize. But here's my notes from skimming over.
> 
> By the way, I had not realized that the port names are specified with a separate member function. I don't know how I would have combined it with the `Port` type or the `splitPorts` function, but anyway that's a question for the future, not for your or this documentation.

Yes, this is an oddity. Also, the 'portNames()' method returns a list whose length must be equal to the tuple-width of 'p' in the 'splitPorts()' and 'unSplitPorts()'.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: > Using math notation is fine here, though the use of lowercase letters for concrete types threw me at first. An option would be to use `T1` and `T2` (perhaps even in monospace type). But it's also fine how it is.

I prefer it the way it is.  Using a slanting font and subscripts makes
it clear that this is not a literal, just a placeholder in the text narrative.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: I have substantially revamped all the indexing for SplitPorts.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: I'm not aware of any such function for lists (for vectors there is the
'vec()' function that uses the 'BuildVector' typeclass.
I have changed the example to 'toList (vec (....))'

I believe BSV has always been forgiving about omitting 'tagged' on
constructor applications, but this is moot here since I replaced
the Cons by 'toList (vec (...))'.

doc/libraries_ref_guide/version.tex
rsnikhil: Agreed, but I leave it to someone else to do since I don’t know what’s the mechanism for automating the date update.

rsnikhil:
> I'll try to find some time to review this tomorrow.
> 
> > Bluesim does not yet support split ports, in that sense. I believe that split types are rejoined into a single bit vector for Bluesim -- which we should improve at some point.
> 
> That is correct. Properly supporting port splitting in the waveform files produced by port splitting would need a more substantial change in the backend to preserve tuple-ness later. I think there maybe should actually be a deeper re-think of how port-splitting is handled in ASyntax. Perhaps introducing tuples in ASyntax were a mistake, because they only immediately selected from or constructed as the final result of a method.

I’ve removed references to Bluesim

rsnikhil:
> Thank you Nikhil, this looks pretty good!
> 
> I think it might also be helpful to add an example of how to override port splitting behavior in a specific case for a type that already has a `SplitPorts` instance, or for a third-party type where defining a `SplitPorts` instance would be orphaned - just define a new type wrapper around the existing type, and define whatever custom `SplitPorts` instance for the new type.

I’ve added a section on this.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: I’ve added text for this.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: I’ve added text for this.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
quark17: I wasn't thinking of a function, I was thinking of parser support.  But what I was misremembering was support for Array (though that could also be converted to a Vector):
```
Bool res[3] = { True, False, False };
```

I was also recently reminded that BH has the operator `:>` defined in `List` that can replace `Cons`:
```
let res = True :> False :> False :> Nil
```
Just FYI.

doc/libraries_ref_guide/LibDoc/Prelude.tex
```diff
@@ -4247,8 +4392,51 @@ \subsubsection{Rules}
...
 
-
-
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\com{Port}
+\subsubsection{Port}
+\label{sec:SplitPorts_Port_type}
+
+\index{Port@\te{Port} (related type for class \te{SplitPorts})}
+\index{SplitPorts@\te{SplitPorts}!\te{Port} (related type for class \te{SplitPorts})}
```
quark17: In hierarchical entries, you will need `@` also for the sub-level.  As written, `Port` is misalphabetized -- it appears at the start of the list, under `SplitPorts`.  The same occurs for `DeepSplitPorts` and `ShallowSplitPorts`.

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
```diff
@@ -0,0 +1,1619 @@
...
+
+%  ----------------------------------------------------------------
+
+\subsubsection{Example in {\BHb} of simplified canonical splitting}
+
+\label{sec:splitports_examples_BH_simplified}
+
+\index{SplitPorts@\te{SplitPorts}!simplified canonical port splitting}
+\index{Simplified canonical port splitting}
+\index{Simplified canonical port splitting|seealso{\te{SplitPorts}}}
```
quark17, line 710: FYI, this `seealso` entry is so long that it crosses into the second column.

quark17 (commented):
Looks good, aside from the few hierarchical indexings that need `@`.  Thanks!

doc/libraries_ref_guide/LibDoc/SplitPorts.tex
rsnikhil: Hmmm. In the PDF I get, it wraps around to the next line, does not cross into second column.  LaTeX version issue?

doc/libraries_ref_guide/LibDoc/Prelude.tex
rsnikhil: Fixed all three of these.

quark17 (approved):

PR #1086, opened 2026-08-12
Apply an STP patch to prevent hanging on ARM

matx-amy:
On platforms where plain `char` is unsigned (e.g. Linux on AArch64),
Cnf_ReadMsops in the vendored STP's bundled ABC stores -1 sentinels
into plain `char` arrays and then tests `pMemory[k] == -1`.  The
comparison promotes the (unsigned) char 255 to int, so it is never
true, and the loop that computes the SOP sizes never advances `i`.

Declare the sentinel-carrying storage `signed char`, matching the
intent of https://github.com/stp/stp/commit/69ed087a / https://github.com/bitblaze-fuzzball/stp/commit/308c97bf5bc068c6aa30aa92595879910e08bd18 and current
Berkeley ABC.  (https://github.com/stp/stp/commit/69ed087a as committed compares a signed char
against '(char)(-1)', which still misfires when plain char is
unsigned, so the comparison here is left as a plain '== -1' against
the now-signed operand.)

---

If we just wanted to update to vendoring STP 2.4.1, that should pull in this fix with the version bump.

### DIFF
--- a/.github/workflows/build-and-test-ubuntu.yml
+++ b/.github/workflows/build-and-test-ubuntu.yml
@@ -69,11 +69,6 @@ jobs:
           # Generate package info
           export BSC_BUILD_PKGINFO=1
 
-          # An STP test hangs on the ARM runners
-          if [[ "${{ inputs.os }}" == *-arm ]]; then
-              export STP_STUB=1
-            fi
-
           make -j ${CORES} GHCJOBS=2 GHCRTSFLAGS='+RTS -M5G -A128m -RTS' install-src
           tar czf inst.tar.gz inst
       - name: CCache stats
--- a/src/vendor/stp/src/extlib-abc/aig/cnf/cnfData.c
+++ b/src/vendor/stp/src/extlib-abc/aig/cnf/cnfData.c
@@ -4539,7 +4539,7 @@ void Cnf_ReadMsops( char ** ppSopSizes, char *** ppSops )
         { 0x0F0F, 0xF0F0 },
         { 0x00FF, 0xFF00 }
     };
-    char Map[256], * pPrev, * pMemory;
+    signed char Map[256], * pPrev, * pMemory;
     char * pSopSizes, ** pSops;
     int i, k, b, Size;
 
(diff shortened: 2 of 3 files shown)


quark17:
Awesome, thanks for figuring this out!

If someone had the time, I'd suggest eliminating the STP snapshot and instead have the build check out the STP GitHub repo (like we do for Yices) at a newer version.  (The snapshot predates STP being on GitHub!)  As you note, STP 2.4 was recently released.  Yices also has newer releases, that we could upgrade to.  This may be very straightforward to do, but we'd have to check whether the C APIs have changed at all (for the subsets that we use), which would require changing how we import the libraries.  We could also audit whether there are new API functions that we could start using, in place of any contortions or limitations we might currently have in how we use the imported libraries.

quark17:
The CI is failing to compile STP on the Ubuntu 26.04 ARM VM.  Not because of your change, though.  There are some errors complaining that `#include <ctime>` is needed.  I now recall that I had run into this (when I added the ARM VM) and that I could get past it by adding new imports (and that didn't break the other VMs).  But since I was disabling STP anyway, I chose not to include those changes.  Digging up that work, I see that I had commit that added this to `stp/src/sat/cryptominisat2/Solver.h` (right before the include of `time_mem.h`):
```
#include <ctime>
#include <pthread.h>
```
which I think worked, but I don't know if that't the right resolution.  From the CI log, for example, it looks like the `ctime` import might only be needed in `MersenneTwister.h` (which `Solver.h` imports).

matx-amy:
It looks like bumping STP makes it LGPL; given that yices2 is already GPL, I don't know if we care, but worth noting.

We had some discussion about solvers internally; it sounds like Ravi has patches cooking that would make #117 more palatable, and would also point at moving from STP and yices2 to Bitwuzla and CVC5, which we then wouldn't need to build from source, either.

If it'd be strongly preferable to have a more up-to-date STP in the interim, I can do that; otherwise, I'm inclined to take the small win here and just plan on switching solvers in the longer-term.

PR #1094, opened 2026-08-18
Add an -elab-only flag: stop compilation at the .ba

matx-jeffnewbern:
With code generation available from a module's `.ba` alone (`-c`, #1092) and
at link time (#1093), a compile no longer needs to produce backend output
itself: with `-elab-only`, compilation stops after elaboration.  For the
Verilog backend that means `genModuleVerilog` is not run at all -- exactly
as a Bluesim compile already behaves, verified by a test on the `-v` phase
trace -- and each module's `.v` is generated later, from its `.ba`, by `-c`
or by the link.  The flag is backend-agnostic when compiling source: with
`-sim` (or no backend) stopping at the `.ba` is already the behavior, so it
is accepted as a no-op, letting build systems pass it unconditionally.

The wrapper's port properties are unaffected: they come from the APackage
analysis (`getIOPropsA`, #1060), before the backend split, and are recorded
in the `.bo` under `-elab-only` too, so a parent compiled against an
`-elab-only` child deduces exactly what a full compile deduces.  The new
testsuite directory pins byte-identity of the staged flow against a direct
compile -- including under `-semantic-ports-comment`, whose comment analysis
`-c` recomputes from the `.ba`.

The `-u` check correspondingly stops expecting `.v` files from an
`-elab-only` compile, checking the `.ba` instead.  The flag applies only
when compiling source (S0100) -- including the no-filenames link path, which
bypasses `checkLinkFlags` -- and cannot be combined with `-no-elab`, which
would suppress the `.ba` that the deferred code generation needs (S0101).

Stacks on #1093; the diff includes the earlier commits until they merge.

New test directory: `bsc.verilog/elab_only`.  Full testsuite at this commit:
20,926 expected passes, 134 expected failures, 0 unexpected failures.


quark17:
I could have sworn that BSC already had a way to generate Verilog modules from `.ba` (as in #1092) and already had a way to stop at `.ba` for the Verilog gen mode -- but maybe I had only proposed those, back when we added the `-elab` flag, and never actually implemented it.

I wonder if we should invert the sense of the flag, and have something like `-gen-module` (not necessarily that name) which is on by default for Verilog, so to stop at `.ba` you would specify `-no-gen-module`.

Though I like thinking about it as three modes -- an elab mode (`bsc -g`), a backend gen / synth mode (`bsc -c`, say), and a link mode (`bsc -e`) -- where elab and gen can be combined in one call (depending on whether `-c` is also provided) ... though that would require changing the command-line behavior for the Verilog backend (to not generate by default, or else we'd have to allow `-no-c`, say).  Anyway, this is kind of what I suggested in [a comment on 988](https://github.com/B-Lang-org/bsc/pull/988#pullrequestreview-4613235979), where I suggested `-s` rather than `-c` (or `-m` would work too).

Apologies, I haven't fully absorbed these PRs (for example, what does "root only" mean in #1092?), so these are initial thoughts.

nanavati:
The three phase model is exactly what we are going for. The main constraint is that we don't want to break existing uses of bsc.

That constraint means that the behavior across the Bluesim and Verilog backends is different. For Verilog -g is bundled with -c so you get Verilog code and -e only does linking. For Bluesim the bundling is the opposite -g does only elaboration, but -e does both code generation and linking. The Bluesim bundling is a little more natural because you can think of it as linking automatically generating missing or stale artifacts (something I think makes sense to add to the Verilog linking flow after we produce .ba by default), but it is still a bundling.

The root-only comment means that when you use -c on a .ba file you generate code for exactly that module, no more and no less. That's the  behavior that enables clean integration into build systems like Bazel (where we don't want to duplicate work across linking targets and where the build system might want to own the link step anyway).

PR #1097, opened 2026-08-22
Fix swapped low/high duration parameters in bk_define_clock declaration

matx-amy:
The `bk_define_clock` definition in `kernel.cxx` uses its sixth argument as the high-phase duration and its seventh as the low-phase duration (matching `bk_alter_clock`), but the declaration and doc comment in `bluesim_kernel_api.h` had the two names reversed.


matx-amy:
> The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

yet another transient error...

PR #1101, opened 2026-08-24
Fix off-by-one heap overflow in BufferTarget circular writes

matx-amy:
Reproducer:

```
package Foo where

mkFoo :: Module Empty
mkFoo = module
  x :: Reg (Bit 2) <- mkReg 0
  y :: Reg (Bit 16) <- mkReg 0

  rules
    when True ==> do
      $swrite (asReg y) "01%b" 0xff
      x := x + 1
    when x == 2 ==> do
      $display "%h" y
      $finish 0
```


testsuite/bsc.verilog/tasks/tasks.exp
```diff
@@ -58,6 +58,12 @@ test_c_only_bsv StringFormat3
 test_c_only_bsv DynamicFormat
 test_c_only_bsv DynamicFormat2
 
+# Test that $swrite output which overflows the destination register is
+# truncated correctly when it wraps Bluesim's circular buffer (an
+# off-by-one in BufferTarget wrote past the buffer and visibly
+# corrupted the Bit 24 case)
+test_c_only SwriteWrap
```
quark17, line 65: I notice that you're using `test_c_only`.  Which makes sense, as it's what the block of tests above are doing, whose comment says that Verilog testing is turned off for `$swrite` tests.

That could be because the test isn't for BSC, but for the internals of Bluesim, and so running with a Verilog simulator would only be testing whether the Verilog simulator is correct -- which we shouldn't waste our time doing.  But presumably it would be worth confirming that the expected output file (that we test Bluesim with) still matches the Verilog simulator?

The other reason why Verilog might have been turned off is if `$swrite` wasn't well-supported by Icarus Verilog.  But if that were the case, I would have expected the tests to use a "bug" mechanism (like other tests in this file) to make it explicit that we want to test Verilog but can't for Icarus Verilog.  I don't know if the original author mistakenly didn't do that or intentionally didn't.  Current Icarus Verilog compiles the file fine; I'd have to try an older version to see if it didn't before.

What do you think about changing the `test_c_only` to `test_c_veri` here and in the tests above?

Issue #688, opened 2024-04-06
Bluesim divide-by-zero behavior is inconsistent on arm64


quark17:
In `testsuite/bsc.misc/divmod/` there are tests for division-by-zero behavior in Bluesim and Verilog backends.  The Bluesim tests are expected to raise SIGFPE.  For wide bit vectors, this is achieved by defining `operator/` for the `WideData` type that Bluesim uses to represent such vectors.  Smaller types are represented in Bluesim with `uint64_t`, `uint32_t`, `uint8_t`, and Bluesim relies on the operator for those types to raise SIGFPE.  This happens on `x86_64` architectures, but does not happen on `arm64` (which allows the division to return zero).  So on `arm64`, the behavior is different for small vectors than for wide vectors.

I'm guessing that we probably want to have Bluesim raise SIGFPE on all architectures.  We could do that by replacing the `/` operation with a macro or something, which can be defined differently for `arm64` (to check for zero and raise SIGFPE).  Alternatively, I guess we could change the operator for `WideData` to return zero when on ARM, and thus make the behavior on ARM consistent that way (which might have less overhead than guarding every division?).  Any opinions?

In the meantime, I'm going to change the testsuite to expect no error on ARM, so that the testsuite doesn't fail.  GitHub now offers a macOS VM on arm64 hardware (`macos-14`) and I want to add a CI job that builds and tests BSC on this, but currently the testsuite has this failure.

FYI, the `WideData` representation (and operloaded operators) are declared in `bsc/src/bluesim/bs_wide_data.h` and the implementations are in `bsc/src/bluesim/wide_data.cxx`.  It is the implementation of `wide_quot_rem()` that raises SIGFPE (line 1221 in `wide_data.cxx`).

mieszko:
> I'm guessing that we probably want to have Bluesim raise SIGFPE on all architectures.

Why? That certainly does not respond to anything you'd see in hardware, right?

And it's not just ARM that doesn't trap for integer division. RISC-V returns all bits set, no trap. In PowerPC the result is undefined, no trap. The C standard leaves the result undefined.

Perhaps more relevantly, SystemVerilog defines division by 0 to result is `'x`, which if you squint you could argue means undefined in some sense. But either way, no trap.

I think there are two rational things to do: either match Verilog behaviour (which we can't b/c we have no `x`) or leave the behaviour of things like this explicitly undefined in the language spec. I'd argue for the latter.

rsnikhil:
[...]

PR #1120, opened 2026-09-14
Bluesim: define divide-by-zero instead of UB or SIGFPE

matx-william:
Bluesim's division-by-zero behavior is unfortunate: in the best case, it produces a SIGFPE that kills the entire simulation with no information about where or why.  In the worst case, the generated code exhibits Undefined Behavior potentially resulting in nasal demons.  Neither behavior is appropriate for a simulation engine.

This PR addresses this issue by defining pure total `safe_quot` and `safe_rem` functions.  The Verilog backend defers to the Verilog simulator, which yields X. Bluesim is 2-state and has no X, so I went with the result of a divide by zero to be all ones on both paths.  All ones is chosen because it can never be produced by a well-defined remainder (for b != 0 we have a % b <= b - 1 <= 2^w - 2), so an all-ones remainder is an unambiguous marker that a zero divisor was seen.

The helpers need to be pure and total because Bluesim evaluates divisions whose results are discarded: a rule body of the form `if (b != 0) r <= a / b` computes a / b unconditionally and only guards the write.  A zero divisor therefore reaches the operator even in correct designs, which rules out both a fatal error and a runtime warning -- either would fire on code that is not buggy.

They are templates rather than overloads because the operand widths need not match (primQuot :: Bit k -> Bit n -> Bit k), and a plain overload set is ambiguous for mismatched narrow widths such as tUInt64 / tUInt8.

Tests: rewrite the divmod expectations, which encoded the SIGFPE behavior, and add coverage for all ones at widths that are not multiples of the 32-bit word size, and for the body-guarded case, where Bluesim and Verilog now agree.


quark17:
The commit message references Issue #688, but not the PR text, so I wanted to mention that here.  Also, Bluespec's internal bug database has Issue 1342, requesting that Bluesim not evaluate unused defs although the main motivation was divide by zero (in unused paths).  (There, a workaround is proposed -- by writing `let q = (b == 0 ? 0 : divFunc(a,b)` the C++ compiler can see to short-circuit -- which is used in the test `bsc.lib/FixedPoint/Quots.bsv`.  If we're fixing things, we can remove that workaround from the test.)

You might consider making the result value configurable.  Bluesim could have a flag that controls what 2-value vector should be given to X results (even if the options are just 0, 1, and A).  (Then someone can run multiple sims with different values, just to confirm that the design isn't relying on a particular one.)  Note that BSC has an `-unspecified-to` flag, for what it should do with any undefined values that remain when it reaches code-gen (so that Bluesim and Verilog can match, or to specify X to allow Verilog to diverge and optimise).  I wouldn't use that flag here, since it's something different, but I mention it for context.  Though, in theory, BSC could have a flag that instructs it to generate Verilog code with wrappers around div/quot operators to make the error results match what Bluesim does.

matx-william:
Personally, having a configuration option for this feels a touch sledgehammer/fly.  But I'm happy to implement whatever behavior the Bluespec Authorities prefer.

quark17:
In case it wasn't clear: I wouldn't have a flag just for div/quot behavior.  I was saying that Bluesim could have (if it doesn't already) a flag for specifying what value to use in place of X, when doing 2-value sim -- where div/quot is just one source of X.

